PR #1, opened 2016-11-10
A few fixes to CitrineDataRetrieval and added unit tests for retrieval

montoyjh:
This PR fixes a few issues I was having with citrine data retrieval.  I went ahead and added a (very basic) unit test for CitrineDataRetrieval and MPDataRetrieval, following the convention in the MPRester unit test in which they are skipped without proper api credentials in the environment.

Also, as a side note, I had a bit of trouble with the citrine python API.  The PyPI versions of citrination-client do not work properly with MatMiner (nor does the master branch in their github repo).  I think users can figure it out by following the instructions at the link in the CitrineDataRetrieval docstring, but I figured I'd note it here just so it was on the radar.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   python:
-    version: 2.7.12
+    version: 3.5.2
 dependencies:
   override:
     - pip install --upgrade pip
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   python:
-    version: 3.5.2
+    version: 2.7.12
 dependencies:
   override:
     - pip install --upgrade pip
--- a/matminer/data_retrieval/retrieve_Citrine.py
+++ b/matminer/data_retrieval/retrieve_Citrine.py
@@ -120,7 +120,7 @@ class CitrineDataRetrieval:
                             prop_df = prop_df.pivot(columns='property.name', values='property.scalar')
                         elif 'property.matrix' in meas_normdf.columns:
                             prop_df = prop_df.pivot(columns='property.name', values='property.matrix')
-                        prop_df = prop_df.convert_objects(convert_numeric=True)  # Convert columns from object to num
+                        prop_df = prop_df.apply(pd.to_numeric) # Convert columns from object to num
                         # Making a single row DF of non-'measurement.property' columns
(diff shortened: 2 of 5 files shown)


computron:
Thanks!

The CircleCI fails (e.g., click the link) b/c the citrination_client is not installed. Can you add it to requirements.txt and setup.py? I understand the problem existed before this commit but maybe you can just fix that as part of this.

montoyjh:
So citrination_client has an absolute/relative import issue in the top level module that prevents it from being used in python3.5.  I can ask them to fix it on their legacy branch in their github repo.

montoyjh:
I dunno whether they'll be amenable to a new release of their legacy branch though, so we might want to figure out how to align MatMiner with the master branch.

montoyjh:
The citrination folks merged my changes in PCC that fix the import issue in 3.5.  The requirements.txt file points to the legacy branch of their github repo, which I think is appropriate.  It's not as clean as I would like, but I think it's useful to keep folks from having to hunt down the correct python-citrination-client version as I did.  Note also that the MP and CDR tests don't run in CircleCI without API keys set.

computron:
@saurabh02 Can you please sort out this issue? You are responsible for MatMiner

saurabh02:
Yes, I am conversing with Joey about the changes. Seems to be all fixed now.

computron:
@saurabh02 Ok but why can't this work with the master branch of Citrine? why do we need to resort to a legacy branch?

saurabh02:
Because they have changed the way their client search query works in their latest master branch. There's no documentation on it and their website still refers to the old version http://citrineinformatics.github.io/api-documentation/?python#introduction, but I think I may be able to figure out how it works.

montoyjh:
Yeah, I started this process aiming to use the master branch, but it's a heavier task than I anticipated and I wanted to get to a starting point where the code was useable and had explicit, accurate requirements.

saurabh02:
Thanks @montoyjh!

computron:
>> but I think I may be able to figure out how it works.

computron:
>> but I think I may be able to figure out how it works.

@saurabh02 can you just contact Citrine and see what is going on?

saurabh02:
Yes okay

PR #4, opened 2016-12-09
Fist commit for coordination number algorithms

aykol:
Adding multiple coordination number algorithms:
- A base CN class ensures consistency among different algorithms
- Each algorithm takes a structure and site index and returns a dictionary of coordinating atoms.

TODO: 
- Post-processing is required to derive more generic descriptors
- Needs some cleanup, especially for helper functions

### DIFF
--- a/matminer/descriptors/examples/coordination_numbers.ipynb
+++ b/matminer/descriptors/examples/coordination_numbers.ipynb
@@ -1,0 +1,115 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from matminer.descriptors.structure_features import CN_OKeeffes, CN_OKeeffes_mod, CN_BrunnerReal, \\\n",
+    "                                                CN_BrunnerReciprocal, CN_BrunnerRelative, CN_ECoN"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Currently avaialble CN algorithms in MatMiner. \n",
+    "\n",
+    "Each algorithm starts with CN_ and is derived from the CNBase class.\n",
+    "\n",
+    "* **CN_Okeeffes:** Weighted Voronoi CNs (O'Keeffe's method).\n",
+    "* **CN_Okeeffes_mod:** A modified version of OKeeffe's algorithm.\n",
+    "* **CN_ECoN:** Effective Coordination Numbers, ECoN, (Hoppe's method).\n",
+    "* **CN_BrunnerReciprocal:** Brunner's method of largest reciprocal gap in interactomic distances.\n",
+    "* **CN_BrunnerRelative:** Brunner's method of largest relative gap in interactomic distances.\n",
+    "* **CN_BrunnerReal:** Brunner's method of largest gap in interactomic distances."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Each algorithm is a class, which can be initialized with its own parameters\n",
+    "\n",
+    "okeeffe_params = {\"cutoff\": 10.0}\n",
+    "econ_parmas = {\"radius\": 10.0}\n",
+    "\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Example use of algorithms:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "{u'As': 3.9999999982682555, u'Ga': 0.5381161641802799}\n",
+      "{u'As': 4}\n",
+      "{u'As': 4.0, u'Ga': 12.0}\n",
+      "{u'As': 4.0}\n",
+      "{u'As': 4.0, u'Ga': 12.0}\n",
+      "{u'As': 4.000000720303867, u'Ga': 1.896561960444926e-07}\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Each algorithm takes a structure and site index as input and \n",
+    "# returns the coordination numbers for that site as a dict.\n",
+    "\n",
+    "\n",
+    "# Let's calculate CNs for site with index 0 in \"mp-2534\" from Materials Project using our algorithms:\n",
+    "\n",
+    "cn_methods = [CN_OKeeffes(params=okeeffe_params), CN_OKeeffes_mod(), CN_BrunnerReal(), \n",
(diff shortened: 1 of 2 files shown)


computron:
@aykol Not sure about the abstract class design for CNs ...

For example, why does compute return a dict of elements to coordination? It should return a list of coordinated sites. From that you can do element projections, or you can count the number of sites, or you can also use the coordinated site information to build networks/graphs etc

aykol:
@computron because that won't be general enough; many algorithms can't be implemented that way. Not many algorithms return integral coordination numbers. They do some sort of weighting to get a single number.

saurabh02:
@aykol @computron and I guess having different output formats from different algorithms is not desirable? 

One (verbose) solution could be to have a standard format of returning for each site in the structure, a list of sites coordinated to it along with their corresponding weights. So, both types of algorithms, that count the number of sites within a specific cutoff radius and those that assign weights, could be accommodated this way.

aykol:
@saurabh02 @computron When we say "coordination number" the common expectation is to get a single "number" as a metric for number of near neighbors of a site, usually classified based on chemical identity of neighbors, or given as a total number. It's a very basic descriptor of local environment that I expect to get as a "number" that I can use right away, not a list.

If you expect a list of neighboring sites to draw graphs, networks, that sounds more like "connectivity", not CN. That's a higher level description of local environment, and is usually simpler to describe (almost all algorithms start with a Voronoi partitioning or a distance cutoff to define "connectivity" initially; not much variance there). Of course CN algorithms usually build on some sort of "connectivity" description to start with, and I think it's possible that those neighbor-lists and weights can be returned to user, in some way. But returning them in place of CN itself would be very confusing. I believe CN algorithms should return only CNs as their main output.

computron:
@aykol I am completely OK with an interface/abstract class that requires only returning a number.

*But* the actual method you push requires a bizarre dict like
{"O": 2, "F": 2}

If you can project on the elements to produce those sub-numbers (rather than just returning 4), then you can certainly give a real connectivity.

If you want you can have two classes -

-one for CNs that only give you a float
- one (that subclasses the first) that gives you CN for a float as well as connectivity. This class will also have built-in convenience functions that use the connectivity info to project on elements, ions, cations, anions, etc.

aykol:
@computron OK got it. I will separate those as you described. Also to make sure I understand; why do you think {"O": 2, "F": 2} is bizarre? We can add a method that returns total CN as a float for a given site, but not sure what would be a better way to return element-specific CNs. E.g. running the algorithm on a particular Al site in an AlOF compound returns this dict of element-specific CNs (like {"O": 2, "F": 2}) for that Al site.

computron:
I would suggest something like

CNBase
- get_coordination_number

CNAdvanced extends CNBase
- get_coordinated_sites_and_weights
- get_coordination_number (by default, the implementation just calls get_coordinated_sites_and_weights and sums the weights from all sites. The user can reimplement the method within their subclass if they have a more efficient way)
- get_coordination_by_elements (by default, the implementation just calls get_coordinated_sites_and_weights and uses that to create the {"O": 2, "F": 2} dict. The user can reimplement the method within their subclass if they have a more efficient way).

aykol:
Thanks @computron , it's much more clear now. I'll work on implementing this.

shyuep:
Pls shorten the names to get_cn.

I would also suggest avoiding an unnecessary class hierarchy. There is no reason for CNAdvanced to extend CNBase given that the CN is an average obtained from the more complete analysis done by CNAdvanced.

Just have one class - CNCalculator and provide different methods with different degrees of detail. If you want a reference, see how pymatgen's Structure object has get_distance_and_image and get_distance and the latter merely calls the former and provides the relevant info.

You should only implement different classes if there is significantly different functionality and there is a "is-a" relationship. E.g., Slab is a type of Structure, but has functions like surface energies etc. that are not common in Structure. CNBasic and Advanced fail this criteria.

shyuep:
To add, the only reason to have a class hierarchy is if you have two different algos for CN calculation. E.g., SimpleCNCalculator based on just a straight cutoff distance (maybe by ionic radii), VoronoiCNCalculator based on voronoi decompositions, etc.

computron:
@shyuep according to @aykol , there are algorithms that give you a coordination number but don't actually give you coordinated sites. So those algorithms cannot be implemented as CNAdvanced but they could be implemented in CNBase. Otherwise if all algorithms can implement CNAdvanced, there is no reason for CNBase.

Also I disagree with the "fails is-a" criteria. A CNAdvanced algorithm is a CNBasic algorithm but can also can give you particular sites. That is why CNAdvanced also has a get_coordination_number() method just like CNBasic, and it automatically uses the get_coordinated_sites_and_weights to determine this number. If I give you a CNAdvanced algorithm, I am also giving you a CNBasic algorithm. But not vice-versa, since as I mentioned some algorithms don't give you the actual sites according to @aykol 

As for naming
- cn sounds good to me
- if what I said above regarding the need for two classes holds up, there can be a better naming convention than CNBasic and CNAdvanced. I just put up any names to demonstrate the way it works.

computron:
@shyuep also regarding this comment:
"Just have one class - CNCalculator and provide different methods with different degrees of detail. If you want a reference, see how pymatgen's Structure object has get_distance_and_image and get_distance and the latter merely calls the former and provides the relevant info."

Maybe read what I wrote about CNAdvanced in the original thread and how get_coordination_number just calls get_coordinated_sites_and_weights? i.e. I already suggested to do exactly this...

computron:
@aykol Can you clarify if there are actually algorithms that give you a CN number but don't give you actual sites and weights? Or was your comment earlier in the thread simply that some CN algorithms require a weighting scheme (but can indeed tell you what sites are coordinated with those weights)?

Reading the above thread again, I am guessing it is the latter. If so, then you just need one CN class like @shyuep  says. The CNAdvanced and CNCalculator are essentially the same thing, but @shyuep 's names are better.

If the former, then there is a case for two classes

shyuep:
I will wait for @aykol to reply. I cannot see a way for an algorithm that only supports a scalar CN but does not give any info about what are the neighbors that result in that CN. Certainly, some algos do not assign weights (in which case the weight is simply a uniform weight).

aykol:
@computron you're right, it is the latter. my comment was for returning a list of sites alone instead of CN, sorry if it wasn't clear. All algorithms that I'm aware of certainly build on a list of of neighboring atoms and some weighting method (or uniform weights per se). So one base-calss as @shyuep suggested should work. From above discussion it sounds like the class can have two mandatory methods that need to be defined; get_neighbor_list and get_weights. And we can have some build-in methods that derive a variety of coordination properties from these two (total CN, element wise decomposed CN, etc.)

Thank you both for the feedback. I will soon submit a revision based on these.

computron:
@aykol why two methods? why not one method (get_cn_neighbors_and_weights) that returns a tuple?

([neighbors], [weights])

especially since if you are going to do the computation on-the-fly for one of these implementations, you don't want to have to call the method to analyze a site twice. If you are precomputing the data on all sites in advance for all implementations, I guess it doesn't matter.

computron:
(you could also used namedtuple or dict instead of th)

aykol:
@computron sure; returning a dict or tuple sounds good.

computron:
Hi @aykol - this PR is 20 days old, so I am going to close it for now. Please re-submit the PR when ready!

aykol:
@computron absolutely! thanks for all the feedback.

PR #6, opened 2017-01-26
Added minimum relative distances (structural descriptor).

nisse3000:


computron:
Note - similar methods are used as part of the VolumePredictor in matminer. Sayan is working on improvement. You might want to chat with him about his project

PR #7, opened 2017-01-27
Corrected atomic packing fraction descriptor.

nisse3000:
Tried to push directly, but got denied permission.


computron:
I tried inviting you to the Github group, but it says "invitation already sent" (this might have been sometime in the past). Can you check it again?

I will close this PR for now so that we can get you set up with direct commit

PR #8, opened 2017-02-28
added O'Keeffe bond length predictor function and parameters

sayanroyc:
added O'Keeffe bond length predictor function to structure_features.py
created O'Keeffe parameters json file okeeffe_params.json


computron:
@sayanroyc Are the O'Keefe parameters already in pymatgen? If so we don't need to duplicate them here since PMG is a dependency of MatMiner

nisse3000:
They seem to be the same as the one stored in bvparam_1991.yaml (in pymatgen/analysis).

nisse3000:
@computron Sorry for merging prematurely.

sayanroyc:
They are!

nisse3000:
It should be a quick fix because it's only about replace loading a json with a yaml file.

    with open(os.path.join(xxxx, 'bvparam_1991.yaml'), 'r') as f:
        data = yaml.load(f)

where only xxx has to be replaced by the pymatgen install directory + 'pymatgen/analysis/'. And, importing yaml instead of json.

sayanroyc:
Putting fixes in a new pull request

nisse3000:
I've just pushed a consolidated version.  What is still missing is the use of the PMG yaml.

sayanroyc:
from pymatgen.analysis.bond_valence import BV_PARAMS

That should be a dictionary of parameters where the key is the Element object

PR #9, opened 2017-02-28
deleted bv parameters json, now imported from yaml file in pymatgen

sayanroyc:


computron:
@sayanroyc you need to resolve the conflicts before this PR can be accepted

PR #10, opened 2017-05-03
Added Partial Radial Distribution Function

WardLT:
Added the ability to compute partial radial distribution functions

Also added "citrination-client" to setup.py, which seemed to be missing.


computron:
Thanks! I will probably move the citrination client as an optional dependency only for those that want to pull data from Citrine but some of the others were definitely missing.

Issue #11, opened 2017-05-03
Potential Issue with RDF Method


WardLT:
I think there are a few issues with the RDF calculation method ([link for convenience](https://github.com/hackingmaterials/matminer/blob/master/matminer/descriptors/structure_features.py#L40))

1. RDF are typically normalized by number density. This function appears normalize by mass density
2. In the loop over pairs of neighbors, pair distances are used as keys for the `dist_rdf` dict. In the loop when normalizing `dist_rdf`, it seems like the keys are treated as bin indices (based on the shell thickness calculation)

If you agree these are problems, I can make these changes. As you have tests for this function already, I figured it worth checking in before altering both the code and the tests.

shyuep:
Note that I don't think (1) is ultimately an issue. It is merely a scaling factor and has no effect on most trend analysis. That said, I agree it is usually normalized by number density. 

More generally, I find this to be a very poor and overly complicated implementation of a RDF. You do not actually need to do the whole complicated dist_rdf[dist] += 1 (first off, a better implementation of that would use a Counter). You simply need to create a sequence of all distances with no counting at all, then use numpy.histogram to bin all the distances by dist, i.e., somehting like

```
# Use get_all_neighbors instead of get_neighbors
# Compile the distances into a list.
# Concatenate the distances
# Create histogram of distances (numpy will even normalize it properly for you if you specify normed=True)
```

I think this will accomplish the RDF in less than 5 lines, and will probably be 10x or more faster.

computron:
Hi all,

Just catching up with this thread. I haven't been actively developing or reviewing matminer development because we're trying to push through "atomate" at the moment and my full focus (when it's possible for me to code...) is on atomate.

@WardLT - thanks for getting involved with matminer and for your suggestion about normalization. For now I have pulled your PR especially since it helps with setup. In the future it would suggest to separate the PR for setup.py vs the partial RDF but again I am glad that you are taking the time to contribute.

@shyuep - I agree with your implementation suggestions (actually, the get_all_neighbors + numpy histogram method was my suggestion to the original implementor - I guess it was ignored). If @WardLT wants to try that implementation I would be all for it. Otherwise, I will definitely make sure to clean up the code when I get around to making a push for matminer.

WardLT:
[...]

PR #12, opened 2017-05-03
Add prdf

WardLT:
Following discussion in #11, I've changed the implementation of `get_rdf` and `get_prdf` to use `get_all_neighbors`  and numpy's histogram.

In changing this, I've changed those two functions to return dense ndarrays rather than dicts. I have updated the documentation and tests accordingly.


computron:
(Thanks!)

PR #14, opened 2017-05-25
New source of data: MPDS

### LINKED ISSUE #13: More sources of data
Thanks for this project! This is not an issue, but more like a feature request: would you be interested in adding more sources of data (e.g. Nomad, Materials Cloud etc.)?

blokhin:
Partially closes #13 created by me


computron:
Hi @blokhin,

Thanks for taking the time to contribute this data adapter, it is very much appreciated. I did have some comments on it:

1. Can you clarify in the docs somewhere what kind of data is available freely and what requires a commercial license for MPDS? It doesn't have to be overly long or detailed, but just enough that someone that stumbles upon the code can know enough to decide whether they should look more into MPDS if they don't have the budget for a license.
2. I'd like to avoid having a set of environment variables to control import of optional dependencies ("SKIP_MPDS_DEPS").  Typically we just use a try-except method. See for example this behavior in pymatgen: https://github.com/materialsproject/pymatgen/blob/master/pymatgen/io/babel.py . Overall, most of the Materials Projects codebases are moving (even if slowly) towards reducing use of environment variables.
3. It would be preferable, although not required, to perform the ASE functions in pymatgen instead. We have nothing against ASE, it's just that pymatgen is already a required dependency for matminer so users will already have that installed. If users need to additionally install ASE in order to use MPDS adapter, it will be another hurdle for them to cross in order to get this piece of code working. Generally, I would encourage the use of ASE only in "heavyweight" cases where the functionality is either missing or difficult/awkward to implement in pymatgen. Skimming the code, I didn't see that to be the case here. What do you think?

Thanks again!

saurabh02:
Thanks @blokhin for contributing this! I just had some quick PEP comments based on a cursory look (@computron can also comment on these if needed),

- Can the MIT License Info go elsewhere outside the MPDS data retrieval code?
- Can you include information about the arguments (name, type) and return data for functions in the MPDS data retrieval code?
- Can you also include your name as author name in the  files you contributed to?

blokhin:
Dear @computron and @saurabh02 thanks for your kind feedback. I will re-PR as soon as return back from China in a week.

blokhin:
@computron @saurabh02 your points have been taken into account

computron:
Hi @blokhin - thanks for taking the time to consider and incorporate these comments. I'm pulling now so that others can give it a try if they're interested!

PR #17, opened 2017-06-23
Remove invisible non-ascii character from README

mkhorton:
Was causing error `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 5096: ordinal not in range(128)` when trying to install matminer via pip on some systems.

Bit of a weird bug! Character was `\xa0` which is a non-breaking space (aka `&nbsp;` in HTML, probably crept in when copying and pasting the homebrew installation command)


computron:
I would say the problems with these docs go beyond the invisible character ... will probably give it a careful look sometime in the future

PR #20, opened 2017-06-30
Added the option to save html plots without showing them in figrecipes/plotly/make_plots.py

jhfrost314:
Also put some repetitive code in a separate function that calls the plotly functions to create the plots


computron:
Thanks!

PR #22, opened 2017-07-04
some cleanup & minor updates

matk86:
*  py3 compatibility in retrieve_MPDS
* remove the copy/pasted code in pycookiecheat
* update the extra_requires in setup.py, also update the versions of the required packages 
* move all optional dependencies to requirements-optional
* enable circle test with 2.7 and 3.6
* remove unittest2


computron:
Thanks!

PR #23, opened 2017-07-05
Featurizer Refactor

JFChen3:
Added descriptors to composition_features in a class-based pattern


WardLT:
@matk86 Here is the current state of our refactor.

Our new implementation of the composition features is in `composition_features_2.py`. We have kept the original `composition_features.py` for now, but plan to eventually overwrite it once we finish refactoring.

The base class for feature calculators is in `descriptors/base_classes.py`.

matk86:
thanks. merging it for now, will do the refactoring on the featurizer_refactor branch

computron:
@JFChen3 @WardLT make sure to also update the author tags for the various classes to reflect your contributions

PR #24, opened 2017-07-06
added log axis option to scatter plots (in figrecipes)

jhfrost314:


computron:
What does '-' signify? It's not in the docs yet it's the default value for axis type. Also, I don't think you should try to guess for the user what kind of axis they want. Just default to linear unless the user asks otherwise. You can still retain an option for the code to guess on the user's behalf, but the user should select that auto-decide option themselves rather than being defaulted into it.

computron:
Thanks!

PR #25, opened 2017-07-08
descriptors refactor

matk86:
* restored @JFChen3 and @WardLT 's original refactor (0aa4b0e2350ec1686e65569bc9820e47287d51a8)


WardLT:
@matk86 I haven't had a chance to review all of the changes you made. But, to get you comments on this sooner, here are a few things I think we should address before merging:

1. Updating the examples and documentation to use the new feature generator

Once we finalize this refactor, we should make sure these work with the new version.

2. Eliminating mismatch between how "AbstractFeaturizer" is used in the "composition" and "structure" features

For composition features, the class constructor takes only options for the feature generator and the materials to be "featurized" are passed in through the `featurize` method. In the structure features, the material to be "featurized" is provided in the constructor and, presumably, the signature for the `featurize` method would be different. We should decide on a single signature for the `featurize` function for both types.

I am in favor of the pattern used for the composition features. It avoids having to re-instantiate the class each time we evaluate a new material and allows for us to implement common tasks (e.g., computing features and adding them to a Dataframe) in the base class. Do you agree we should make all feature generators follow the examples in "composition"?

3. Deciding whether to "black-box" the source of elemental property data

I do like your implementation how your implementation separates Magpie and Pymatgen-based features (especially the use of singletons), but wonder whether a user needs to know where the data is coming from. Have you thought about whether we can write a "combined" data class that will check either Magpie or pymatgen for a certain elemental property? 

Jiming and I were in favor of having matminer read elemental properties from both for two reasons: (1) it could be confusing to have to know there are two sources for property data, and (2) there may be reasons to use data from both. The main drawbacks I have so far imagined are: (1) Performance [but likely small], (2) discrepancies of having two different properties with similar names to choose from. However, I think both of these problems are avoidable. Do you think we should eventually go this direction?

matk86:
1. yes, it should be done.

2. The CompositionFeaturizer in compostion.py subclasses AbstractFeaturizer and all composition related featurizers in compostion.py inherits from CompositionFeaturizer not AbstractFeaturizer. AbstractFeaturizer is something that all featurizers(composition or structure or something else in the future) should inherit. I agree, all the composition related featurizers should follow the pattern set in compositions.py. For structures, the signature for 'featurize' will be different(or at least that how it seems like). So for the time being i have left out the implementation for featurizers in the strucure.py module.

3. The way it is implemented now ensures that the property values can be obtained from either data source using a common api and in the composition featurizers the default source is set to be magpie if the user doesn't explicitly specify the source or don't really care about where the data is from. This model ensures that in the future if someone decides to use a different data source that is not already in matminer they could just add their own data source class in data.py module and the only thing they have to ensure is that their implementation subclasses the AbstractData class to ensure compatibility with the featurizers using data sources. I'm not quite sure how a black box function to the data would work when there is data available in both(could be more in the future) the sources for the same property. Also in my opinion, using the the other source of data in the case of one source missing the require property  seems inconsistent and makes it impossible to track down the source of the discrepant data.

WardLT:
Hello Kiran,

A few follow-ups,

2. How different do you imagine the signature for structure featurizers being than that for composition? My original picture was to have all of the featurizers, regardless of what sort of data they take as input, follow the usage pattern:

``` 
f = Featurizer()
# For a single material
features = f.featurize(<object describing material>)
# For many materials
features = f.featurize_all(<list of objects describing materials>)
```

The type of object used to describe a material would, of course, vary between different featurizers. Right now, given how you've refactored AbstractFeaturizer and the current implementations in structures.py, it seems like they would be used like:

```
f = Featurizer(<object describing material>)
# For a single material
features = f.featurize()
# For many materials
features = [Featurizer(m).featurize() for m in <list describing materials>]
```

Does either one of these capture your plan for the structure features? I realize you haven't implemented the `featurize` function yet, but want to make sure Jiming, you, and I agree to a similar pattern before any of us gets too far.

3. I think you make a good point about the difficulties in tracking down the source of property data if we create some scheme for merging them. Also considering your point about keeping AbstractData easy to implement for new sources, I agree with should keep it as-is.

matk86:
I havent thought much about the structurefeaturizer per se but  i intend to use the site featurizer in structure.py the second way i.e give the structure to the init and invoke featurize with the site index as argument to get that site's features

WardLT:
Do you agree that we should decide on what the `__init__` and `featurize` operations takes for inputs (i.e, between those two options in my last message) before we merge this branch?

I'm concerned that having two different patterns for using featurizers will be confusing if they are both used for a single training set. For example, it is not uncommon to use both "composition-based" and "structure-based" features to describe materials, and it would be tedious to have a different usage pattern for each type of feature. 

That said, I'm not sure how best to make the first pattern work for the site-based features. Perhaps `featurize` should take a tuple of (structure, site index)? Regardless, we can figure that out later. What I'm looking to do before we merge is gather consensus on the general pattern for using an `AbstractFeaturizer`. I would argue for adopting the first pattern (i.e., featurizer options to constructor, description of entry to `featurize`), and am interested to get your opinion.

computron:
Hi all,

First, thanks to everyone for your efforts on this. Also sorry for my delay on this - I've been catching up on things and probably won't be back to normal until next week sometime. I'm sure some of the confusion I'll refer to below is because I wasn't able to review all the changes immediately.

Second, some procedural aspects. I don't usually like to over-specify procedure and prefer to just let people work, but here is an instance where things got really disorganized. The original featurizer_refactor branch from @WardLT and @JFChen3 was a good step at least in terms of organizing changes. However, this branch got overtaken by @matk86 with severe changes that as far as I can tell were not approved or even discussed with the people originating and working in that branch. Please don't do this. If you have your own ideas and just want to work on them without the hassle of discussing them with others, that's fine, but then create your own branch for that. Otherwise it becomes very difficult for me to see the different opinions and also for the original authors to continue with their idea (I also now have yet another PR that seems to be working off the featurizer_refactor branch, but with the original scheme (?) rather than @matk86 's revised scheme - further making this situation complicated). So if you want to go solo, then do it on your own branch and don't start mixing it in with other people's work.

Finally, I haven't had a chance to look at the code in detail. Some minor aspects of @matk86 's change seem fine - e.g., rename "composition_features" to "composition". However, I am not a fan of the refactor in general. Multiple reasons:

* Take SiteAttribute() vs CompositionFeaturizer(). In the former, the data object to be operated on (structure) is in the constructor of the class. In the latter, the data object to be operated on (the composition) is not in the constructor but rather an argument of the featurize() function. This kind of inconsistency is not acceptable. I would say that the each subclass of the base featurizer should represent a particular *descriptor* (or perhaps, a set of related descriptors such as pymatgen elemental descriptors). Then the featurize() method takes in a specific data object (e.g., composition or structure) and returns that feature for that data object. This is also in-line with the original vision of having separate classes for each feature, and perhaps having some of the class attributes be things like a reference/citation for that particular feature. That is, we would *not* create a new class for each new structure or composition to process. As for the specific way in which the base featurizer should look (e.g., whether it takes in a dataframe and column name for the data object or simply just takes the appropriate data object without worrying about dataframes), I haven't thought much about it or decided anything. But the current major difference in usage/ style between different subclasses of AbstractFeaturizer is not OK.
* There is already too many hierarchies of classes and subclasses - it's a lot to take on and understand. Ideally, there is just going to be one base class for a featurizer and all descriptors are just going to subclass that master base class. I could deal with perhaps one more intermediate layer of hierarchy if it seems important, e.g., BaseFeaturizer->CompositionFeaturizer->SpecificFeature and also BaseFeaturizer->StructureFeaturizer->SpecificFeature - only if *really* needed. The current solution has something like this latter case, except that (for example) StructureFeaturizer is an intermediary layer that also has a bunch of specific features mixed into it.
* Some of the terminology is strange, e.g. SiteFeaturizer that actually returns structure features (volume per site is a structure feature, e.g. it is not different for each different site??)

I'll try to actually sort through some of this in more detail hopefully next week. But in the meantime I would like:

* The featurizer_refactor to be restored back to where it was left off by @WardLT and @JFChen3, so I can independently see whether those changes by themselves were a step in a good direction without the further changes 
* Any proposed further changes on top of that to be in a new branch (or, discussed / approved with @WardLT / @JFChen3)

matk86:
everything restored

computron:
Hi all,

Sorry it took me so long to review this PR.

First, note that the CircleCI unit tests are failing (see links in this thread). Those need to be fixed before I can do any PR. It might just be minor things like Python3 print() vs Python2 print.

For the content, I like the basic structure but would suggest a few modifications.

### BaseFeaturizer shouldn't refer to Compositions. It should be more general and should also just be an abstract class.

The methods can be:
- featurize(data) - the type of data that is received will depend on the subclass, i.e., "data" might represent a Composition or it might represent a Structure. This can raise a NotImplementedError in the base class.
- featurize_dataframe(df, colname) - this is the same thing as your featurize_all, but renamed. Again, nothing Composition specific. Here you can raise a NotImplementedError but it's better if you can implement this in a general way so subclasses don't need to implement it. Don't worry about casting object types like string to Composition. You can assume that the column has data of the correct type (e.g., we have a column in the dataframe with a bunch of Composition objects rather than string objects). If a subclass really feels like casting objects or doing something fancy is needed, they can override this method (hopefully this will not be necessary)
- credits(self) - this can just be a string that gives a citation or something, preferably as BibTeX. Again you can raise a NotImplementedError.

## Implement a few different Composition features that extend BaseFeaturizer to see how it works

e.g., an "PymatgenFeaturizer" or "MagpieFeaturizer" or "BandCenterFeaturizer", etc. The init() can set options - e.g., for PymatgenPropertyFeaturizer() the init() can set what properties you want to get back as data and also what averaging functions you want (e.g. average, standard deviation, holder mean) to apply so you get back a single float for each Composition rather than an array of floats.

(maybe/probably) combine PymatgenFeaturizer and MagpieFeaturizer into a single ElementFeaturizer where the init() again is used to control the behavior
(maybe) drop the "Featurizer" suffix from all these and come up with some shorter name for things that subclass BaseFeaturizer.

## (maybe) introduce an intermediate abstraction layer

For example, if you find that the averaging function is needed as an input for many different types of Composition features, you might introduce an intermediate class between (say) BaseFeaturizer and ElementFeaturizer that removes this duplication. But I wouldn't add this unless it *really* seems necessary.


@WardLT @JFChen3 What do you think?
- agree in principle?
- agree enough to make the changes yourself?

computron:
Also I forgot to add ``feature_labels()`` as a (not implemented) method of BaseFeaturizer - the same thing as your ``generate_labels`` but renamed

JFChen3:
Hi @computron ,

Thanks for your comments. We agree with your suggestions for BaseFeaturizer and will implement them. 

Our most recent version of the code actually has each descriptor class (ElemPropertyAttributes, ValenceOrbitalAttributes, etc.) as a subclass of BaseFeaturizer. 

For the averaging functions, we made a new singleton class called PropertyStats. (As of now, it's implemented in my fork.) What do you think of this?

computron:
@JFChen3 - OK this sounds great! I hadn't looked at your fork so might have missed your most recent changes. Your fork seems to be missing some of the changes to master so I guess make sure to pull those in. Other than that, it looks good. Looking forward to the PR!

In other news, I talked to someone who is going to work on implementing some of the various crystal structure descriptors published by the community start in beginning of Aug. Having the basefeaturizer in place will help him provide some structure to those contributions.

Also, if you are planning to submit the PR based on your fork rather than this branch, it might be worth closing this PR and opening an new one for your fork when ready.

JFChen3:
@computron Just to clarify, should I submit the PR to master or featurizer_refactor?

computron:
Submit to master when ready

WardLT:
I suppose we can close this branch now that Jiming's fork is merged into this master.

computron:
@WardLT @JFChen3 Please delete as well if you don't think it contains anything that's needed for later

PR #26, opened 2017-07-10
clarified figrecipes documentation

jhfrost314:
added an explanation that static plotting requires plotly account credentials

### DIFF
--- a/matminer/figrecipes/plotly/make_plots.py
+++ b/matminer/figrecipes/plotly/make_plots.py
@@ -2,6 +2,8 @@ from __future__ import division, unicode_literals, print_function
 
 import warnings
 
+import os.path
+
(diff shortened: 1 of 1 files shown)


albalu:
@jhfrost314 , 1) the documentation says that username and api_key are required if plot_mode=='static' but no exception is raised if they are not provided (see how ValueError is raised in the 'online' mode if either of these two are not provided.

2) Right now username and api_key are not passed even if I set these as inputs; see the following line under 'online' that is missing from 'static':
plotly.tools.set_credentials_file(username=self.username, api_key=self.api_key)

albalu:
ok thanks!

computron:
Thanks and sorry for the delay in merging this

PR #28, opened 2017-07-27
Featurizer Refactor

JFChen3:
Refactor of composition features to class-based pattern


computron:
Thanks! I think there is some things I'd like to change, but I'll pull first, change later otherwise I think the merge might be delayed.

PR #29, opened 2017-07-31
Updated implementers

JFChen3:


computron:
Thanks!

nisse3000:
Hi @JFChen3,
what is the big advantage to re-implement the feature_labels, reference, and implementors functions in each and every Featurizer class derived from BaseFeaturizer?  In my opinion, we set corresponding strings or list of strings in the constructor of some derived class and have a single implementation of those functions in the abstract class.  Much like:

class BaseClass(object):
    def __init__(self):
        self.l = None
    def labels(self):
        if self.l is None:
            raise NotImplementedError("label/s is/are not defined!")
        return self.l
class DerivedClass(BaseClass):
    def __init__(self):
        BaseClass.__init__(self)
        self.l = "My label"
if __name__ == '__main__':
    try:
        BaseClass().labels()
    except NotImplementedError:
        print('Correctly caught exception.')
    print(DerivedClass().labels())

For which I get following output:

Correctly caught exception.
My label
 
Does that make sense?  It'll greatly reduce the code length and adhere to basic inheritance principles of object oriented program design.

WardLT:
Hello @nisse3000 ,

This is an interesting proposal. For now, I'm in favor of keeping the pattern of always providing the information for `feature_labels`, `references`, and `implementators` in the function definition rather than the constructor to keep all of the featurizers with a consistent style.

We implemented each of these operations as functions so that their output could change depending on the current options for the featurizer. Generating these lists on-demand means that a user could change the options of the featurizer without re-instantiating it and prevents the feature labels from ever being out-of-sync with the current options. Of course, dynamically-generating the feature labels, implementors, or references is unnecessary for featurizers that lack options and adds a few lines of code (2?) per class. However, I worry that the advantages of the succinctness that your suggestion provides are offset by the added complexity of having some classes where the lists are implemented in the constructor and others where they are implemented in class methods.

Plus, on a minor note, I like keeping the operations in the constructor developed purely to setting the options of featurizer. Consistency concerns aside, assigning distinct functions of the featurizer to distinct class methods seems conceptually neater to me.

Are you convinced by these arguments? My concerns are almost purely aesthetic, and I don't have any technical objections to what you propose.

Thanks,
Logan

nisse3000:
Hi @WardLT,

first of all, thanks to @JFChen3 and you for all your efforts. Changing Matminers framework to feature a proper class structure is much appreciated in my opinion.

Now, back to our discussion. I don't see the argument with the options because you can simply use [function overloading](https://en.wikipedia.org/wiki/Function_overloading), which is an elegant way to deal with this.  Consider:

class BaseClass(object):
    def __init__(self):
        self.l = None
    def labels(self):
        if self.l is None:
            raise NotImplementedError("label/s is/are not defined!")
        return self.l
class DerivedClass(BaseClass):
    def __init__(self):
        BaseClass.__init__(self)
        self.l = "My label"
class DerivedClass2(BaseClass):
    def labels(self, p):
        return "Successfully overloaded labels function using parameter {} now.".format(p)
if __name__ == '__main__':
    try:
        BaseClass().labels()
    except NotImplementedError:
        print('Correctly caught exception.')
    print(DerivedClass2().labels('p'))

This yields:

Correctly caught exception.
Successfully overloaded labels function using parameter p now.

Does that make sense?

Also, I don't anticipate that neither the citation nor the implementer is likely to change depending on the option of a featurizer. Can you think of a realistic situation where that might happen? So far, all featurizer classes are based on one (list of) paper(s)---if at all.

WardLT:
Hello @nisse3000 ,

Thanks! I'm glad you find the class-based structure useful. 

Your proposal does make sense and, I agree, function overloading would certainly allow us to allow both labels defined in the constructor or in the function itself. My thought is that we should pick one of those two patterns. I would prefer that what the constructor does and does not do is consistent across all featurizers. And, if we were to pick one, I vote for always defining it in the function call as I'm not concerned about the additional 2 lines/featurizer that this requires.

That said, I don't think having two different patterns for how feature labels are generated is by any means show-stopper for matminer. 

In reference to the citation/implementor changing depending on options: One potential example is the [ElementProperty](https://github.com/hackingmaterials/matminer/blob/master/matminer/featurizers/composition.py#L115) featurizer. A few different papers vary in which statistics they compute (e.g., Holder means vs mean absolute deviation vs kurtosis), and we could reference different papers depending on which statistics a certain instance of a featurizer uses [nb: we currently don't do this, but could]. Likewise, each statistic could have been implemented by different people.

computron:
Ok let me intervene here

Both solutions are technically workable. Even in the case where different options of a featurizer lead to different citations or implemented_by strings, the new solution would still work as far as I can tell. So it is just a matter of style. I'll just list cons and skip pros for each solution.

CONS of:

Original solution (method for everything):
- verbose / lengthy. Kind of a pain to scroll around in. Requires extra typing

Proposed replacement (use instance vars):
- works in Python, but style inconsistent with other OOP languages (e.g., you can't do this in Java). A minor point but many programmers will be used to method overloading, thus the proposed replacement just feels "weird".
- less "secure", i.e., it's not clear that self.labels and self.implemented_by are private and should not be directly accessed or modified but rather should be accessed by the method. I can easily see some casual coders deciding to just directly access the instance variables rather than calling the methods that access them, which is an off-label usage and technically would not be guaranteed to be supported.
- the code itself doesn't define a clear contract of what instance variables need to be set for subclasses. Yes, you can write this in the documentation, but this is not as clear a contract as having the methods to be overloaded as actual base class code. For example, when drawing class diagrams the original solution already makes it clear what is expected to be done from subclasses. You couldn't infer from a class diagram what are the specific instance variables that need to be overloaded in the new proposed solution.
- can lead to messy / disorganized code. In the proposal, the init() of the subclass would be responsible for doing multiple things - i.e., setting a bunch of instance vars that relate to references, a big string with credits, an array with implementers, etc. as well as setting up the parameters of the class itself. For example, some coders might define the "implemented_by" right at the beginning of the init(). Others might put it at the end. It's cleaner to have the code related to implementors right near the implemented_by() method. Or take labels() - in some cases, this is multiple lines of code. It's better if those lines of code that are just are part of the feature_labels() function instead of in some random place in __init__(). 
- (not really a big deal IMO) - *requires* storing implemented_by() and labels() when there is often no need to store these.

Basically, I understand that having many methods feels like extra typing / reading. But there are style reasons to have the extra baggage in there

PR #30, opened 2017-08-10
Added ability to compute AGNI fingerprints

WardLT:
Implemented the AGNI fingerprints developed by [Botu *et al*](http://doi.wiley.com/10.1002/qua.24836).

The bigger thing that should be reviewed is changes to `BaseFeaturizer`. I needed to be able to pass both a structure and site to the new featurizer, so I modified the `featurizer`  method to take a variable number of arguments and the `featurize_dataframe` operation to allow you to generate features based on more than one column. I made sure to do this in a way that did not require modifying any of the other featurizer classes. 

What I'm unsure of is whether this additional flexibility is helpful or would open the door to too much inconsistency between the featurizers. Any thoughts?

### DIFF
--- a/matminer/featurizers/atomic_site.py
+++ b/matminer/featurizers/atomic_site.py
@@ -1,0 +1,121 @@
+"""Features that describe the local environment of a single atom
+
+The `featurize` function takes two arguments:
+    strc (Structure): Object representing the structure containing the site of interest
+    site (int): Index of the site to be featurized
+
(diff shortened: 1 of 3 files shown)


computron:
Thanks! I am ok with the changes to BaseFeaturizer. I will probably just rename "atomic_site.py" to "site.py" since that is more consistent with the pymatgen lingo

PR #33, opened 2017-08-19
Sine Matrix, OFM, LFS datasets, and stats.py additions

kylebystrom:
Hi Anubhav,

These are the updates I made for Codebusters. I ran into some roadblocks getting everything pushed to my fork because of some mistakes on my part and because the LFS was acting a bit funky. I'm pretty sure it's all good now, and Github says everything is compatible for auto-merge, but let me know if something goes wrong and I'll look into it!

Thanks again for the opportunity to do some work on matminer; it's been a lot of fun and a good learning experience. As mentioned, I'll have a couple of my new descriptors and possibly the Ewald matrix ready in a couple weeks if you're interested! I might also be able to do a couple other structure descriptors that Haoyan showed me, depending on how things go.

Cheers,
Kyle


computron:
Thanks! Looks good. I will merge and do a little bit of housecleaning afterward.

It'll be good to connect again after you finish your current project with Mark. Hopefully we'll have some more progress on matminer by then.

PR #40, opened 2017-09-12
SiteOrderParameters class and n_numerical_modes updated

albalu:
modified SitesOrderParameters to conveniently work with a DataFrame (similar to ElementProperty for example) taking in features, and stats as inputs. Note that SiteOrderParameters works just like before if stats is None (i.e. it would return calculated order parameters for each site in a list). It is also backwards-compatible as none of the arguments (including data_source) is required.

In the process, I had to also change PropertyStats.n_numerical_modes method as it had a couple of issues:

1) it was returning error for an example input of [6.0, 6.0] for which clearly the first mode should be 6.0

2) It wasn't handling the cases where the nth mode doesn't exist and would return lists with unexpected lengths: I added conditions so now it returns nan for each mode that is not found.

### DIFF
--- a/matminer/featurizers/stats.py
+++ b/matminer/featurizers/stats.py
@@ -134,18 +134,21 @@ class PropertyStats(object):
     @staticmethod
     def n_numerical_modes(data_lst, n, dl=0.1):
         """
-        Determines the n first modes of a data set that are obtained with
-        a finite bin size for the underlying frequency distribution.
+        Returns the n first modes of a data set that are obtained with
+            a finite bin size for the underlying frequency distribution.
         Args:
-            data_lst (list of floats): data values.
+            data_lst ([float]): data values.
             n (integer): number of most frequent elements to be determined.
             dl (float): bin size of underlying (coarsened) distribution.
-        Returns: 
-            modes (list of float): first n most frequent elements.
+        Returns:
+            ([float]): first n most frequent entries (or nan if not found).
         """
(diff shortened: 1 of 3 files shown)


albalu:
I can add tests with various stats once this structure is discussed and approved.

computron:
Thanks! Why is there a ``data_source``? If you can't think of any real implementation for ``data source`` other than pymatgen that will be on the horizon, I would just remove it and explicitly code pymatgen. This will result in clearer code and one less parameter to worry about.

albalu:
1) I was thinking of other implementations of site-dependent order parameters that may be available in the future? But I guess a separate class can be defined for those?

2) I thought all the featurizers to require (or optional) similar inputs; the usage is then easier to remember. All data_source args can be default to 'matminer'. For example, the next time I want to use a featurizer for the first time, I know that the first things I look for are what are the types of data_source available and what features are provided by each of them.

computron:
1. yes let's go with a separate class for those. Let's stick to Nils's order parameters in this class

2. Nope, no "knobs" are required. You might be looking specifically at the ElementProperty descriptor but that one has 3 different data sources from day 1.

You should just remove the data_source parameter. It's easy to add this back later if we need it, I don't think we will though.

computron:
Thanks! Is it ready to merge? If yes I will merge after tests pass

albalu:
Sure, Nils and I will add some tests that involve stats later.

PR #41, opened 2017-09-15
Start a simple band structure featurizer and a unittest for it

albalu:
this BandFeaturizer extract a few basic features from a pymatgen band structure object (or dict) regardless of whether it is in line_mode or not (maybe in future, features would defer based on line_mode). Although, it does not need a Structure in its current shape it will return additional features such as band degeneracy if bs.structure

A simple unittest is also added.


albalu:
I need help here. I am not sure how to provide API_KEY for circleci.

computron:
It's better to not use MPRester in matminer unit tests in the first place.  Instead serialize the structure and band structure (e.g. to JSON file in the tests dir) and load from there. Otherwise, if MP goes down then the matminer unit tests will start failing, which is not what we want (since matminer is actually still fine).

As an aside, I actually don't know the best way to set it, would need to check the CircleCI docs for instructions on either setting an environment var or executing a command (e.g., "cat PMG_MAPI_KEY=xyz > ~/.pmgrc.yaml") before running the test

albalu:
Thanks Anubhav. I included Si structure and band structure in the tests folder. Hopefully it'll be useful for other tests as well.

computron:
Thanks!

Also why are you operating on the dict rather than the object?

- dict format is more subject to change than object API
- even if object API changes, it can be caught by refactoring (or by warnings issued in pre-release of change) and matminer can be updated
- generally it is considered more proper to be working with the object functions

computron:
Also need to implement nband and nextremum (currently don't do anything?)

computron:
Also why is the extremum seem to only be for valence band (e.g. see docstring) and not conduction band?

computron:
nevermind the previous question about extremum, looks like there is both "n" and "p"

albalu:
Hi Anubhav, I somehow found it harder to work with BS object as it keeps switching to method, property, dict etc for each property that I wanted to access while dict is consistent. I reverted the implementation to operate on the bs object anyway and got rid of the unused arguments such as nband though I will be using them soon but in a separate future PR.

albalu:
BTW I found a bug in the process so thanks!

PR #43, opened 2017-09-19
Bug: featurize_dataframe fails when only one feature

WardLT:
Turns out the `featurize_dataframe` operation fails when the class only generates a single feature. 

The problem was with how numpy.array works. If there is more than 1 feature, this line in the code creates a 2D array. If there is only 1 feature, it creates a 1D array. I fixed it by adding code to manually change 1D->2D. (Note: the ndmin kwarg of np.array puts the new dimension on the wrong side)

Also added test case that fails before this fix was added, and should now pass


computron:
Thanks!

PR #44, opened 2017-09-20
Orbital Field Matrix Update

kylebystrom:
Small pull request for the updated orbital field matrix based on my summer project, which takes into account the periods of elements.  I haven't added myself to the citations for this method because I have nothing to cite, but if I'm lucky I might be able to get the improvements into an undergrad journal and add a citation then.

I will start the MBTR wrapper next week.

### DIFF
--- a/matminer/featurizers/structure.py
+++ b/matminer/featurizers/structure.py
@@ -473,10 +473,10 @@ class OrbitalFieldMatrix(BaseFeaturizer):
         period_tag (bool): In the original OFM, an element is represented
                 by a vector of length 32, where each element is 1 or 0,
                 which represents the valence subshell of the element.
-                With period_tag=True (default), the vector size is increased
+                With period_tag=True, the vector size is increased
                 to 39, where the 7 extra elements represent the period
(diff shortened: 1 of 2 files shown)


computron:
Thanks! Can you just set the default period_tag to False? Then the default implementation will be in-line w/the original orbital field matrix paper.

kylebystrom:
Sure thing. I noticed that both Coulomb matrices also have a default setting that is different from the paper they are based on, should I fix that too?

computron:
Hi Kyle, I think it would be nice for default settings to match their respective papers globally. If there is a particular combination of parameters (or even a single parameter choice) that you think matminer should advocate and that is different than the paper defaults, feel free to add a from_preset() function like in the ElementProperty featurizer. Then one can do something like "OrbitalFieldMatrix.from_preset("matminer")" and get your recommended parameter settings

kylebystrom:
All fixed! Looks like a few checks are queued, but should be good other than that.

computron:
Thanks!

PR #47, opened 2017-10-12
 Added OPSiteFingerprint class and unit tests. 

nisse3000:


computron:
Will wait for new PMG version before merging

computron:
Thanks! Merging

I will just make one change for now after the merge is fix the mutable default arg for stats. I might have other changes later when I get around to doing a more thorough code review of matminer

PR #48, opened 2017-10-17
Ironed out a small flaw in …

nisse3000:
…the neighbor-finding scheme of OPSiteFingerprint and using tuples for default stats in OPStructureFingerprint now.


computron:
@nisse3000 you should resolve the conflict since I already did the tuple conversion

PR #49, opened 2017-10-19
added PDOS featureizer

dyllamt:
I added a DOSFeaturizer class under the bandstructure module, recycling Anubhavs code that extracts PDOS information from a pymatgen dos object. The main features for VB and CB are:

maximum fractional contribution from one orbital: tells you something about amount of hybridization
orbital character from the main contributing orbital: (s, p, or d)


computron:
Hi @dyllamt , some comments to get started

1. Needs unit tests.
2. Pretty sure that you need a CompleteDos object, not a DOS object (update docs as well as Dos.from_dict)
3. Return a string rather than one-hot encoding. e.g. "s" not [1,0,0]. It's easy to turn a categorical string into a one-hot encoding with sklearn.
4. Option to include element, not just orbital? e.g. "Na:s" rather than just "s". Note that implementing the previous change will help with this.
5. Documentation could be better. e.g., band depth is a float in eV? Also this is more like an energy_cutoff than a band depth. Also why get rid of the Fermi smearing way?
6. Probably want an option to return top N contributors rather than just the "main" contributor. For example, if a contribution is 40% Ti:d and 39% O:p and 21% other stuff, it's usually good to know information about the second place contributor. This comment is not strictly necessary though.

dyllamt:
Hi @computron ,

I'm working on some of these comments, including the unit tests, specification of the correct dos object, and the string vs one-hot encoding.

I got rid of the Fermi smearing because I felt that similar functionality is obtained by changing the energy_cutoff. Also, Fermi smearing is going to always only allow you to sample the band edge (less variability than just implementing an energy cutoff).

I have a question about returning a variable number of contributors. Would it be a problem for sklearn if I return only contributors that score (for example) 10% or higher. There may not be the same number of contributors returned every time with this option.

computron:
For returning contributors 10% or higher: I wouldn't suggest having a variable number of features (i.e, it's fine if up-front the Featurizer allows the user to choose the top N options and always returns N options, but not for it to return <N or >N options in that case).

Basically two reasons:

* It might mess up some technical things like the featurize_dataframe() method which featurizes every row and adds it the dataframe. If different rows have different numbers of features returned, it could cause a problem. These technical things could be fixed (if they are a problem in the first place) with some effort though.
* We will need each row to contain the same number of features for scikit-learn at the end. We can certainly fill any missing values with NaN to make it work out if different rows have different numbers of features. However, It is less desirable to have (say) the 4th feature be "contribution of 4th highest orbital if it's >10% and float.NaN otherwise" rather than just "contribution of 4th highest orbital" which is a more continuous measure. Thus, since you need the same number of features in the end, it's good to just have it be that way since the beginning and use a smoother / more consistent definition of that feature.

Thus I'd suggest just letting the user choose the number N in the top N features regardless of any hard cutoff like 10%.

For Fermi smearing, certainly implement how you'd like, but note that hard cutoffs can lead to sharper discontinuities in feature values as a function of that cutoff.

dyllamt:
Made updates to DOSFeaturizer:

-number of features for vbm, cbm is an integer number of orbitals, specified by the user
-additional feature added that reports the number of orbitals with contributions above a user specified threshold (e.g. the number of orbitals with contributions to the DOS above 10%)
-added unit tests

computron:
Thanks! I am going to merge and make some adjustments

PR #50, opened 2017-10-20
include CBM/VBM degeneracy as band features + test

albalu:
I am not sure the best location for isolated remove_duplicate_kpoints and get_k_degen are so I just included them at the top of the file.

TODO: maybe have the option to lump points that slightly deviate from the high symmetry points together to avoid getting artificially high degeneracy such as 24!

### DIFF
--- a/matminer/featurizers/bandstructure.py
+++ b/matminer/featurizers/bandstructure.py
@@ -22,7 +22,7 @@ def remove_duplicate_kpoints(kpts, dk=0.001):
     """
     rm_list = []
     kdist = [norm(k) for k in kpts]
-    ktuple = zip(kdist, kpts)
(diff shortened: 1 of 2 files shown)


computron:
Some preliminary comments

1. For periodic distance checks, probably copy the logic in PeriodicSite.is_periodic_image() which has a pbc_diff method. Note that your method won't tell you that 0.5 and -0.4999 are the same point.

2. code is kind of sloppy and unintuitive. e.g., get_k_degen shouldn't have to call a method that looks like it's removing k points.

3. Getting the k-point degeneracy should probably go somewhere in pymatgen, e.g., the BandStructure class

albalu:
ok I will implement these points and open a new PR.

Issue #46, opened 2017-09-28
PropertyStats should include a "reduced mass" type average


computron:
i.e., a parallel addition type term

1 / avg = 1 / x1 + 1 / x2 + 1 / x3 ...

Sometimes the "reduced mass" of a system can be more informative than an average mass for example

WardLT:
Is this just the holder mean with n == -1? If so, we do already support this.

computron:
The holder mean with n = -1 is pretty close, it is just the number of samples times what I originally suggested.

e.g. the reduced mass of a 2-body system with m1=2 and m2=2 is 1, whereas the holder mean is 2 (twice of 1, since we have 2 samples). I think the holder mean with n^-1 is closer to what I want since it generalizes better to >2 entries.

PR #58, opened 2017-10-25
Added geometric std. dev. and inverse mean

WardLT:
I added another set of statistics to the "Property Stats" class. I have a few questions before we would merge this:

1. Regarding #46: Is it sufficient to add in a special calculator for the inverse mean? 
1. Regarding @computron 's comment in the header: @kylebystrom, any thoughts on where I should put the kernel operations? They don't appear to be used elsewhere in the code, so my first thought was to delete them for now.
1. @computron : Is there any other cleaning you'd like to see done? I don't know the normal standards/procedure, but am happy to make the changes.

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -662,9 +662,13 @@ class TMetalFraction(BaseFeaturizer):
 
 
 # TODO: why is this a "feature" of a compound? Seems more like a pymatgen analysis thing?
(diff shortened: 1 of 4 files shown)


computron:
1. I don't think we need the inverse mean at all, I think the Holder means are good enough as generalized means. I am going to merge but then delete it, feel free to restore if you think it's useful rather than just noise.
2. Let's start a new issue on that rather than mixing with this one.
3. For the cleaning / etc., just search for the tag "@WardLT" in the code. At a later point I might translate those to issues as well.

PR #59, opened 2017-10-25
Refactor of Citrine data retrieval tool

saurabh02:
i) for one record, compressed multiple rows into one row
ii) fixed two bugs: a value of type=list with more than one item gets inserted into as one cell in the dataframe instead across multiple rows; and multiple properties with same name don't get overwritten in the dataframe.
iii) added a unit test


computron:
Thanks!

PR #60, opened 2017-10-28
A FigRecipes and corresponding example Fix

ardunn:
- Bulk modulus example ranked feature importances out of order. Fixed
- Previous fix for PlotlyFig caused issues with bar plot used in bulkmod example. Fixed again, works with bar plot now
- Recompiled docs w/minor edit to example

*PS if it is less annoying to just automatically merge these minor docs/FigRecipes changes I can do that from now on*


computron:
Thanks! If there are no functional changes, just doc changes, feel free to just merge yourself.

ardunn:
Sounds good. This one had some edits to FigRecipes (which I noticed when fixing the doc example), but they were very minor.

PR #64, opened 2017-11-02
add CBM/VBM degeneracy to BandFeaturizer

albalu:
added CBM/VBM degeneracy as features of Bandstructure if bs.structure is available; None otherwise. Also added test for Si.


computron:
Thanks! Note you forgot to update docs but I'll just do it after merging since it's simple

PR #65, opened 2017-11-06
Fixed constructor for AGNIFingerprints

WardLT:
Problem with using 'x or y' syntax for setting self.etas. If the user passes a numpy array, this results in a `ValueError` about the truth value of 'or' being undefined for arrays. I suspect that this is because
numpy arrays overload the function __or__. Either way, this fix changes it.

I also added a test case that fails with the old version and passes with this new one.


computron:
Thanks!

PR #70, opened 2017-11-16
add Miedema model and Voronoi indices

Qi-max:
Add Miedema model (for binary alloys) and Voronoi indices as composition and site descriptors, respectively.

The following files have been modified:

added a data_file of Miedema.csv, containing parameters of the Miedema model;
composition.py and test_composition.py
site.py and test_site.py


computron:
Thanks! Some notes for next time:

* you should try to keep PRs as short as possible. In this case, it's best if you separate the Voronoi index and Miedema model as two different PRs.
* you should not to mix formatting changes and code changes in the same PR. If you want to use auto-format, you should select your own code and auto-format that only so the PR doesn't include changes to all the various functions.  If you would like to auto-format one or more full files, then you should submit a PR for which the only changes are auto-formatting (not mixing formatting changes with function additions / deletions / etc)

Qi-max:
OK, get! Thanks for the helpful suggestions!

PR #71, opened 2017-11-17
add the possibility of including the energy of a given list of k-points normalized to the CBM/VBM as features.

albalu:
1) add normalized energy at given k-points as features (the fractional coordinates of the k-points should be given by the user, if any 
2) restructure featurize to make it more compact
3) store features in a dict to make it easier to implement apply_weight method; apply_weight is not run by default (if weight==1); however. I found it useful in identifying band structures that are more similar to each other based on the Euclidean distance of BS feature vector.

### DIFF
--- a/matminer/featurizers/bandstructure.py
+++ b/matminer/featurizers/bandstructure.py
@@ -138,7 +138,7 @@ class BandFeaturizer(BaseFeaturizer):
                 is_gap_direct (0.0|1.0): whether the band gap is direct or not
                 direct_gap (eV): the minimum direct distance of the last
                     valence band and the first conduction band
-                p_ex1_norm (flaot): k-space distance between Gamma point
+                p_ex1_norm (float): k-space distance between Gamma point
                     and k-point of VBM
                 n_ex1_norm (float): k-space distance between Gamma point
                     and k-point of CBM
--- a/matminer/featurizers/bandstructure.py
+++ b/matminer/featurizers/bandstructure.py
@@ -190,7 +190,7 @@ class BandFeaturizer(BaseFeaturizer):
                 self.feat['{}_ex1_degen'] = float('NaN')
         if self.weight > 1:
             self.apply_weights(self.weight)
-        return self.feat.values()
+        return list(self.feat.values())
 
     def apply_weights(self, weight):
         """
@@ -206,7 +206,7 @@ class BandFeaturizer(BaseFeaturizer):
             self.feat[key] *= max(1, weight/2.)
 
     def feature_labels(self):
-        return self.feat.keys()
+        return list(self.feat.keys())
 
     @staticmethod
     def get_bindex_bspin(extremum, is_cbm):
--- a/matminer/featurizers/bandstructure.py
+++ b/matminer/featurizers/bandstructure.py
@@ -108,8 +108,12 @@ class BandFeaturizer(BaseFeaturizer):
     Featurizes a pymatgen band structure object.
     """
 
-    def __init__(self):
-        pass
+    def __init__(self, kpoints=None, atol=0.01, include_symeqks=False,
+                 weight=1):
+        self.atol = atol
+        self.include_symeqks = include_symeqks
+        self.kpoints = kpoints
+        self.weight = weight
 
     def featurize(self, bs):
         """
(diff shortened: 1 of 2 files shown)


matminer/featurizers/bandstructure.py
```diff
@@ -108,15 +109,28 @@ class BandFeaturizer(BaseFeaturizer):
...
                 The band structure to featurize. To obtain all features, bs
                 should include the structure attribute.
+            kpoints ([1x3 numpy array]): list of fractional coordinates of
+                k-points at which energy is extracted.
+            atol (float): the absolute tolerance for fractional coordinate
+                below which two coordinates are considered the same
+            include_symeqks (bool): if True, when looking for energy of the
+                bands at given kpoints, symmetrically equivalent kpoints are
+                also considered (recommended to keep False for speed)
+            weight (int >= 1): if larger than 1, some features are multiplied
```
computron: remove this parameter; the goal of the featurizer should be to produce physically relevant features, not to balance the weight of various features. If someone is designing a distance metric or needs to normalize data a certain way, they should do it *after* running the featurizer. There are many ways to weight / normalize data and it is not the job of featurizers to do that.

Note that I will at some point start a "distance_metrics" package. There, if you want to define some default weights for judging band structure distance, you can do so.

matminer/featurizers/bandstructure.py
```diff
@@ -108,15 +109,28 @@ class BandFeaturizer(BaseFeaturizer):
...
         """
         Args:
             bs (pymatgen BandStructure or BandStructureSymmLine or their dict):
                 The band structure to featurize. To obtain all features, bs
                 should include the structure attribute.
+            kpoints ([1x3 numpy array]): list of fractional coordinates of
+                k-points at which energy is extracted.
+            atol (float): the absolute tolerance for fractional coordinate
+                below which two coordinates are considered the same
+            include_symeqks (bool): if True, when looking for energy of the
```
computron: it is unclear to me from the documentation what is the purpose of setting this to True. Do I get different results? If so, for which features?

matminer/featurizers/bandstructure.py
albalu: ok I will remove it and add it later to distance_metrics

matminer/featurizers/bandstructure.py
albalu: This is for when a given kpoint is not available in the input BandStructure object but its symmetrically equivalent one is; so instead of returning nan we find that symmetrically equivalent kpoint and return its energy. I thought this is what you wanted? It will be more useful if the input BandStructure is in uniform kmesh.

computron: Btw, if the purpose of this is to automatically automatically add more k-points if there are symmetrically equivalent kpoints to the ones the user provides (e.g., the user provides [0,0,0.5] and the code automatically adds [0.5, 0, 0] and [0, 0.5, 0] so you get features like n_0.0;0.0;0.5_en, n_0.0;0.5;0.0_en:, n_0.5;0.0;0.0_en:), I would suggest to remove this parameter. Typically one wants to know how many features they are getting out in advance so that the number of features will match for different materials.  Instead, I could see there being a use for a separate featurize where you provide a list of user k-points and it returns a list of weights for those kpoints.

albalu: no it does not provide additional unwanted features it just looks among all symmetrically equivalent k-points to find the energy of the given k-point.

computron: Oh I see, the documentation doesn't say that at all. In that case, there should be a parameter called ``use_closest_k`` and for which the documentation says:

"If True, will substitute data from the closest available k-point if a k-point specified by the user is not present in the band structure".

Is this the correct description? If so, not sure why the recommendation is for that to be False.

* If all the specified k-points are already there in the band structure, then this parameter should not affect speed if implemented properly.
* If all the specified kpoints are not already there in the band structure, then this parameter becomes necessary to produce data

computron:
(let me know when finished with revisions are ready to merge)

albalu:
ok it should be ready.

computron:
Thanks!

PR #72, opened 2017-11-17
Update code for newest version of Citrine's API; bug fix for limiting number of records when total number of records desired is > 100

saurabh02:


computron:
Thanks!

PR #73, opened 2017-11-17
Add ignore_errors param to BaseFeaturizer

ardunn:
Adds an ignore_errors parameter to BaseFeaturizer. The rationale for this is that if you have a huge dataframe which is being featurized, and a few materials throw difficult errors, you have the option to just set those rows to NaN (to filer out later) rather than try to fix unusual exceptions in pymatgen and have to re-featurize the your entire dataframe.

Also:

- (very) minor fixes to PlotlyFig


albalu:
+1 for ignore_errors with default of False; it's practical.

computron:
You should use float("nan") as default values, this is the generally the standard for missing data in pandas. I will pull and change

computron:
(note that np.nan should also be ok)

PR #75, opened 2017-11-21
Added general similarity computation function to base.py

nisse3000:


WardLT:
Hello @nisse3000 ,

I like this idea of adding an easy interface to using similarity metrics to the base class. It would simplify the use case of people just wanting to compare two materials. A few thoughts/questions related to improving the interface:

1. I find the use of varargs a little confusing. Am I correct in assuming that `x` can either be two or three values? The confusing part is that sometimes `x` is only the values to be compared, and sometimes it includes options. If so, maybe breaking them apart using kwargs would be clearer (e.g., `get_similarity(self, simtype, x1, x2, **kwargs)`). Here, `order` could be one of the kwargs that is used only by Minkowski distances.
2. Isn't the use of `int(...)` in the unit tests redundant with `assertAlmostEqual`? `assertAlmostEqual` lets you set the number of places to round to. Adjusting the tolerance for the equality test is what you're effectively doing by multiplying the value by 1000, right?
3. Is this function a better home for the pairwise-similarity metrics that are currently in PropertyStats? If so, should we move those metrics to `get_similarity`?
4. Would it make sense to have the code accept names of metrics from (sklearn's pairwise module)[http://scikit-learn.org/stable/modules/metrics.html#metrics]?

computron:
Hi all,

Thanks for getting this started, but I think this requires a redesign. Not all featurizers should be distance metrics, and not all distance metrics should be featurizers. e.g.

- a featurizer that returns a class as string (e.g., main orbital component of VBM like "Fe_d")  wouldn't work with a distance function at all. 
- Featurizers that return a single float aren't really great as distance metrics (although they should technically "work")
- Some distance metrics have no good featurizer counterpart. For example, let's say someone wants to define the distance between two Compositions as the percent of elements they have in common. That's a valid (if simplistic) distance metric, but there's no "featurizer" equivalent.

Thus, these two interfaces should be kept separate. I'll think about it more but my current suggestion is:

- keep base featurizer as-is
- add a new distance metric base class
- add a new class called DistanceFeaturizer (or something) that extends both base featurizer and base distance metric, and can contain some kind of built-in automatic distance computation.

I also haven't looked much at the details of the distance functions. In particular, there's a lot in scipy (https://docs.scipy.org/doc/scipy/reference/spatial.distance.html) and I would likely advocate just leveraging that.

computron:
Hi @nisse3000 - let me know when you are done making revisions to this branch and I will review when you are finished.

computron:
@nisse3000 I think you should probably kill this PR, but I do think you should submit a separate PR (after getting the code versions synced) with the ChemEnv stuff (and not the distance stuff) in the short term.

PR #80, opened 2017-12-09
restructuring DosFeaturizer 

albalu:
restructuring DosFeaturizer  so that it returns single value/str and consistent number of features rather than lists and returns the same number of features regardless of the input DOS. Also modified the tests accordingly and try to convert columns to float in the base featurizer as the line features = np.array(features) will automatically convert everything to unicode when we have a mixture of floats and str in the initial features.

### DIFF
--- a/matminer/featurizers/tests/test_bandstructure.py
+++ b/matminer/featurizers/tests/test_bandstructure.py
@@ -127,7 +127,6 @@ class DOSFeaturesTest(PymatgenTest):
 
     def test_DOSFeaturizer(self):
         df_df = DOSFeaturizer(contributors=2).featurize_dataframe(self.df, col_id=['dos'])
-        df_df.to_csv('test_DOS_featurized.csv')
         # CBM:
         self.assertAlmostEqual(df_df['cbm_score_1'][0], 0.258, 3)
         self.assertAlmostEqual(df_df['cbm_score_2'][0], 0.258, 3)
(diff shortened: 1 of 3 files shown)


matminer/featurizers/tests/test_bandstructure.py
```diff
@@ -126,20 +126,23 @@ def setUp(self):
...
-        self.assertEqual(df_df['cbm_significant_contributors'][0], 4)
-        self.assertAlmostEqual(df_df['vbm_percents'][0][0], 0.490, 3)
-        self.assertEqual(df_df['vbm_locations'][0], [[0.0, 0.0, 0.0]])
-        self.assertEqual(df_df['vbm_species'][0], ['Si'])
-        self.assertEqual(df_df['vbm_characters'][0], ['p'])
-        self.assertEqual(df_df['vbm_coordinations'][0], ['tet'])
-        self.assertEqual(df_df['vbm_significant_contributors'][0], 2)
+        df_df = DOSFeaturizer(contributors=2).featurize_dataframe(self.df, col_id=['dos'])
+        # CBM:
+        self.assertAlmostEqual(df_df['cbm_score_1'][0], 0.258, 3)
```
computron: Why are all these values still arrays?

I expect that after calling featurize_dataframe(), each column has a single float or str in it.

e.g. df_df['cbm_score_1'] = 0.258 not df_df['cbm_score_1'][0] = 0.258

Same with all the others. Note that if the featurizer is trying to get the top N contributions, then it the number of feature columns should increase (keeping 1 float or str per feature)

computron:
Thanks!

PR #86, opened 2017-12-15
Included ChemEnvFingerprint

nisse3000:


matminer/featurizers/tests/test_site.py
```diff
@@ -106,6 +106,26 @@ def test_op_site_fingerprint(self):
         self.assertAlmostEqual(int(1000 * ops[opsf.feature_labels().index(
             'bcc CN_8')] + 0.5), 955)
 
+    def test_chemenv_site_fingerprint(self):
+        cefp = ChemEnvSiteFingerprint.from_preset('multi_weights')
```
computron: Can you change the multi_weights test to return something different from the simple test? e.g., a structure for which the local env is not perfect. Otherwise, the test cannot distinguish whether the multi-weight strategy is doing the right thing.

One simple way to do this is to distort the CsCl structure and then make sure that the simple strategy still gives 1 but the multi strategy gives a value between, say, 0.8 and 1

computron: (but not exactly 1 for the latter)

computron: a better way would be to actually add a distorted structure to the test files. This is probably going to come in handy for testing the other site/structure fingerprints as well.

computron:
I am just going to pull this for now since I'm doing the matminer release and add an issue for revising the test

PR #95, opened 2017-12-19
extend Miedema model to ternaries and more

Qi-max:
Main changes:
-extend Miedema model to ternaries and more (it is reported that this extension works well for ternaries, but should be taken carefully for quaternary alloys and more; have included a comment for this message)
-fix the read datafile problem
-add test for returning nans if containing elements not in the model (the model contain parameters for 73 elements)
-refine the code structure and comments


computron:
Thanks!

Issue #84, opened 2017-12-14
some remaining items for DosFeaturizer


computron:
@dyllamt @albalu 

* move to featurizers/dos.py (instead of bandstructure.py)
* the xbm_location_i should probably return fractional, not Cartesian coords. Can't think of any good reason for it to be Cartesian
* the coordination variable in ``get_cbm_vbm_scores`` is way too hacky. Only works for tet vs oct, uses a custom function, etc. Remove this var. If someone want this function they should code it well (works for multiple environments, etc), or keep it inside their own test code. Note that there is another issue open for a featurizer that takes in a site and returns back a string label for the coordination number of that site.
* get_cbm_vbm_scores() should explain in the docs what it's doing

albalu:
@dyllamt , if you can add the doc to get_cbm_vbm_scores(), I can take care of the rest.

dyllamt:
@albalu, that works for me. Maybe we can discuss more during our meeting today?

dyllamt:
Remaining Issues:
- features stored in unordered, standard dictionary
- class attributes are initialized in the featurize() function
- feature_labels() can’t be called until after featurize() has been called

I'll work on fixing these in the next week. BandFeaturizer currently has the same issues.

computron:
(this is completed)

PR #96, opened 2017-12-19
moving DosFeaturizer to dos.py and more

albalu:
I addressed the comments in Issue #84 including moving DosFeaturizer to dos.py, removing coordination feature for now until we have a robust code for returning the CN or string given a site and finally add more docs to get_cbm_vbm_scores.


computron:
Thanks!

PR #98, opened 2017-12-22
AtomicOrbitals featurizer

dyllamt:
@computron 

I implemented an AtomicOrbitals composition featurizer that outputs the HOMO, LUMO, and gap of a given composition. List of additions below.

added Class AtomicOrbitals
- returns homo and lumo: character, element, and energy
- returns gap

added test for AtomicOrbitals


matminer/featurizers/composition.py
```diff
@@ -157,6 +158,54 @@ def implementors(self):
...
 
+class AtomicOrbitals(BaseFeaturizer):
+    '''
+    class to determine the highest occupied molecular orbital (HOMO) and
+    lowest unocupied molecular orbital LUMO in a composition. The atomic 
+    orbital energies of neutral ions with LDA DFT were computed by NIST.
+    https://www.nist.gov/pml/data/atomic-reference-data-electronic-structure-calculations
+    '''
+
+    def __init__(self):
```
computron: don't need this

matminer/featurizers/composition.py
```diff
@@ -157,6 +158,54 @@ def implementors(self):
...
+            LUMO_character: (str)
+            LUMO_element: (Element)
+            LUMO_energy: (float in eV)
+            gap: (float in eV)
+                the estimated bandgap from HOMO and LUMO energeis
+        '''
+
+        string_comp = comp.reduced_formula
+
+        HOMO_LUMO = MolecularOrbitals(string_comp).band_edges
```
computron: look up PEP8 for variable names

rename to ``homo_lumo``

matminer/featurizers/composition.py
```diff
@@ -157,6 +158,54 @@ def implementors(self):
...
+
+        self.feat = {}
```
computron: don't use a plain dict for this. The order of keys() and values() can change in a dict. It will be consistent 99% of the time so you won't typically notice it, but it *does* change.

You can either use two lists (one for the feature names, one for the feature values) or you can use an OrderedDict.

As it stands, the code will potentially return the keys and values in different orders leading to bad values

matminer/featurizers/composition.py
```diff
@@ -157,6 +158,54 @@ def implementors(self):
...
+        self.feat['gap'] = self.feat['LUMO_energy'] - self.feat['HOMO_energy']
+
+        return list(self.feat.values())
+
+    def feature_labels(self):
+        return list(self.feat.keys())
+
+    def implementors(self):
+        return ['Maxwell Dylla']
+
```
computron: add a citations() method with the NIST data

matminer/featurizers/tests/test_composition.py
```diff
@@ -101,6 +102,16 @@ def test_fere_corr(self):
         self.assertAlmostEqual(df_fere_corr["mean FERE correction"][0], 0.077146274)
         self.assertAlmostEqual(df_fere_corr["std_dev FERE correction"][0], 0.270209766)
 
+    def test_atomic_orbitals(self):
+        df_atomic_orbitals = AtomicOrbitals().featurize_dataframe(self.df, col_id="composition")
+        self.assertEqual(df_atomic_orbitals['HOMO_character'][0], 'd')
+        self.assertEqual(df_atomic_orbitals['HOMO_element'][0], Element('Fe'))
```
computron: don't return Element objects in the featurizer. Featurizers should either return floats or strings. The various scikit-learn machine learning algorithms tend to operate on floats or strings, but none of them can operate on an Element object

computron:
Thanks! I went through and did a quick code review

matminer/featurizers/composition.py
computron: ( referring to the empty __init__ method, not the reference to NIST - you can just remove the __init__ method altogether)

dyllamt:
@computron 

I went through and made the short list of changes, which you recommended.

computron:
Thanks!

computron:
@dyllamt  Note: after pulling this commit, I realize you set self.feat inside the featurize() function. This means that one can't get the feature labels until first calling featurize(), which is not the intended behavior. I pushed a commit afterwards to fix this so one can call feature_labels() at any time after initializing the object.

Note that, under normal circumstances, the only time you should be creating class variable like "self.X" should be in the ``__init__()`` function.

dyllamt:
@computron 

Thanks for making this fix. I looked at the changes you made in the source code and I see why they were necessary.

I realized that the DOSFeaturizer and BandFeaturizer also have some of the same issues: they store features in standard dictionaries, class attributes are initialized in the featurize() function, and feature_labels() can’t be called until after featurize() has been called.

I will take point on fixing the DOSFeaturizer for now.

computron:
Ok @albalu can you look at BandFeaturizer?

albalu:
Hi @computron, I addressed these issues for BandFeaturizer in #104 .

Issue #85, opened 2017-12-15
Improving Performance of Composition Features


WardLT:
I recently compared the performance of `matminer` to `Magpie` for computing composition-based features. For ~250k entries, `matminer` requires 24 minutes to evaluate the same features that take `Magpie` 4 seconds. 

From a quick profiling run, it seems like a fair amount of the time (25%) is spent retrieving elemental properties and, [surprisingly](https://github.com/scipy/scipy/issues/3035), computing the mode of a list. Digging further, the slow part of retrieving elemental properties is sorting the properties by atomic number and calling `Composition.get_el_amt_dict()`. My plan is to adjust the AbstractData API such that it is possible to avoid performing either of these operations.

I have opened this issue to see if anyone else has ideas for speeding up the code, and to let you know I am overhauling the AbstractData API. Let me know if I should make a public branch if you want to work on this together and we can avoid merge conflicts.

computron:
Thanks for looking into this! I don't plan to work on this at the moment so don't need the public branch

computron:
(this is completed)

PR #99, opened 2017-12-29
Overhaul of AbstractData API

WardLT:
This merge request contains a fairly extensive set of changes to the AbstractData API. The original motivation for this change was to speed up the calculation of composition-based features (see #85) and, along the way, I needed to change many of the classes that depended on the AbstractData classes.

The speed benefits from this change are not huge. Before the change, matminer took 24 minutes to compute features from [Ward et al](https://www.nature.com/articles/npjcompumats201628) for all 258k entries in the OQMD. This new version takes only 17 minutes, even with changes to the ElementProperty featurizer that increase the computation complexity (and make it match the Ward paper better). I can also confirm that the code is no longer limited by retrieving elemental property data.

## Changes to the AbstractData API

Things that AbstractData now does:
1. Retrieve element property given Element/Specie object
1. Retrieve oxidation states of a certain Element

Things AbstractData used to do, that it no longer does:
1. Retrieve elemental properties for a Composition object. About 25% of the feature calculation time before this merge was spent retrieving Element keys from a Composition object. I eliminated the ability to get properties for an entire composition to discourage needlessly retrieving keys.
1. Estimate oxidation states. Previous versions of the DemlData class would estimate oxidation states following the procedure used in [a paper by Deml et al](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.93.085142). While that code did allow for exactly reproducing this paper, the differences between this code and pymatgen's oxidation state guesser are minor enough that having this code seems more confusing to users than beneficial, in retrospect.
1. Create singleton objects. These objects are small enough in memory and fast enough to initialize that not being able to serialize them is more of a problem than the advantages of being a singleton class.

## Major Changes to Featurizers

*ElementProperty*:
1. No longer computes oxidation-state-dependent properties. This change only effects the "deml" preset. I have introduced new classes to cover the features that are no longer supported
1. Compute concentration-dependent statistics. Previous version of this class did not include the concentration of each element as weights when computing statistics.

*ElectronAffinity*/*ElectronegativityDifference*:
1. No longer determine oxidation states automatically. These classes now use oxidation states that are stored as part of the Composition object, to prevent guessing oxidation states many times. 

*IonProperty*:
1. Introduced a "fast" mode. The "fast" mode only checks if the composition can form a neutral, ionic compound if each element assumes only a single oxidation state. (This is what Magpie does)
1. Modified default behavior to use `oxi_state_guesses`. The `oxi_state_guesses` code from pymatgen allows elements to take >= 1 oxidation state.


computron:
Thanks! Looks like a unit test still fails though. Also the Codacy check points out an unnecessary ``pass`` statement, I think this might be due to indentation (i.e., perhaps the ``pass`` needs to be indented once more)

WardLT:
Sorry for not catching these. I've now fixed the unit tests, but Codacy doesn't like that I added a kwarg when overloading an operation. Should I be concerned about that?

computron:
Hi @WardLT 

Some questions/comments
* does checking for the existence of oxidation states in the various featurizers slow down the code? If so, we could add an option to skip the tests if someone knows that all the compositions already have oxidation states
* For the Codacy check, I do think it's good to not deviate from an interface once you define one. How about allowing the ``get_oxidation_states()`` to take in ``**kwargs``? Then the pymatgen one can have a kwarg for the "oxidation_table" which can have a value of "common" or "all" (or something else if there are other tables in the future, which I think there might be).

Since this is a bit of a longer PR request, it might take me a few rounds of checking through the code to get through everything

WardLT:
Hello @computron ,

To answer your questions:

- The time spent checking for the oxidation states is noticeable, but not large. The featurizer where the effect of checking should be the largest is OxidationStates, which only measures statistics of the oxidation states and does not do any other calculations. The total time taken for checking there is 7% of the total time to featurize dataframe. What do you think of adding a general "debug" mode to the BaseFeaturizer class as a uniform way of turning off/on checking? This could encourage other featurizer implements to make user-friendly error messages, and not worry about the computational cost.
- I think you make a good point about not deviating from the interface. Rather than kwargs, I opted for adding the ability to change which oxidation states are retrieved as a class attribute. Otherwise, keeping track of which AbstractData classes support what kwargs could get confusing.

I look forward to your comments on this PR. Sorry for its large size.

computron:
Hi @WardLT - I think I am just going to merge this and review it more closely later since I'm having a harder time going through this all at once.

PR #101, opened 2017-12-31
Improve featurize dataframe

### LINKED ISSUE #100: featurize_dataframe mishandles features that are equally sized numpy arrays
When BaseFeaturizer.featurize_dataframe() is called for a featurizer which returns equal-dimension
numpy arrays (such as OrbitalFieldMatrix and the in-dev ManyBodyTensor), the code exits in error,
as shown in the Python code and output attached in testfeat.txt. The reason for this is that featurize_dataframe joins the features into a numpy array, which is interpreted as a multidimensional array if and only if each element of the array has the same length. If the features are single values, the array is 1D, and if they are variable-length arrays, the array is interpreted as a 1D collection of other objects. The `assign` call in `featurize_dataframe` does not handle multidimensional arrays.

A possible fix is shown in testfeat.py, in which a dataframe is featurized by adding a Series initialized
from a list of features. As shown in testnew.txt, there does not seem to be a significant speed difference between these methods. I apologize if the fix is underthought; I am not familiar with the reasoning for the original design, so I may well be missing something.

[testfeat.txt](https://github.com/hackingmaterials/matminer/files/1594941/testfeat.txt)
[testnew.txt](https://github.com/hackingmaterials/matminer/files/1594942/testnew.txt)

WardLT:
Minor changes to `featurize_dataframe`:

1. Allow features that are matrices (fixes #100)
1. Prevent copying the dataframe by default. The `assign` operation creates a copy of the dataframe, which is unnecessary and could be expensive.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -7,7 +7,7 @@ from six import string_types
 class BaseFeaturizer(object):
     """Abstract class to calculate attributes for compounds"""
 
-    def featurize_dataframe(self, df, col_id, ignore_errors=False):
+    def featurize_dataframe(self, df, col_id, ignore_errors=False, inplace=True):
         """
         Compute features for all entries contained in input dataframe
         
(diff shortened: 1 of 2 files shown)


computron:
Great, thanks! I think ``assign`` is fine, no need to test

PR #102, opened 2018-01-02
Made OPSiteFingerprints comply with new pymatgen version (bcc factor).

nisse3000:
@WardLT, these changes should resolve the OPSiteFingerprint-related build issues.


matminer/featurizers/tests/test_site.py
```diff
@@ -96,15 +96,15 @@ def test_op_site_fingerprint(self):
             self.assertEqual(l[i], t[i])
         ops = opsf.featurize(self.sc, 0)
         self.assertEqual(len(ops), 35)
-        self.assertAlmostEqual(int(1000 * ops[opsf.feature_labels().index(
-            'oct CN_6')]), 999)
+        self.assertAlmostEqual(round(ops[opsf.feature_labels().index(
```
computron: Why do you need to round if you have assertAlmostEqual() ?

If you are looking for a value very close to 1 (like 0.99), you should just use assertAlmostEqual() and not round (use the places kwarg if you need to set the number of decimal places)

If you are looking for anything that rounds to 1 (i.e., >0.5), then you should assert that the round exactly equals 1. (self.assertEqual() not self.assertAlmostEqual())

nisse3000: @computron, thanks for catching that. Will change it shortly.

WardLT:
Thanks, @nisse3000 !

computron:
@nisse3000 Thanks! I merged. Note that CNSiteFingerprint also had the bcc renormalization, I changed that in a subsequent commit

PR #106, opened 2018-01-08
DOSFeaturizer: fixed semantic error in featurize() and feature_labels()

dyllamt:
changes to featurize()
- _class attribute_ self.feat (standard dict) --> _function variable_ feat (ordered dict)

changes to feature_labels()
- is no longer a call to self.feat.keys()
- can be called before featurize()


computron:
Thanks!

PR #107, opened 2018-01-09
Improve datasets module

kylebystrom:
When a csv dataset is loaded using the dataframe_loader.py functions, the returned dataframe is decorated with pymatgen structure objects, and numerical lists are converted to numpy arrays (in the csv file they are stored as strings). Unit tests are changed to match.


kylebystrom:
Sorry, I had an import statement for a package that isn't available currently, so the circleci tests failed. It should be fine now.

computron:
Thanks!

PR #109, opened 2018-01-13
Fix dataframe indices for featurize_dataframe

WardLT:
Fixes a problem with `featurize_dataframe`

The current implementation leads to features not being matched to the proper entry because the indices of the dataframe are not passed to the new series containing each feature. The mismatch of indices leads to NaN values for many entries because their index is not present in the new series. The current code only works when the first entry is index 0, the second is index 1, and so on.

I added a test case that fails without the fix, and a simple fix to this problem.


computron:
Thanks!

PR #110, opened 2018-01-19
Added a class for running multiple featurizers 

WardLT:
The main use case for this class is when featurizers are used in a different application. This way, that application need not write functions that loop over all featurizers in a list or set or have logic that is different for when a user passes a single featurizer or a list of featurizers.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import pandas as pd
+import numpy as np
 from six import string_types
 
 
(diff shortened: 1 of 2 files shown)


WardLT:
It seems like the 'unused variable' errors in Codacy are false, and a bug in pylint (see [here](https://github.com/PyCQA/pylint/issues/1609)). Do you agree?

computron:
* don't worry about Codacy
* can you modify so that citations() and implementors() so that you don't get repeated entries (e.g., the same implementors or the same papers multiple times?) e.g. you can make it as a set and then convert to list at the end

WardLT:
`citations` and `implementors` do eliminate repeated entries.

computron:
Strange, I missed it the first time - nevermind!

PR #111, opened 2018-01-20
Cleaned up datasets module

kylebystrom:


kylebystrom:
Wait, not quite ready yet sorry, I need to set material_id as the index column.

computron:
Thanks!

computron:
Note that the material_id index didn't seem to work for some reason (e.g., try printing df.index.values)

That said, I think it's fine with the array index and leaving the materials id simply as an extra column. Let's just keep it like that. I'll remove the code about setting the material_id as an index since it's not working anyway.

PR #115, opened 2018-01-23
Started changing implementors and citations functions so that they always return a tuple.

nisse3000:
@computron and @WardLT, should I continue like that?

### DIFF
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -128,13 +128,13 @@ class AGNIFingerprints(BaseFeaturizer):
         return labels
 
     def citations(self):
-        return "@article{Botu2015, author = {Botu, Venkatesh and Ramprasad, Rampi},doi = {10.1002/qua.24836}," \
+        return ("@article{Botu2015, author = {Botu, Venkatesh and Ramprasad, Rampi},doi = {10.1002/qua.24836}," \
                "journal = {International Journal of Quantum Chemistry},number = {16},pages = {1074--1083}," \
                "title = {{Adaptive machine learning framework to accelerate ab initio molecular dynamics}}," \
-               "volume = {115},year = {2015}}"
+               "volume = {115},year = {2015}}",)
 
     def implementors(self):
-        return ['Logan Ward']
+        return ('Logan Ward',)
 
 
 class OPSiteFingerprint(BaseFeaturizer):
@@ -361,10 +361,10 @@ class OPSiteFingerprint(BaseFeaturizer):
     def citations(self):
         return ('@article{zimmermann_jain_2017, title={Applications of order'
                 ' parameter feature vectors}, journal={in progress}, author={'
-                'Zimmermann, N. E. R. and Jain, A.}, year={2017}}')
+                'Zimmermann, N. E. R. and Jain, A.}, year={2017}}',)
 
     def implementors(self):
-        return ['Nils E. R. Zimmermann']
+        return ('Nils E. R. Zimmermann',)
 
 
 # TODO: unit tests!!
@@ -552,10 +552,10 @@ class CrystalSiteFingerprint(BaseFeaturizer):
         return labels
 
     def citations(self):
-        return ['']
+        return tuple([])
 
     def implementors(self):
-        return ['Anubhav Jain', 'Nils E.R. Zimmermann']
+        return ('Anubhav Jain', 'Nils E.R. Zimmermann')
 
     @staticmethod
     def _semicircle_integral(x, r=1):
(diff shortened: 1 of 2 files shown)


computron:
What's the reason for tuple?

My main gripe with tuples is that one-element tuples can be a "gotcha" that leads to errors.

e.g. if you write
citations = ("my_paper")
it doesn't work since it becomes a string. you need to do citations = ("my_paper", ) with the comma which people often forget

nisse3000:
I've seen that @WardLT uses 1-element tuples and figured I follow his lead.  The nice thing about tuples in the context of implementors and citations is that they are, in fact, immutable.  Thought that would be neatly underscored by using tuples over lists.  But my personal main goal is to increase consistency after having found consensus, which this is all about.

computron:
I don’t understand the advantage of immutability here. Eg one might want to
build up a list of citations based on the parameters that were set (eg a
base citation to start and appending one or more citations if certain flags
were set in the featurizer). Its easier to write this code with mutable
objects since you can use if/else statements to build up a citation list.

Similarly one might want to start with the citation list for the first
featurizer and keep extending the list with the citations of subsequent
featurizer which is a bit easier if you return as list.

So I guess I dont understand why immutability is desired here?

(I agree it should be consistent)

WardLT:
@nisse3000 Thanks for noticing this inconsistancy.

Good question about lists or tuples are better.

I actually didn't have much of a rationale for choosing to use tuples. Immutability does make sense for lists of people who wrote an operation, but we don't have much need for that in the software. 

After some more thought, I'm sold by @computron's argument that the mutability of lists allowing for easier modification (e.g., if the implementors changes depending on parameters). 

In that case, my vote is to actually use set. We would loose the order, but I do not view that as too important. Sound reasonable?

nisse3000:
@computron, the first issue is not that bad because the only thing we do (e.g., in CoordinationNumber) is to "return tuple(citation_list)".  Similarly, adding 2 tuples to a new 3 tuple as the composite tuple with "+" works just fine, right?

But, again, I've only started that because I saw that @WardLT did that here and there and thought that there might be a fundamental rationale that I don't see.  Apparently, that's not the case and I just wanna know how we wanna proceed.  I guess the vote went to using lists.

Since we all like consistency, if you don't merge but close this PR, then I know what I gotta do.  :-)

computron:
>> the first issue is not that bad because the only thing we do (e.g., in CoordinationNumber) is to "return tuple(citation_list)

yes, but why make it awkward and force the developer to make that conversion? Unless there is some real benefit to having it be a tuple at the end of the day.

>> Similarly, adding 2 tuples to a new composite tuple with "+" works just fine, right?

Yes, you could also keep redefining your original tuple. e.g.,

```
x = ("citation1",)
if something:
    x += ("citation2",)
return x
```

But this is not as natural as extending lists which are meant to be mutable. Also, I know this seems like a small thing, but the need for a comma for single-element tuples is actually pretty annoying. I know I've fixed the missing comma for other people's functions in matminer before (can't locate the commit though) because they forgot it.

>> I guess the vote went to using lists.

Well not really. Logan voted to use sets. I am not sure. When collecting together lots of citations and implementors from a bunch of different featurizers, I don't think the order really matters anymore so sets are fine. But for a single featurizer, I think it's nice to retain an ordering to the implementors and maybe even the citations. e.g., for the MiedemaFeaturizer, Alireza helped implement a little bit but Qi is really the "boss". So I think it's nice if Qi's name can always come first so people know she is the main contact.

WardLT:
I agree with @computron's point about the advantage of keeping order.

In that case, I change my vote to "list" because it is easier to modify.


On Jan 23, 2018 5:54 PM, "Anubhav Jain" <notifications@github.com> wrote:

the first issue is not that bad because the only thing we do (e.g., in
CoordinationNumber) is to "return tuple(citation_list)

yes, but why make it awkward and force the developer to make that
conversion? Unless there is some real benefit to having it be a tuple at
the end of the day.

Similarly, adding 2 tuples to a new composite tuple with "+" works just
fine, right?

Yes, you could also keep redefining your original tuple. e.g.,

x = ("citation1",)
if something:
    x += ("citation2",)
return x

But this is not as natural as extending lists which are meant to be
mutable. Also, I know this seems like a small thing, but the need for a
comma for single-element tuples is actually pretty annoying. I know I've
fixed the missing comma for other people's functions in matminer before
(can't locate the commit though) because they forgot it.

I guess the vote went to using lists.

Well not really. Logan voted to use sets. I am not sure. When collecting
together lots of citations and implementors from a bunch of different
featurizers, I don't think the order really matters anymore so sets are
fine. But for a single featurizer, I think it's nice to retain an ordering
to the implementors and maybe even the citations. e.g., for the
MiedemaFeaturizer, Alireza helped implement a little bit but Qi is really
the "boss". So I think it's nice if Qi's name can always come first so
people know she is the main contact.

—
<https://github.com/hackingmaterials/matminer/pull/115#issuecomment-359972686>,
or mute the thread

computron:
Ok so we have a decision - everything is going to be lists!

I will close the current PR, but feel free to start a new one with everything consistent as lists

PR #116, opened 2018-01-24
Better like this?

nisse3000:


WardLT:
Looks good to me!

computron:
Thanks!

nisse3000:
@computron and @WardLT, should we also switch over to using lists instead of tuples as return type of feature labels (cf., EwaldEnergy featurizer in structure.py)? I can see people wanting to slightly modify labels in an easy way (not converting tuples to lists). Does that make sense?

computron:
For my standpoint everything should be lists

-features
-feature labels
-citations
-implementors

PR #118, opened 2018-01-24
Consitently returning lists for implementors, citations, and feature_…

nisse3000:
…labels function in featureizers.


nisse3000:
I know.  I never liked to use tuples for that same reason in the first place.

Issue #114, opened 2018-01-23
support multiprocessing in featurize_dataframe


computron:
have a multiprocessing option in featurize dataframe to speed it up. Should detect the number of processors (straightforward with Python) and use multiprocessing package to parallelize.

ardunn:
This is done

Issue #113, opened 2018-01-23
support for grouped DataFrames in featurizers


computron:
There is a way to group columns in DataFrames which makes things look much prettier and more organized. See for example this example from Patrick Huck in MP:

![2018-01-23 10 42 57 am](https://user-images.githubusercontent.com/986759/35293959-323e54d6-002a-11e8-8fa6-843d338c51a9.png)

It would be nice if:

* sample dataframes included grouped columns (e.g., an "outputs" group, a "metadata" group, an "inputs" group) ( @kylebystrom )
* the featurize_dataframe() method could featurize into a group so that all features for a particular featurizer (or multifeaturizer) were grouped together.

computron:
(maybe @WardLT @JFChen3 for the latter improvements to featurize_dataframe() ?)

WardLT:
Nice! I didn't realize you could group columns this way. The code looks decently straightforward. I could give it a shot for marking which featurizer a feature comes from for MultiFeaturizer.

computron:
@ardunn do you consider this complete? if so can you close?

ardunn:
@computron This is definitely not complete. I removed it from a previous PR because it made the PR too complicated. It is on the "todos" though

Issue #93, opened 2017-12-15
Documentation / standardization for BaseFeaturizer


computron:
as suggested by @WardLT 

- Specify a common format for the documentation for the featurizers (e.g., how to list which features it generates)
- Should functions return `nan` or raise an exception if featurization fails?

computron:
So to follow up on this:

For the documentation, I don't have much guidance at this time. If someone has some proposals of what they'd like to see, I can provide some feedback.

As for ``nan`` vs exception, I think the behavior I'd most like is for an individual featurize() function (for one object) to throw an exception if something goes wrong. For example, if the featurizer requires cations and anions (needs + and - charged sites) but you feed in a metal where all the charges are zero - to throw a meaningful exception like RuntimeError("no cation site found"). Hopefully not just a random failure of some line in the code unless the user tries something really crazy (e.g., if they throw in a crystal structure to a composition object, a weird failure is fine).

Note that I am saying this because the ``featurize_dataframe()`` method includes an ``ignore_errors`` key, so I am assuming that if 999 of my entries are good and 1 entry is incompatible with a featurizer, I can still call featurize_dataframe and get data for the 999 and ``nan`` for the 1 without everything crashing. Subsequently, I can use pandas to fill ``nan`` values with 0 or some other value I like.

ardunn:
For `nan` vs exception, I found the most efficient workflow for featurizing a lot (all ~60k MP mats) of (potentially) error prone data was to ignore_errors = True and filter mostly nan features with pandas nan threshold afterwards. But I can also see why throwing exceptions is good, in the case of accidentally putting in a structure for a comp or something, because if you had ignore_errors on, you would be essentially eliminating a good feature and not even know it.

Maybe in the future we can throw a warning to the user for each feature that has 90+% (or some parameter percentage) nan values. That way, even if you have a df with 300+ columns, you might not see which columns are mostly nan (and possibly due to a user error) unless you explicitly check, but you would get a warning saying "Look at feature X to make sure all these nan values are actually due to some crazy error, not due to accidentally putting in an invalid argument.

computron:
@ardunn @WardLT Should this issue be considered closed?

Note that for 90% ``nan`` errors, I think we should just have a separate utility function for that OR just put that kind of filtering in our example notebooks using standard pandas tools (rather than build it into the featurize_dataframe).

ardunn:
[...]

PR #119, opened 2018-01-25
Parallelizing featurization

ardunn:
- Added multiprocessing ability as in #114 
- Added MultiIndex pandas columns as in #113 
- Revamped + minor refactor of featurize_dataframe
- May affect #93 

Notes:

- Multiprocessing needs the exterior wrapper function to function correctly. If a method inside BaseFeaturizer is used as the function passed to Pool.map, python will sometimes throw a cryptic Pickle error. Open to other suggestions if they work
- MultiIndexing uses the Class names as the "outermost" column names and the feature names as the innermost column names. For example ElementProperty on the outside, and 'minimum Number', 'maximum Number', and 'mean Number' on the inside. So right now it is limited to 2 'layers', but in the future we could extend it to a more general case with more layers (e.g., 3 layers, ElementProperty --> Number ---> minimum, maximum, mean). It just depends on how complex/thorough we want to be about the grouped columns. 

Edit:

**This PR is now only adding multiprocessing. There is no MultiIndexing**


WardLT:
A quick idea for the the parallelization: should we split it off into a separate operation? This would allow users to employ parallel featurization even if they are not using `featurize_dataframe`.

If you agree this a good idea, I implemented a version of parallel featurization a few months ago following this model: [link](https://github.com/WardLT/matminer/blob/parallel/matminer/featurizers/base.py)

It still has some problems (does not work with Python 2, requires input to be a list), but could be useful. It also does not create the `Pool` unless users request >1 thread, which has performance implications and Jupyter notebooks on Windows do not always work with Pool.

computron:
Some preliminary thoughts:

- having a featurize_many() function that includes a parallelization option sounds good to me. That way, you can process multiple entries without going through a dataframe
- I think we can drop Python 2 compatibility; pymatgen has already stopped requiring new features to be Py2 compatible. We can just be a little more aggressive about Py3 in the message here: https://hackingmaterials.github.io/matminer/installation.html
- Agree that we should only create the Pool if there is >1 worker
- Only thing I am unsure of is whether the default n_jobs should be 1. My initial thought is to have it parallelize by default. If one is running a Jupyter notebook or really doesn't want to parallelize, they should be able to run a line of code like "MMConfig().AUTO_PARALLELIZE = False" which would set the default to False. e.g. you could run that line of code at the top of a Jupyter notebook and then all subsequent calls to featurize_many() would not parallelize without having to set n_jobs each time. Note that I am flexible on the actual form of how it's achieved, but some kind of global variable that can be set once to turn off parallelization and that can be put in the top of Jupyter notebooks. We could also do the inverse - i.e., the default being parallelize=False, but being able to run one line of code near the top that just always turns it on without having to set n_jobs each time.

Another note is that Alex's commit includes many new features, not just parallelization, e.g. the MultiIndex. So we should figure out how to merge all this stuff together.

ardunn:
@WardLT @computron having the featurize_many seems like a good idea to me as well. I am revising this PR:

- Have featurize_many
- Pool only if >1 worker
- Set default n_jobs=None (auto-detect num of cores)

computron:
Note - it's also cleaner to not have the wrapper as in @WardLT 's solution. I was thinking to mainly use that implementation for parallelization unless there are issues

ardunn:
Fine by me, although we may want to change how it is parallelized. As I am reading the code, it looks like a parallel process is spawned for each entry?

```
with Pool(n_jobs) as p:
    return p.starmap(self.featurize, entries)
```


Could it be more efficient to separate the list into n_jobs chunks and process them with n_jobs processes?

WardLT:
I think this code should only generate `n_jobs` threads, and does chunk the `entries` array. What am I missing?

ardunn:
Oh, ok, maybe I misunderstood it. Full steam ahead then!

ardunn:
Closed so we can use @WardLT implementation. I'll make a separate PR for multiindex

computron:
@WardLT can you submit a PR so I can merge and then @ardunn can build on top?

WardLT:
Sure, I'll do that when I'm back on WiFi in an hour or so.

ardunn:
@WardLT I used your code and added the ability to ignore_errors. Besides that it is similar to your parallel branch (multiindex removed).

If your current parallel branch is newer and already addresses this though, I'll close this PR for good

WardLT:
Sounds great! I hadn't integrated ignore error yet, so your branch sounds
better.


On Jan 26, 2018 6:59 PM, "Alex Dunn" <notifications@github.com> wrote:

Reopened #119 <https://github.com/hackingmaterials/matminer/pull/119>.

—

<https://github.com/hackingmaterials/matminer/pull/119#event-1444622315>,
or mute the thread

ardunn:
@WardLT Ok great! Also beware incoming Py2.x test failures! @computron

ardunn:
Note that `n_jobs=cpu_count()` by default will fail Py 2.x tests. Right now `n_jobs =1` by default, so it passes 2.x. Perhaps can help @computron decide on how to proceed

ardunn:
@WardLT I've just noticed that both our implementations (or anything featurizing with `Pool.map` or `Pool.starmap`) have the possibility to corrupt a Structure object when they are passed to the pool. This can be avoided by passing the structure objects as their dict representations (`Structure.as_dict()`) but it is definitely something to look out for or fix soon. I think it may be due to the `chunksize` parameter of `.map`/`.starmap`...

computron:
@ardunn Why not retain starmap? My understanding was so that the code would also work for featurizers that require multiple arguments like the structure featurizers:

http://n-s-f.github.io/2016/12/23/starmap-pattern.html

Can you confirm that the current PR also works when each featurize() call takes in multiple args and that starmap isn't needed?

computron:
@ardunn You should also include a unit test for new functions like featurize_many(). The unit test should include one example of a straight list entry [obj1, obj2, obj3] as well as one with multiple inputs per function [(obj1a, obj1b), (obj2a, obj2b), ...]

ardunn:
@computron I can confirm it works for multiple argument featurizers. I tried, for example, the MinimumRelativeDistances featurizer with both structure and cutoff arguments and it works for `njobs>=1`.  The current implementation is doing the exact same thing as starmap, but just has an if-then for `ignore_errors` in a wrapper method. If we want to do away with `ignore_errors` however, starmap should be used in the original way @WardLT  wrote (for cleanness).


I'll make those tests, but a heads up, a `featurize_many` test with `n_jobs>1` **will definitely fail on Py2.x**, so anytime a 2.x person runs tests, they'll be met with at least one error.

computron:
A few notes:

* should have a fix for Structure corruption before I merge
* don't modify the unit tests for Py2. Instead, you can add some code in ``featurize_many()`` itself that, if the user set n_jobs > 1 and also has Python < 3, will (i) use warnings.warn() to tell the user that multiprocessing has been disabled due to Python 2, and (ii) override n_jobs =1.
* I updated the CircleCI tests to only test Python 3.6 (and not 2.7). So now the PR won't complain if you break Py2.

WardLT:
@ardunn What kind of errors do you get with the Structure object? Does this have to do with pickling the Structure object?

A few other comments:
- Should we use `functools.partial` to pass `ignore_errors` so that we do not have to make `ignores_errors` a class attribute? 
- `featurize_wrapper` seems like it should be a private function. Do you agree?
- If we want Py2 support, I think we could replace `with Pool() as p` with:

 ```python
p = Pool(n_jobs)
try:
  p.map(...)
finally:
  p.close()
```

If I recall correctly, this should work in both Py2 and Py3.

**Comment by ardunn** [comment]
@WardLT I don't get an Error, but the structures are silently corrupted. Let me know if you can reproduce this with the multiprocessing code currently in this PR:

```python
class ShowStructureCorruption(BaseFeaturizer):

    def featurize(self, s):
        return [s]

    def feature_labels(self):
        return ["Corrupt Structure"]

if __name__ == "__main__":
    # Using structures from StructureFeaturesTest...
    df = pd.DataFrame.from_dict({'Correct structures': [ni3al, nacl, cscl]})
    sc = ShowStructureCorruption()
    df = sc.featurize_dataframe(df, 'Correct structures', n_jobs=2)
    print(df)
```

This code gives:

```
                                                               Correct structures                                                              Corrupt structures
0  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]
1                                      [[0, 0, 0] Na+, [2.324, 1.643, 4.025] Cl-]                                      [[0, 0, 0] Cs+, [2.324, 1.643, 4.025] Cs+]
2                                      [[2.105, 2.105, 2.105] Cl-, [0, 0, 0] Cs+]                                      [[2.105, 2.105, 2.105] Cs+, [0, 0, 0] Cs+]
```

Note the species are wrong in NaCl and CsCl. Also note that if you change the order of `nacl` and `cscl` in `structs`, the the species are corrupted in a different way:

```python
if __name__ == "__main__":
    df = pd.DataFrame.from_dict({'Correct structures': [ni3al, nacl, cscl]})
    sc = ShowStructureCorruptiuon()
    df = sc.featurize_dataframe(df, 'Correct structures', n_jobs=2)
    print(df)
```
```
                                                               Correct structures                                                              Corrupt structures
0  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]
1                                      [[0, 0, 0] Na+, [2.324, 1.643, 4.025] Cl-]                                      [[0, 0, 0] Cl-, [2.324, 1.643, 4.025] Cl-]
2                                      [[2.105, 2.105, 2.105] Cl-, [0, 0, 0] Cs+]                                      [[2.105, 2.105, 2.105] Cl-, [0, 0, 0] Cl-]
```
With `n_jobs=1`, the structures are not corrupted:
```
                                                               Correct structures                                                              Corrupt structures
0  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0.   1.76] Ni, [0.   1.76 1.76] Ni]
1                                      [[0, 0, 0] Na+, [2.324, 1.643, 4.025] Cl-]                                      [[0, 0, 0] Na+, [2.324, 1.643, 4.025] Cl-]
2                                      [[2.105, 2.105, 2.105] Cl-, [0, 0, 0] Cs+]                                      [[2.105, 2.105, 2.105] Cl-, [0, 0, 0] Cs+]
```

I've checked the format of each structure as it's being processed, and the structure goes from correct to corrupted during `Pool.map()`, so it is probably something to do with pickling. I've also since realized that converting the structures to dictionaries _does not_ fix the problem.

ardunn:
Regarding the other points:

- `functools.partial` : Sure, would the reason for this be to prevent subclasses from overriding .ignore_errors? 
- I agree.
- Need @computron input

computron:
I am on the fence for rearranging for Py2 support.

Part of me says that this is a small concession to make for continuing Py2 support.

But part of me just wants to blow up Py2 support - make it clear that we are not supporting Py2 so that we don't have to keep deciding whether or not to support Py2 in the future over and over. This is a minor concession but then if we start making 10 minor concessions it starts to become a major concession.

So up to you both.

WardLT:
`functools.partial`: Yes, I am concerned about subclasses overriding `ignore_classes`. Not very worried, but it seemed like an easy change.

`Structure` and `map`: @ardunn That's a surprising error! I haven't yet run into pickling corruption issues. Any luck on figuring it out? 

Py2 support for multiprocessing: I'm fine with blowing up Py2 support.

ardunn:
@WardLT 

- `ignore_classes` :  Couldn't we just make it `__ignore_classes` and Python automatically prepends the base class name? I know it works for methods but not sure for attrs.

- `Structure` corruption: No progress. Not sure what is the problem. However I think this is a major issue so I'll be looking into it some more. IMHO its even worse than multiprocessing throwing an exception because a user might just be featurizing parts of dataframe in parallel (probably will be a common theme with structure featurizers, since for example NearNeighbors algorithm is computationally intense), and get back a nice dataframe without even knowing anything is wrong. Like I showed in the code above, it only corrupts some structures, which is very weird.

computron:
Do you know if it still corrupts it if it is an IStructure (immutable Structure) instead of Structure?

ardunn:
@computron Both structure and IStructure seem to be corrupted. However I've noticed it only corrupts structures defined in the same way as the `test_structure.py` file, _e.g._:
```python
ni3al = Structure(
    Lattice([[3.52, 0, 0], [0, 3.52, 0], [0, 0, 3.52]]),
    ["Al", ] + ["Ni"] * 3,
    [[0, 0, 0], [0.5, 0.5, 0], [0.5, 0, 0.5], [0, 0.5, 0.5]],
    validate_proximity=False, to_unit_cell=False,
    coords_are_cartesian=False, site_properties=None)
```
I just tried with ~1000 structures from MP (using `MPDataRetrieval`) and none of them were corrupted. Is there something different between the way these test structures are created and the way MPRester creates structures? Both are `<class 'pymatgen.core.structure.Structure'>` objects.

computron:
I don't know offhand of any problems (you have an unneeded comma after
"Al", but that's just stylistic, shouldn't affect anything)

I think you'd need to look into the Structure code or send a note to
pymatgen. You can also try systematically changing the constructor to try
different ways of creating the Structure to figure out what "breaks" it

Best,
Anubhav

WardLT:
@ardunn Making `ignore_errors` a `__` attribute sounds like a good solution to me. Regardless of whether Python prepends the class name to it, using the `__` makes it clear `ignore_errors` should not be touched by subclasses.

ardunn:
@WardLT ok!

Regarding test structure corruption, maybe this is something @nisse3000 knows?

ardunn:
@computron Done

@WardLT Thanks to @nisse3000, the root of the structure corruption has been identified. It does not have to do with the code in this PR, but rather the pymatgen `Structure` and pickling. When oxidation states are present, the structures are corrupted after serialization/deserialization. 

```python
cscl = Structure(
    Lattice([[4.209, 0, 0], [0, 4.209, 0], [0, 0, 4.209]]),
    ["Cl1-", "Cs1+"], [[2.105, 2.105, 2.105], [0, 0, 0]],
    validate_proximity=False, to_unit_cell=False,
    coords_are_cartesian=True, site_properties=None)


pickle.dump(cscl, open('cscl.pickle', 'wb'))
print(pickle.load(open('cscl.pickle', 'rb')))
```
```
>>>
Full Formula (Cs1)
Reduced Formula: Cs
abc   :   4.209000   4.209000   4.209000
angles:  90.000000  90.000000  90.000000
Sites (2)
  #  SP           a         b         c
---  ----  --------  --------  --------
  0  Cs+   0.500119  0.500119  0.500119
  1  Cs+   0         0         0
```


Whereas the same structure with no oxidation states does not corrupt the structure:
```python
cscl_nooxi = Structure(
    Lattice([[4.209, 0, 0], [0, 4.209, 0], [0, 0, 4.209]]),
    ["Cl", "Cs"], [[2.105, 2.105, 2.105], [0, 0, 0]],
    validate_proximity=False, to_unit_cell=False,
    coords_are_cartesian=True, site_properties=None)

pickle.dump(cscl_nooxi, open('cscl_nooxi.pickle', 'wb'))
print(pickle.load(open('cscl_nooxi.pickle', 'rb')))
```
```
>>>
Full Formula (Cs1 Cl1)
Reduced Formula: CsCl
abc   :   4.209000   4.209000   4.209000
angles:  90.000000  90.000000  90.000000
Sites (2)
  #  SP           a         b         c
---  ----  --------  --------  --------
  0  Cl    0.500119  0.500119  0.500119
  1  Cs    0         0         0
```


The reason structures are corrupted in different ways during multiprocessing based on their list order is because (1) the corrupt pickled structures are chunked so their data can be processed, and (2) the order of the chunked data affects the exact way the structures are corrupted. 


I'll make an issue on pymatgen for this. However, since this issue does not originate in matminer, I think this PR can be merged.

ardunn:
Also a heads up to @computron I think py27 tests are still running on CircleCI, was not sure if that is desired or not

computron:
Thanks!

Issue #97, opened 2017-12-21
Add a BagofBonds structure featurizer


computron:
This would take in a structure and retain a list that contains the number of bonds (probably as fraction). For example, if a structure had 2 Li-O bonds and 3 Li-P bonds, it could return [0.4, 0.6] as the featurizer value and ["Li-O", "Li-P"] as the feature labels.

Some aspects to note:
* You probably want to allow any pymatgen NearNeighbor class to be used as the bond detection method. Some presets might help with usage
* If you are featurizing an entire data frame, you need the feature labels to match and be a concatenation of all the bonds in all the various structures in the data frame. Also some complications if you fit a model and then predict a structure for which there are no bonds in the training set. Some hashing tricks from text mining might help with this. Talk to AJ if you are implementing and this is confusing to you.

ardunn:
this is done (excluding hashing tricks)

PR #120, opened 2018-01-27
Added a 'BagofBonds' Featurizer

ardunn:
- Addresses #97 

- __For a single structure:__
Returns the fraction of each kind of bond combination, based on a pymatgen `NearNeighbors` method

- __For a dataframe of structures:__
Returns the 'unified' list of all bond fractions, concatenated from the bonds present in each structure. Bonds present in the unified list but not in an individual structure (e.g., no C-C bond in CsCl) are represented as NaN:

For example, on a dataframe containing the `diamond_no_oxi`, `ni3al`, and `cscl` structures from `test_structures.py`:
```python
bf = BondFractions.from_preset("MinimumDistanceNN")
bf.featurize_dataframe(df, 'structure')
```
```
>>>
  composition            structure  C-C bond frac.  Al-Ni bond frac.  Al-Al bond frac.  Ni-Ni bond frac.  Cl--Cs+ bond frac.  Cl--Cl- bond frac.  Cs+-Cs+ bond frac.
0         (C)  [[2.554, 1.806, ...             1.0               NaN               NaN               NaN                 NaN                 NaN                 NaN
1    (Al, Ni)  [[0. 0. 0.] Al, ...             NaN               0.5               0.0               0.5                 NaN                 NaN                 NaN
2  (Cl-, Cs+)  [[2.105, 2.105, ...             NaN               NaN               NaN               NaN                 1.0                 0.0                 0.0

```

Alternatively, the `NaN` values can be replaced during featurization with a 'bad bond value' (`bbv`, open to name changes),  an argument/attribute of BondFractions.

```python
bf.bbv = 0.0
bf.featurize_dataframe(df, 'structure', inplace=True)
```
```
>>>
  composition            structure  C-C bond frac.  Al-Ni bond frac.  Al-Al bond frac.  Ni-Ni bond frac.  Cl--Cs+ bond frac.  Cl--Cl- bond frac.  Cs+-Cs+ bond frac.
0         (C)  [[2.554, 1.806, ...             1.0               0.0               0.0               0.0                 0.0                 0.0                 0.0
1    (Al, Ni)  [[0. 0. 0.] Al, ...             0.0               0.5               0.0               0.5                 0.0                 0.0                 0.0
2  (Cl-, Cs+)  [[2.105, 2.105, ...             0.0               0.0               0.0               0.0                 1.0                 0.0                 0.0

```
- Citation is this paper [http://pubs.acs.org/doi/ipdf/10.1021/acs.jpclett.5b00831](url) although the concept here is simpler than what they do. Considering they coined the `BagofBonds` term with the same concept though we should probably cite them.

- Was unsure how to proceed on hashing the bond fractions as mentioned in #97 so that is not included in this PR.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -1,16 +1,15 @@
-from __future__ import division, unicode_literals
+from __future__ import division
 
 import pandas as pd
 import numpy as np
 from six import string_types
-from multiprocessing import Pool
 
 
 class BaseFeaturizer(object):
     """Abstract class to calculate attributes for compounds"""
 
     def featurize_dataframe(self, df, col_id, ignore_errors=False,
-                            inplace=True, n_jobs=1):
+                            inplace=True):
         """
         Compute features for all entries contained in input dataframe
 
@@ -23,10 +22,6 @@ class BaseFeaturizer(object):
                 exceptions are thrown if True. If False, exceptions
                 are thrown as normal.
             inplace (bool): Whether to add new columns to input dataframe (df)
-            n_jobs (int): Number of parallel processes to execute when
-                featurizing the dataframe. If None, automatically determines the
-                number of processing cores on the system and sets n_procs to
-                this number.
         Returns:
             updated Dataframe
         """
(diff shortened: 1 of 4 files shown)


computron:
why add ``crystal_system_int``?

One can use pandas to get the data as ints:

http://benalexkeen.com/mapping-categorical-data-in-pandas/

Plus for some models like linear regression, having them as strings allows you to call the get_dummies function which lets you do a one-hot encoding matrix of vars

So I think if someone really wants to get the categories as ints they can use pandas to convert. That way matminer doesn't need to duplicate the info as string and int in every single featurizer that returns a categorical var.

If you think it would be helpful to put some of the code in that link in matminer for reference, you could put it in the ``utils/conversions.py`` module to help people convert from string to int or string to dummy-matrix

ardunn:
@computron I do not think that is a part of the code I added?  The changes I see regarding `crystal_system_int` is just a Pycharm PEP8 auto-reformatting added it on a new line. The only new code is in BagofBonds, the rest of the changes in structure.py are PEP8 reformatting. 

Or if you mean something else, could you clarify?

computron:
Ah I see, OK I was just looking at the changes very quickly

Note: you don't need to change code to make Codacy happy. It's there for suggestions only.

computron:
Side note: I remembered why I added ``crystal_system_int``. It's because you can assign a real order to the crystal system in terms of increasing or decreasing symmetry. So they are not just dummy indices for each category, there is a specific order of crystal systems

PR #121, opened 2018-01-29
Made changes so that OP changes in pymatgen are addressed.

nisse3000:


computron:
@nisse3000 realized this after the merge, but when updating requirements you should update both requirements.txt and setup.py. I don't really know why there are two places, just that some installs can break if you don't do both. I pushed a commit to do this after pulling your merge, so this is just for the future

nisse3000:
@computron, sorry about that.  Will remember next time.

PR #122, opened 2018-01-31
Update data.py

mtdg-wagner:
Let users know which particular elemental property is not found in MagpieData rather than simply stating the KeyError.

### DIFF
--- a/matminer/utils/data.py
+++ b/matminer/utils/data.py
@@ -210,7 +210,11 @@ class MagpieData(AbstractData, OxidationStatesMixin):
                         Element.from_Z(atomic_no).symbol] = prop_value
 
     def get_elemental_property(self, elem, property_name):
-        return self.all_elemental_props[property_name][elem.symbol]
+        try:
+            return self.all_elemental_props[property_name][elem.symbol]
+        except KeyError:
+            print('Property: {} not found for {}'.format(property_name, elem.symbol))
+            return float("NaN")
 
     def get_oxidation_states(self, elem):
         return self.all_elemental_props["OxidationStates"][elem.symbol]


WardLT:
Hello @mtdg-wagner ! And, thanks for the PR!

I agree with giving users a better warning message about which properties are missing, but I think this code should raise an Exception rather than returning `NaN`. 

When we read in the Magpie data files, we already put `NaN`s when a property is missing for a certain element. So, there should only be `KeyErrors` when the user of MagpieData passes a property that doesn't exist. I think missing (or misspelled) properties should return a error rather than a warning message.

Am I right about this code failing only when properties are missing? And, if so, do you agree an Exception is preferred?

mtdg-wagner:
Hi Logan! Raising the exception is a good idea if we could just let the user know which property is missing.

This code failed when calling featurize_dataframe with a dataframe of compositions that should have worked fine. I'll attach the Jupyter notebook I am working with and the csv file which fails to be featurized. 
[featurizing.tar.gz](https://github.com/hackingmaterials/matminer/files/1682910/featurizing.tar.gz)

mtdg-wagner:
I can't get the error to replicate running just now on my work desktop. It could just be an issue with the libraries on my home Windows computer.

WardLT:
Interesting. How did you install it on your Windows computer? 

I think this error might have to do with the *.table files used by matminer not being copied by pip. Does matminer work on your Windows computer when you install it with `pip install -e .`?

mtdg-wagner:
I will check when I get home. I installed using pip without the -e flag, so maybe that is indeed the issue.

computron:
I am pretty sure pip just installs .py files by default. We probably need to update the ``package_data`` in setup.py as well as MANIFEST.in to get the tables files and other data files included. I'll add a commit to do this now and try to release an updated matminer version soon.

computron:
Ok I just now pushed matminer 0.2.2 - please try reinstalling / upgrading to this version and see if that fixes any missing tables

mtdg-wagner:
Upgrading using normal pip still resulted in the tables being absent. 

I followed Logan's suggestion of actually cloning the repo and running `pip install -e .`, and now it works. Might be a problem with using pip on Windows.

computron:
Since the root of this problem seems to installation, I am going to close this PR. I think the help message won't really be necessary and I think it's better if we just have the code fail on incorrect property setting than returning NaN. Thanks for reporting this, it helped us iron out the installation issues which we were unaware of.

Issue #94, opened 2017-12-15
Adding MDF data retrieval class


computron:
Talk to @WardLT if interested

WardLT:
Awesome! Are you using the MDF forge package to build it? Feel free to bug
me if you have questions.

PR #124, opened 2018-02-05
MDF interface

montoyjh:
Sorry for the delay on this, I got tied up with a bug related to the anonymous invocation of the forge (this functionality is nice, because you can write unittests without an API key or logging in through globus).  Intended to address #94.

* Adds an MDF data retrieval tool, data_retrieval.retrieve_MDF.MDFDataRetrieval, which invokes a Forge object from mdf_forge (added as a dependency) and accepts any kwargs used to construct Forge.
* MDFDataRetrieval has two methods `search` and `search_by_query` which allow a user to search either by a set of keyword arguments (documented in the docstring for `search`) or a direct MDF query.
* I also accidentally wrote my own flatten_dict method, which I thought was a bit more readable (although no more or less functional) than the one already in MatMiner, so I replaced the old one and added a unittest in utils.

Usage:
```
mdf_dr = MDFDataRetrieval(anonymous=True) # or no anonymous, will be prompted for auth
results = mdf_dr.search(sources=['oqmd'], elements=['Be', 'Ag', 'V'])
```

Working through codacy.


montoyjh:
For some reason, I'm no longer able to see the specific issues codacy is highlighting with respect to this PR.

montoyjh:
gonna rebase and resubmit with all of these codacy fixes squashed.  As an aside, is there a generic fix to the issue with the unused __class__ variable that I'm seeing in the codacy reports?  I don't understand what the issue is there.

computron:
I think it's some kind of Codacy bug, don't worry about it. Codacy is just there as a guideline anyway, sometimes if you try to make all its suggestions go away the code gets worse. So humans > Codacy.

Will take a look at this PR and merge hopefully soon, if I ever escape from meetings ...

computron:
Thanks! My only thought is whether forge should be a required or optional dependency. I get nervous about required dependencies sometimes since they can take down a matminer installation.

For now I'll merge as is, if we get any reports of breaking installs due to forge we should move it to optional.

Issue #123, opened 2018-02-01
fix pip install


computron:
Apparently pip install of matminer does not copy data tables - which results in much of the core code breaking. Someone needs to investigate and fix. See:

https://github.com/hackingmaterials/matminer/pull/122#issuecomment-362127852

PR #125, opened 2018-02-06
modified setup.py to work with pip on windows

montoyjh:
I'm not really sure why this works, but it seems to fix the setup.py issues on windows for me (which mirror the pip issue #123 references.  I think it might have something to do with how the tree for the glob expressions is constructed in setuptools, i. e.

matminer->utils->data_files->['*.csv', '*.json']
overrides
matminer->utils->data_files->magpie_element_data->[*.table]

It's baffling why this would be particular to windows, though.


computron:
Thanks!

PR #127, opened 2018-02-08
Avoid overwriting on input df columns + ignore_errors in BagofBonds

albalu:
3 small changes:

1) extended the implementation of ignore_errors inside BagofBonds so it will skip entries with bad structures.
2) rename gap to gap_AO in AtomicOrbitals as "gap" has a higher chance of already being used in input df. 
3) extended the logic in previous point and raise if input df column names and feature_labels overlap.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -26,6 +26,13 @@ class BaseFeaturizer(object):
             updated Dataframe
         """
 
+        # Generate the feature labels
+        labels = self.feature_labels()
+
+        for col in df.columns.values:
+            if col in labels:
+                raise ValueError('"{}" exists in input dataframe'.format(col))
+
         # If only one column and user provided a string, put it inside a list
         if isinstance(col_id, string_types):
             col_id = [col_id]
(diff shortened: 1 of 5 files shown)


albalu:
regarding bad structures in BoB:
my case was for when structure is None so in that case there are no bonds but in the case you described, you are right and it might cause issues that are hard to track. I will change that part.

regarding inplace: as far as I know the inplace option makes the difference between the following two in neither of the cases I expect matminer to overwrite my current columns but just to add features:
df = some_featurizer.featurize_dataframe(df, inplace=True or False)
vs.
some_featurizer.featurize_dataframe(df, inplace=True)

ardunn:
- BoB structures: Ok good my me. If there is a bad structure (ie `s` is not `None` but is bad in some way) it is better for `enumerate_all_bonds` to at least preliminarily identify *all* the possible bonds for *all* the structures and just wind up having .bbv for all values in those columns if `ignore_errors=True`. That way we will never have a scenario where `featurize` is trying to write bonds back to the dataframe when those bonds don't exist as columns (what a nightmare). 👍 

- inplace: Ah ok. That is a very good point. You would not want a previous column to be overwritten in any case, especially when inplace=True by default. 👍

computron:
1. need to resolve conflicts since after the most recent PRs there are merge conflicts

2. I don't really understand the point of ignore_errors in BoB. Why would a structure be None?

Note that what I would really like to see in BoB is:
- taking in a list of "allowed bonds" - e.g., if you training set only contained 1000 distinct bonds, then these should be the only allowed bonds for your test set
- If your structure contains a new bond that is not in allowed_bonds (e.g., test set contains a bond not in training set), you can either (i) return NaN or (ii) find the "closest" bond and just fill the value there. For defining closest bond, one thing you can do is distance in Mendeleev numbers between the bonds. e.g., "Li-O" and "Na-O" are similar because the Mendeleev no of Li is 1 and Na is 2, so the distance between these bonds is only one. If there are multiple allowed_bonds with the same Mendeelev number distance, just distribute that bond over multiple of the allowed_bonds (e.g., half in one, half in the other)

Also best not to mix thigns like renaming gap to gap_AO (simple) with more complex things like BoB

albalu:
ok BoB needs more thinking and discussion that I didn't expect. I'll just close this make a new PR to address column name issues for now.

ardunn:
@albalu I think you should be able to just copy-paste the input df code into the newest parallelized master branch? There is no hard conflict.

PR #130, opened 2018-02-09
strictly checking feature_labels vs. current labels to avoid overwritting

albalu:
This is to address the first 2 parts of #127 . Some new tests had to be updated.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -40,6 +40,11 @@ class BaseFeaturizer(object):
         # Generate the feature labels
         labels = self.feature_labels()
 
+        # Check names to avoid overwriting the current columns
+        for col in df.columns.values:
+            if col in labels:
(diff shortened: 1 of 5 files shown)


ardunn:
Why is BoB test being updated?

albalu:
"restart df to avoid overwriting features" to avoid the error cause by new name checking I added

albalu:
strange: locally test_chemicalSRO in test_site.py fails for me but not here. Anyone else experience the same thing? @Qi-max ?

Qi-max:
eh, the test passes when submitted yesterday, double check ing...

Qi-max:
Understood. This is due to the new modification of featurize_dataframe, which call feature_labels before featurize (previously the sequence is reversed), but my label contains the elements name in the structure that can not be determined beforehand.

Issue #117, opened 2018-01-24
matminer can include capability to functionalize features


computron:
Part 1: This would be a tool that takes in a feature (or list of features) and returns some functional outputs like:

* x
* 1/x
* x^05.
* x^-0.5
* x^2
* x^-2
* x^3
* x^-3
* ln(x)
* 1/ln(x)
* exp(x)
* exp(-x)

where x is the original feature value. Note that the list of features above is from "Machine Learning and Materials Informatics: Recent Applications and Prospects" by Ramprasad et al.

Part 2: This would take in a list of features (either raw features or perhaps after applying the functions above) and combine them, i.e., give you x1*x2 for all features. Note that if you applied the functions above, you will also have features like x1, x2, 1/x1, 1/x2, etc... so by multiplying all feature combinations you will end up with things like x1*x2, x1/x2, x2/x1, etc. Unfortunately, some of these will just be 1 (x1*1/x1) or redundant (x1*x1^2 is the same as x1^3).

PR #131, opened 2018-02-10
Function featurize

montoyjh:
## Featurization of functional outputs of pre-existing dataframe values

This PR adds functionality intended for functionalization of dataframe entries per #117.

* Introduces matminer.featurizers.FunctionFeaturizer class, which is initialized with a list of string expressions (e. g. ['1 / x', 'x ** 2', etc.]) and includes defaults according to #117.
* FunctionFeaturizer adds features by substituting the provided expressions with values from the featurized columns (e. g. col_id in featurize_dataframe).
* FunctionFeaturizer supports multiple combined features, and allows for a custom "combo_function" to combine them (default np.prod, i. e. cumulative product)
* The sympy approach allows for descriptive and optionally latexified column names according to the provided col_ids, and also allows for the removal of redundant features.

## Additional changes

* Support for **kwargs to BaseFeaturizer.featurize_dataframe, which are passed to featurize
* Support for feature_label customization via an optional "label_props" kwarg for BaseFeaturizer.featurize_dataframe.  This is used for the label customization in FunctionalFeaturizer.featurize_dataframe
* Per my last pull request on the MDF interface, updated requirements.txt both to include sympy and to move mdf_forge to an optional dependency.


computron:
Thanks! A couple of comments:

1. Needs some docs (somewhere)
2. There is a lot of added complexity (and changing of standard method signatures) in order to try make things general, but I don't *think* it's needed.
- Do we really need an option for postprocessing as float vs complex, etc? Can we just postprocess everything as float? Or, if you really want it, put this in the constructor rather than the featurize() method? We are really trying to keep options out of featurize() to the extent possible.
- Similarly, there is a lot of added complexity to the feature_labels with very general passing of any type of argument into there. Can we just change this to "don't make me think"? i.e., for formatting, etc., just automatically do what's reasonable. If someone really wants to tune something, put an option in the constructor of the specific featurizer to change that option. But again try to keep the method signatures clean and not add too much complexity there so we don't end up with a BaseFeaturizer() that has a million potential actions and knobs.

Basically, I am saying to sacrifice generality of the code for simplicity, ease of first use, ease of understanding + maintenance, etc.

WardLT:
Would it make more sense to implement FunctionFeaturizer using scikit-learn's `TransformerMixin` subclass (as in [PolynomialFeatures](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.PolynomialFeatures.html#sklearn.preprocessing.PolynomialFeatures)) and not the `BaseFeaturizer` ?

Having FunctionFeaturizer as a BaseFeaturizer seems a little bit unnatural. When using a model, I would have to keep track of multiple Featurizers, the order in which I call them, and that each featurizer takes different data. I can't just stick them all in to a MultiFeaturizer and forget about it.

While this 'negative' about having FunctionFeaturizer as a BaseFeaturizer is somewhat aesthetic (as we do not yet many applications that assume a ML model has exactly one featurizer), I'd also argue that implementing with the `TransformerMixin` base class will allow us to use the FunctionFeaturizer with GridSearchCV and other tools designed to perform hyperparameter optimization with sklearn.


Some other thoughts:
- Should the columns being operated on be stored as a class attribute? This way, you can call `feature_labels` with its current API and without having to modify `featurize_dataframe`.
- I think it would make sense to cite [Ghiringhelli 2015](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.114.105503) in addition to/in place of the npj review paper. AFAIK, the Ghiringhelli PRL was the first use of this feature expansion technique

montoyjh:
@computron 
* Planning on adding docs with the reformatting of the tutorials/examples that I'm taking on (will be doing these Thurs-Sun this week).  If you want I can add a preliminary example in this PR to be expanded with the other featurizers.
* postprocessing we can definitely get rid of.  I'm not sure how you feel about the descriptive labels, I think it's nice to actually know what the feature is, and this requires processing the input features explicitly (i. e. adding an kwarg to feature_labels).  This might be achieveable in another way though.
@WardLT 
* Personally, I think it'd be simpler to keep the same base class structure where possible, but it might be possible to use some of TransformerMixin's capabilities on top of the BaseFeaturizer.
* I dunno about the class attribute because it removes a key abstraction from the featurizer.  In my mind, a featurizer's construction is supposed to be independent of the target dataframe's content.
* Will update citation as suggested.

computron:
By descriptive labels, do you mean ``col_id`` (aside: should probably be ``col_ids`` since this is an array)?

If so, then I agree it's important to list the col_ids in the feature labels. The issue is that there isn't really a good framework for this.
* Currently, most featurizers with this kind of complication assume that you will call featurize() before feature_labels(). The featurize() might store some information (like col_ids). Then the feature_labels can reference this information. This has issues, see #135 
* Your solution is for feature_labels to also take in the same type of arguments as featurize. This is possible, but for many featurizers (not yours) it would require essentially re-processing the data in order to figure out the feature labels. This is because many feature labels depend on both the rows and columns of the dataframe - different rows give different features. I'd like to avoid this as a long-term solution since it "fixes" your case but not others.

I think we should maybe move that discussion to #135

WardLT:
Hello @montoyjh,

- Personally, I think it'd be simpler to keep the same base class structure where possible, but it might be possible to use some of TransformerMixin's capabilities on top of the BaseFeaturizer.

I agree in keeping the same class structure, But that is what leads me to think TransformerMixin is a better choice for FunctionFeaturizer. I'd use this functionality in the same way I'd use PolynomialFeatures or other transformers from scikit-learn: modifying the list of features generated by `matminer` to make a machine learning model perform better.

- I dunno about the class attribute because it removes a key abstraction from the featurizer. In my mind, a featurizer's construction is supposed to be independent of the target dataframe's content, but I'm new to this sort of thing, so maybe I'm wrong here.

Solid point! I agree that we should keep the featurizer's construction separate from the dataframe content. But, I'd balance that concern with the goal of the `feature_labels()` operation to tell a user what kind of features this class should compute without having to provide details about the data being passed. 

Maybe the solution here will become apparent once we resolve #135. We may remove the `feature_labels` operation altogether, or set the list of `col_id`s during a training phase.

montoyjh:
Should we close this for now and I'll resubmit once the dust settles on #135?

computron:
Up to you how you want to do it, but yes we should wait until the refactoring of BaseFeaturizer is complete. It will likely solve the issues faced

computron:
@montoyjh is now a good time to revisit this? BaseFeaturizer refactor is done

montoyjh:
Yep, I'm looking it over now, and it doesn't really seem to change much that's relevant to my implementation of the featurizer.  I'm still facing an issue where the static nature of the 'feature_labels' method doesn't quite fit with the notion of flexibly providing inputs to a set of functions which will provide output features of variable length.  I resolved this by keeping my addition of "label_props" as a kwarg to featurize dataframe, such that one can override featurize dataframe simply by specifying some set of parameters for the label generation and supplying them to a super(self, BaseFeaturizer).featurize_dataframe call.

WardLT:
What about using the `fit` operation to determine the number of features to generate? You could add "label_props" as a kwarg to fit so that the feature names are set during the fit operation too, and then you do not have to alter the signature of `feature_labels`.

montoyjh:
Just to clarify, are you suggesting that I use the fit function to alter attributes of the featurizer itself and then yield those in the feature_labels method?

WardLT:
That's right. The two attributes that will change are which/how many features to expect for `featurize`, which controls how many features are generated. Does that help clear things up?

montoyjh:
I've gone ahead to use fit and resolved the differences in function signatures, but I wanted to add a bit more feedback on the issue.  I'm not a fan of this design pattern for this purpose.  In my mind a "fit" method strongly suggests we're altering some mathematical component of how inputs lead to outputs, e. g. a coefficient in a linear regression.

We're proposing here to use it for bookkeeping to avoid altering the signature of a method, despite the fact that the results of the method need to be (and still will be, in this scheme) altered.  If we're dedicated to having "feature_labels" be a static place where we store information relevant to the data generated by a featurizer, it shouldn't be able to change based on the identity of the input dataframe labels, it probably shouldn't be a method, and we should have a separate procedure that actually assigns the labels of a featurized dataframe that can include information from the dataframe columns.

WardLT:
I think these are fair points and thanks for the detailed reply, @montoyjh! I agree, getting FunctionFeaturizer to mesh with BaseFeaturizer is a little trickier than the other ones, as it takes a dataframe as input rather than a material. That said, I still think we can make it work and work sanely without modifying the behavior of any methods inherited from BaseFeaturizer.

The choice of the word `fit` is somewhat misleading and is a result of us conforming to scikit-learn's BaseEstimator/Transformer API. It is better to think of the "fit" operation as "configure." The main point of the fit operation is to set any options of the featurizer that are not known when initializing the featurizer and are dependent on a training set. We separate "fit" from the featurization operations to prevent the nature of features changing between when a featurizer is used on multiple datasets (e.g., a training and test set). The number of features or how they are computed changing between training and test sets, of course, would be a huge issue. 

FunctionFeaturizer does not really need the `fit` operation. As shown in how [PolynomialFeatures](https://github.com/scikit-learn/scikit-learn/blob/a24c8b46/sklearn/preprocessing/data.py#L1180) - a close analogy to your class - is implemented, `fit` only sets the number of inputs for checking purposes and computes the number of outputs for use when allocating the output array. While these are helpful, if you make the assumption that your user will always pass the same type of data to `transform`/`featurize` or need to optimize memory usage then `fit` can be eliminated.

I you agree with this logic, here is my proposal for changes FunctionFeaturizer:

- `fit` - No changes to base functionality. It could be made more robust by making `input_feature_names` optional by generating fake names (e.g., "x0", "x1") if none are provided. Additionally, it would be nice to have type checking so that if `X` is a DataFrame, we read the column labels from DataFrame.

- `featurize_dataframe` - Remove call to `fit`, so that `featurize_dataframe` will not change the state of the featurizer. Here, the reasoning is to provide some notion that the featurizer is configured on the training dataset and then will not change behavior on any future calls.

If I understand right, then the biggest change to using FunctionFeaturizer will be a need to call `fit` before `featurize_dataframe`. I think this is acceptable, as this is how scikit-learn's feature expansion tools (i.e., PolynomialFeatures) is used.

montoyjh:
> If I understand right, then the biggest change to using FunctionFeaturizer will be a need to call fit before featurize_dataframe. I think this is acceptable, as this is how scikit-learn's feature expansion tools (i.e., PolynomialFeatures) is used.

I strongly disagree.  It's not intuitive, it's error prone, and it adds an unnecessary burden on the user, for whom we're trying to simplify this process.  I have another solution though.

In my latest update,`feature_labels` just returns None, since `feature_labels` doesn't really make sense for this class anyway.  Pandas assigns numerical columns if feature_labels is None, which are are reassigned in postprocessing in an overridden fit_dataframe.  I do have to slightly modify the duplicate-checking procedure in BaseFeaturizer.featurize_dataframe in order to prevent errors since column labels are being handled in the overridden featurize_dataframe.  There's no overriding of fit and the user only has to specify column labels in featurize_dataframe once.  This requires minimal user input and doesn't mess with any class design patterns.

WardLT:
I like this version without `fit` and `feature_labels` turned off the best, and am fine with merging it after addressing:

- Should we change the docstring of `featurize_dataframe` to describe what the operation does rather than how it was implemented (i.e., just copy the one from BaseFeaturizer)?
- I am concerned about the compatibility of "`new_df.columns = df.columns.tolist() + new_col_names`" with Python 2.7 (and possibly before 3.6, which introduced using insertion-ordered dicts by default). My worry is that the new columns will always be the last ones in the dataframe after running `featurize_dataframe`. If this is indeed a problem [I'll do some tests tomorrow, don't have time now], I have a few ideas for addressing it.

To make sure we're in consensus, my reasoning for why I like it is below.

Thanks for putting this together. I really like how flexible FunctionFeaturizer is, and I look forward to trying it out soon!

---------

The main advantage I could think of for storing the feature names/number of input features with `fit` is providing warnings if the number of features changes, which does not seem very useful. I agree, it is much nicer to be able to avoid using `fit` altogether. 

The `featurize_dataframe` still works in the same way as in the other featurizers: what features and how they are computed will not change after subsequent evaluations (unless the input features are changed, which should certainly change the output). So, I'm on board with overriding it.

This implementation will also work nicely in Pipeline objects. The way I plan to use this to put it in a pipeline after other featurizers. Something like:

```python
p = Pipeline([
    ('featurizers', ElementProperty().from_preset('magpie')),
    ('expander', FunctionFeaturizer())
])
```

In short, I could still call `p.fit_transform` and expect it to return features, and `transform` and expect none of the configurations of the featurizers to change.

montoyjh:
Modified the docstring.  I was initially concerned about the second issue as well, the only place I think it might break is in pd.concat, which I remember used to reorder columns (I think this changed at some point in pandas, though).  I added a small piece of code in the operation of non-in-place featurize_dataframe that ensures the columns are ordered "old_columns" + "new_columns" that should correspond strictly to the new column ordering.

matminer/featurizers/function.py
```diff
@@ -0,0 +1,240 @@
...
+
+from matminer.featurizers.base import BaseFeaturizer
+
+
+# Default expressions to include in function featurizer
+default_exps = ["x", "1/x", "sqrt(x)", "1/sqrt(x)", "x**2", "x**-2", "x**3",
+                "x**-3", "log(x)", "1/log(x)", "exp(x)", "exp(-x)"]
+
+
+# TODO: feature labels might be latexified
```
WardLT: Is this already solved? If so, please remove this todo.

matminer/featurizers/function.py
```diff
@@ -0,0 +1,240 @@
...
+# TODO: parallelization of substitution, think this might be
```
WardLT: I agree, parallelization over rows is already implemented and probably preferred.

matminer/featurizers/function.py
```diff
@@ -0,0 +1,240 @@
...
+
+        """
+        self.expressions = expressions or default_exps
+        self.multi_feature_depth = multi_feature_depth
+        self.combo_function = combo_function or np.prod
+        self.latexify_labels = latexify_labels
+        self.postprocess = postprocess or float
+
+        # Generate lists of sympy expressions keyed by number of features
+        self.exp_dict = OrderedDict(
```
WardLT: It seems problematic to have this in the constructor (e.g., what if someone changes the list of expressions or multifeature depth after creation?).

That said, I'm not sure of where else to put it. It seems costly to put it inside `_exp_iter`. Any thoughts?

matminer/featurizers/function.py
```diff
@@ -0,0 +1,229 @@
...
+                    # TODO: this is a workaround for the problem
+                    # TODO: postprocessing functional incompatility,
+                    # TODO: e. g. sqrt(-1), 1 / 0
+                    try:
+                        yield postprocess(exp.subs(subs_dict))
+                    except (TypeError, ValueError):
+                        yield None
+
+    def citations(self):
+        return ["@article{Ramprasad2017,"
```
WardLT: I think we should cite Ghiringhelli et al. in place of or in addition to this reference.

WardLT:
Besides the inline comments I've made, the code and tests look good to me.

computron:
For the citation, the specific functional forms we used are from the Ramprasad paper. I agree that Ghiringelli was the first to report the technique in materials science but it is a much older idea than them to combine individual descriptors into mini equations. They didn’t even write in the paper what functional forms were tried (if I remember correctly). I would maybe put Ghiringelli as a secondary reference but would also be OK with leaving it out; the featurizer is in the same spirit as things like PolynomialFeaturizer and I don’t think that combining descriptors was really a new technique.

WardLT:
You're right. On further review of the Ghiringhelli paper, FunctionFeaturizer does more closely resemble the Ramprasad paper. I'm now fine with keeping only the Ramprasad reference. Thanks for point that out!

matminer/featurizers/function.py
montoyjh: I made exp_dict a property so it gets evaluated on the fly (also the combo_function wasn't getting passed to the expression generator, so that's fixed as well).  I don't think it'll be too costly unless someone decides they want to use like 50 functions or a really extreme multi-feature depth, and at that point the majority of the cost will be in iterating over the combinations of input features anyway.  It should be simple to add a lru_cached method for generation, if we find that performance is an issue, but I'd opt for simplicity at this point.

WardLT:
Looks good to me. Unless anyone objects, I'll merge later this morning.

Issue #128, opened 2018-02-09
too many presets


computron:
Some of the featurizers have a large amount of presets, all of which are doing essentially the same thing. Here's one from a recent PR, although there were many before this of the same form (some committed by me).

```
+        if preset == "VoronoiNN":
 +            return ChemicalSRO(VoronoiNN())
 +        elif preset == "JMolNN":
 +            return ChemicalSRO(JMolNN())
 +        elif preset == "MinimumDistanceNN":
 +            return ChemicalSRO(MinimumDistanceNN())
 +        elif preset == "MinimumOKeeffeNN":
 +            return ChemicalSRO(MinimumOKeeffeNN())
 +        elif preset == "MinimumVIRENN":
 +            return ChemicalSRO(MinimumVIRENN())
 +        else:
 +            raise RuntimeError('Unknown preset.')
```

We should:

1. Clean up the code to shorten it, using something like this:
https://stackoverflow.com/questions/4821104/python-dynamic-instantiation-from-string-name-of-a-class-in-dynamically-imported

It will also allow us to automatically support new NearNeighbor algorithms ,etc.

2. Consider having more guidance to presets. e.g. instead of just a billion presets, to do more like what the ElementProperty presets and do and just give a few "vetted" suggestions.

nisse3000:
@computron, sounds sensible.  I'll take care about the example you gave.

nisse3000:
@computron, done (in my fork).  I'm waiting with the pull request until @shyuep has released the next pymatgen version so that the additions of new and optimized OPs will also be part of that PR.

nisse3000:
See PR #136.

computron:
Closing for now. Note that another option is that if a preset is "strong" enough (we really advocate for that preset, or there is a paper that advocates for that combination of settings that can be cited) we can create a new featurizer for it. That new featurizer can basically just call the old featurizer with the chosen parameter set. It should help in discoverability of that preset.

PR #133, opened 2018-02-10
BagofBonds update

ardunn:
## Summary
- Performance improvements
- Made BoB presets better as mentioned in #128 
- Added `allowed_bonds` in BoB
- Added `approx_bonds` in BoB
- originally brought up in #127 

## Details
- Define `allowed_bonds` to limit the kinds of bonds which can be returned. If a structure has a forbidden bond, skips it and returns all other bonds normally (including bad bond values, where the forbidden list has bonds the structure being featurized does not). This can result in some rows having bond fractions summing to less than 1. 

- Allowed bonds can be defined by the list of feature labels a BoB object returns after featurization (`bob.feature_labels()`)

- If you do not want `allowed_bonds` to skip forbidden bonds, set `approx_bonds` to true to approximate forbidden bonds with chemically similar bonds found in `allowed_bonds`. Chemically similar bonds are determined by finding the bond in `allowed_bonds` with the smallest "distance" to the forbidden bond. Approximate bonds with equal distances to the forbidden bond divide the bond fraction equally among them. 

- The "distance" between bonds is determined by converting each bond into a 2-tuple of mendeleev numbers (MN) from their species, so each bond gets an (x,y) coordinate in MN space. A bond's x and y are swapped automatically if necessary (ie if we are comparing (O-Li bond) and (Na-O bond), making sure we calculate the MN distance between (Li and Na) and (O and O), _not_ (Li and O) and (O and Na)). Then the distance between (x1, y1) and (x2, y2) is Euclidean. 

## Examples
Get dataframe of some Na-O and Li-O compounds
```python
df = mpr.get_dataframe(criteria={"material_id": {"$in": ['mp-1960', 'mp-841', 'mp-1018789', 'mp-2352', 'mp-1901']}}, properties=['structure', 'spacegroup.number', 'pretty_formula'])
```
Featurize only the LiO compounds, for training:
```python
bob = BagofBonds.from_preset("MinimumDistanceNN")
print(bob.featurize_dataframe(df.iloc[[0, 2, 4]], 'structure'))
```
```
                                                     structure  spacegroup.number pretty_formula  Li - Li bond frac.  Li - O bond frac.  O - O bond frac.
material_id                                                                                                                                              
mp-1018789   {'lattice': {'a': 3.98566942, 'c': 2.96110099,...                 58           LiO2                 0.0           0.750000          0.250000
mp-1960      {'lattice': {'a': 3.291071792359756, 'c': 3.29...                225           Li2O                 0.0           1.000000          0.000000
mp-841       {'lattice': {'a': 3.18036198, 'c': 7.70352187,...                194          Li2O2                 0.0           0.857143          0.142857
```
Set the allowed bonds to the feature_labels of the previous featurization. Featurize the dataframe allowing only bonds containing Li and O, say, for testing:
```python
allowed = bob.feature_labels()
bob.allowed_bonds = allowed
print(bob.featurize_dataframe(df.iloc[[1, 3]], 'structure'))
```
```
                                                     structure  spacegroup.number pretty_formula  Li - Li bond frac.  Li - O bond frac.  O - O bond frac.
material_id                                                                                                                                              
mp-1901      {'lattice': {'a': 3.47402655, 'c': 5.67466155,...                 58           NaO2                 NaN                NaN              0.25
mp-2352      {'lattice': {'a': 3.9563243674727504, 'c': 3.9...                225           Na2O                 NaN                NaN              0.00
```
The Na-O and Na-Na bonds were skipped, because they was not in allowed_bonds. The `NaN` values are the bad_bond_values, which can be set with `bob.bbv`. 
If we do not want the Na-O and Na-Na bonds skipped, use `approx_bonds` to approximate these bonds with their most chemically similar, Li-O and Li-Li. 
```python
bob.approx_bonds = True
print(bob.featurize_dataframe(df.iloc[[1, 3]], 'structure'))
```
```
                                                     structure  spacegroup.number pretty_formula  Li - Li bond frac.  Li - O bond frac.  O - O bond frac.
material_id                                                                                                                                              
mp-1901      {'lattice': {'a': 3.47402655, 'c': 5.67466155,...                 58           NaO2                 0.0               0.75              0.25
mp-2352      {'lattice': {'a': 3.9563243674727504, 'c': 3.9...                225           Na2O                 0.0               1.00              0.00
```

### DIFF
--- a/matminer/featurizers/structure.py
+++ b/matminer/featurizers/structure.py
@@ -1225,16 +1225,28 @@ class BagofBonds(BaseFeaturizer):
             if lb in self.unified_bonds:
                 ubonds_data[lb] += local_bonds[lb]
             else:
-                species = self._species_from_bondstr(lb)
+                lbs = self._species_from_bondstr(lb)
 
                 nearest = []
                 d_min = None
                 for ubs in ubonds_species.keys():
-                    dm0 = float(ubs[0].element.mendeleev_no -
-                                species[0].element.mendeleev_no)
-                    dm1 = float(ubs[1].element.mendeleev_no -
-                                species[1].element.mendeleev_no)
-                    d = (dm0**2.0 + dm1**2.0)**0.5
+
+                    u_mends = [j.element.mendeleev_no for j in ubs]
+                    l_mends = [j.element.mendeleev_no for j in lbs]
+
+                    # The distance between bonds is euclidean. To get a good
+                    # measure of the coordinate between mendeleev numbers for
+                    # each specie, we use the minumum difference. ie, for
+                    # finding the distance between Na-O and O-Li, we would
+                    # not want the distance between (Na and O) and (O and Li),
+                    # we want the distance between (Na and Li) and (O and O).
+                    d1 = min([abs(u_mends[0] - l_mends[0]),
+                              abs(u_mends[1] - l_mends[0])])
+
+                    d2 = min([abs(u_mends[0] - l_mends[1]),
+                              abs(u_mends[1] - l_mends[1])])
+
+                    d = (d1**2.0 + d2**2.0)**0.5
                     if not d_min:
                         d_min = d
                         nearest = [ubs]
@@ -1273,6 +1285,7 @@ class BagofBonds(BaseFeaturizer):
 
         bond_types = tuple(self.enumerate_bonds(s))
         bonds = {k: 0.0 for k in bond_types}
+        tot_bonds = 0.0
 
         # If featurize is being called from a dataframe or featurize_many,
         # a comprehensize 'unified' bond list is created. The following code
@@ -1310,13 +1323,14 @@ class BagofBonds(BaseFeaturizer):
                 b = btup[0] + ' - ' + btup[1]
                 # The bond will not be in bonds if it is a forbidden bond
                 # (when a local bond is not in allowed_bonds)
+                tot_bonds += 1.0
                 if b in bonds:
                     bonds[b] += 1.0
 
         if self.approx_bonds:
             bonds = self._approximate_bonds(bonds)
 
-        tot_bonds = sum(v for v in bonds.values() if not np.isnan(v))
+        # tot_bonds = sum(v for v in bonds.values() if not np.isnan(v))
 
         # If allowed_bonds caused no bonds to be present, all bonds will be 0.
         # Prevent division by zero error.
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

Some comments:
- The way the featurizer is intended to work with most models, probably for most cases you want the "bad bond value" to just be zero, not NaN. i.e., there is no Li-Li bond component to Na2O. It doesn't matter that it's not possible to have that component.
- If you define the element order in the bond in order of Mendeelev number (lowest element Mendeleev number first, highest Mendeelev number second), I think you can have a consistent bond label (no "Li-O" vs "O-Li") *plus* not need to check multiple combinations for distance (just take the direct distance).

A more general comment is that multiple of these featurizers require knowing in advance all the entries of the dataframe in order to determine the feature columns. It might be better to look at the ``fit_transform`` methods in scikit-learn often used for text mining as an inspiration rather than trying to cram things into the current Featurizer format. I'll submit a separate issue about this, just mentioning it here.

ardunn:
@computron Both the bulleted comments have been addressed with the latest commits.

One possible workaround for BoB for the more general comment is to make `enumerate_bonds` and `enumerate_all_bonds` static methods or utility functions. Then there would be no need for overriding `featurize_dataframe`, if a list of bonds was passed into BoB's constructor as a required argument:


```python
bonds_list = BagofBonds.enumerate_all_bonds(structure_list)
bob = BagofBonds(bonds_list, nn=VoronoiNN())
bob.featurize_dataframe(df, 'structures') # featurizes according to bonds_list
```

However this kind of adds an extra step to the usage of the featurizer.

computron:
Merging although we might still need to redo later

Note that you shouldn't change the operation of Featurizer(). Once you have a contract, you should stick to it - otherwise, don't have a contract. We can define a different contract for things that require pre-processing, but we shouldn't violate expectations for a given contract.

ardunn:
Sure, however I don't think any implementation of BoB will change the operation of Featurizer. I was suggesting that if we wanted to avoid the issues in #135 , we could require that a BoB object is created from a list of the desired bonds passed as an argument to the constructor. Having a function to create such a list could be done by moving BoB's `enumerate_all_bonds` to `utils`.  Then when a BoB object is created, it's feature labels are 100% set before any featurization and there will be no `super()` nonsense or bad `feature_labels` calls. This seems like the most logical solution to me (at least for BoB).

Issue #135, opened 2018-02-13
BaseFeaturizer() doesn't support certain types of featurizers well


computron:
Many featurizers don't have a set number or type of features that they return - the features depends on the inut. For example, the BagofBonds featurizers (currently as a PR) will return one value for each potential bond combination in the material. For Li2O this would mean Li-Li, Li-O, and O-O. But for Li (metal) the only feature is Li-Li. Or for Na2S it is Na-Na, Na-S, S-S. So the features are different for each entry. There are multiple featurizers like this, like ChemicalSRO.

There are two issues here:
1. You get this clunky thing where feature_labels() only really works well after you've called featurize(), and is only really pertinent to the most recent featurize() call. This makes the Featurizer a stateful object which is less preferable and more prone to breakage.
2. Sometimes, you need to override featurize_dataframe() to get this to work.

It would be better if someone could design an improvement to BaseFeaturize() to support these kinds of features. I should say that text mining handles this kind of use case all the time. For example, the ``CountVectorizer`` in scikit-learn will return one "feature" for every distinct word in a block of text, and the features will be different for each text block. The way to do this is to use the ``fit_transform()`` method in scikit-learn.

I am wondering if something similar is needed / helpful for some of our featurizers.

WardLT:
I think this is a good point @computron.

Could we support this case by adding a 'fit' operation that takes a list of
entries and then sets the necessary parameters of the featurization (e.g.,
which bonds to list)?

That would also conceptionally separate actions that prepare a featurizer
to be used from those used when running/training a model.

Another option that I was considering while reading these PRs (sorry that I
have not had time to contribute) was to advocate that all options that
affect which features are computed should go in the initializer.




On Feb 12, 2018 6:08 PM, "Anubhav Jain" <notifications@github.com> wrote:

Many featurizers don't have a set number or type of features that they
return - the features depends on the inut. For example, the BagofBonds
featurizers (currently as a PR) will return one value for each potential
bond combination in the material. For Li2O this would mean Li-Li, Li-O, and
O-O. But for Li (metal) the only feature is Li-Li. Or for Na2S it is Na-Na,
Na-S, S-S. So the features are different for each entry. There are multiple
featurizers like this, like ChemicalSRO.

There are two issues here:

   1. You get this clunky thing where feature_labels() only really works
   well after you've called featurize(), and is only really pertinent to the
   most recent featurize() call. This makes the Featurizer a stateful object
   which is less preferable and more prone to breakage.
[...]

PR #138, opened 2018-02-16
GRDF, AFS

dyllamt:
Added:
- GeneralizedRadialDistributionFunction class with three operation modes
- tests for said class

Issues:
- Fails tests with _featurize_df()_ because two modes output heterogeneous features for different compounds. However, this is the desired behavior.
- Need to allow for heterogeneous features in BaseFeaturizer using DataFrames instead of relying on separate lists of _features_ and _feature_labels_.

Should I take out the two more complex modes of operation? Or rework the BaseFeaturizer? BaseFeaturizer can be reworked by having the returns of _featurize()_ be DataFrames and concating.


computron:
@dyllamt for the two issues you mentioned you should follow updates to Issue #135 

The next thing I intend to do for matminer is to think hard about that issue since it's tripping up a lot of featurizers

dyllamt:
Added:
- `HeterogeneousFeaturizer` a child class of `BaseFeaturizer`. This takes care of problems with having heterogeneous features for multiple input samples, although all featurizers could follow this paradigm since it eliminates redundancy in looping over variables for `featurize` and `feature_labels`.
- `AngularFourierSeries`--a featurizer which takes into account angular and distance information in a structure.
- tests for AFS

Total pull request is now:
- `HeterogeneousFeaturizer`--covers featurizing featurizers that may not always return the same features
- `GeneralizedRadialDistributionFunction`--three types of featurizing behavior including site-specific and site-averaged bahavior
- `AngularFourierSeries`--four types of behavior depending on `init` arguments

If this pull request is accepted:
- GRDF can replace `RadialDistributionFunction` featurizer since RDF is one of the options for GRDF
- The `PartialRadialDistributionFunction` needs to be revisited, since it is a special subcase of GRDF, but it would be nice to have this as a convenience class. (semantically, PRDF is now doing a element-wise pairing and not a site-wise pairing)

WardLT:
I've got a few questions/comments about this PR. Would you mind waiting to
pull? I should have a chance to write them up early this afternoon.

WardLT:
First of all, thanks for putting all of this together!

A few questions/comments:

- If the Partial GRDF returns a list features for each site, would it make more sense to implement as a separate, "site" featurizer (e.g., in `site.py`)? I think this would alleviate our problem of having "pairwise_RDF" and "GRDF" returning different shapes of features (and, generally, reduce the complexity of the `featurize` method). [I'm also curious if this change would eliminate the need for Heterogeneous featurizer]

- My impression is that "Partial" RDF (PRDF) is broken down into pairs of elements? This is show PRDFs are implemented in [Schutt 2014[(https://journals.aps.org/prb/abstract/10.1103/PhysRevB.89.205118), which is what Seko references as an example of PRDFs. In short, I don't think we should change `PartialRadialDistributionFunction` to do pairwise sites or have the Partial GRDF return a list of pairwise sites. Could you justify why you implemented the GRDF to yield site-wise RDFs?

- Would it make sense to have the `bins` parameter contain both the function and a name for that function? This could give more meaningful feature names. (E.g., `bins = [('gaussian b=2', lambda x: exp((x - 2)**2))]`

- Could we create defaults that generate the same features as [Seko 2017](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.95.144110)? 

- Is the RDF as implemented in `RadialDistributionFunction` than the one in your general implementation? It is probably not worth holding up the PR to answer this, but I'm curious whether the RDF is worth keeping separate for performance reasons (`np.hist` is pretty quick). 

- I have similar comments about the site-wise pairing for AFS. So, depending on what consensus we reach about them, it might be worth revisiting them.

dyllamt:
- I hadn't thought about putting the GRDF and partial-GRDF implementations in the site featurizer before! You're right that it does fix the problems of having a heterogeneous feature set for different compounds. The problems I see with this alternative implementation is that a user is not likely to be interested in a single site in the crystal and that you split-up one algorithm into multiple places in Matminer. That being said, I am happy to move those implementations into the site featurizers, I just want to debate the advantages/disadvantages of doing so. With this in mind, many of the structural featurizers could be moved to the site featurizer module.

- I think implementing the pairwise-GRDF as a site-wise pairing vs element-wise pairing offers greater flexibility. In short, a site-wise pairing allows you to do both structural comparisons keeping the chemistry constant and chemistry comparisons keeping the crystal structure constant (an element-wise pairing only allows you to do the former). My research is in bandstructure, and in my experience it is better to try and understand this property within a certain family of compounds, where only the site chemistry is changing, and maybe there are slight structural distortions.

- Absolutely, I will add this once we work out the other details for this pull request. This will make the output of the featurizers much easier to read!

- I couldn't find the exact combinations of bin functionals the Seko paper used. They specify a number of functionals (e.g. they used 20 gaussians) and the cutoff radius they used but not the std dev and centers for the gaussians. I could set the default cutoff radius to what they used?

- I'll do a speed test. I'll have an answer by our next round of discussion.

dyllamt:
@computron ,

I moved the featurizers to the site module. Good news is that the `fit` method @WardLT implemented works wonderfully for the GRDF!

computron:
see failing tests ...

dyllamt:
My mistake, forgot to remove a broken import.

Fixed now.

matminer/featurizers/site.py
```diff
@@ -1,13 +1,17 @@
 from __future__ import division
 
-"""Features that describe the local environment of a single atom
+"""
+Features that describe the local environment of a single atom. Note that
+structural featues can be constructed from a combination of site features from
+every stite in the structure.
```
computron: spelling: site

matminer/featurizers/site.py
```diff
@@ -1307,3 +1314,247 @@ def citations(self):
...
+        features. Requires `fit` for determining the max number of sites for
+
+    The GRDF is a generalization of the partial radial distribution function
+    (PRDF). In contrast with the PRDF, the bins of the GRDF are not mutually-
+    exclusive and need not carry a constant weight of 1. The PRDF is a case of
+    the GRDF when the bins are rectangular functions. Examples of other
+    functions to use with the GRDF are gaussians, trig, and bessel functions.
+
+    Args:
+        bins:   (list of touples) a list of (str, functions). The str is a text
```
computron: spelling: tuples

matminer/featurizers/site.py
```diff
@@ -1307,3 +1314,247 @@ def citations(self):
...
+    product of distance functionals (g_n, g_n') between two pairs of atoms
+    (sharing the common central site) and the cosine of the angle between the
+    two pairs. The AFS is a 2-dimentional feature (the axes are g_n, g_n').
+
+    Examples of distance functionals are square functions, gaussians, trig
+    functions, and bessel functions. An example for gaussians:
+        lambda d: exp( -(d - d_n)**2 ), where d_n is the coefficient for g_n
+
+    Args:
+        bins:   (list of touples) a list of (str, functions). The str is a text
```
computron: spelling: tuples

matminer/featurizers/site.py
```diff
@@ -1307,3 +1314,247 @@ def citations(self):
...
+        Returns:
+            Flattened list of AFS values. the list order is:
+                g_n g_n'
+        """
+
+        if not struct.is_ordered:
+            raise ValueError("Disordered structure support not built yet")
+
+        # Generate list of neighbor position vectors (relative to central
+        # atom) and distances from each central site as touples
```
computron: spelling: again touples (maybe just find and replace?)

computron:
Ok just some spelling things now, after that I'll merge.

Are you able to remove the "not_ready_to_merge" label when you have the merge ready?

dyllamt:
@computron,

I ran a spell check on the site module.

I poked around trying to change the label, but I don't think I'm authorized to make those kinds of changes.

computron:
I don't think the changes were committed? The PR is identical to before

dyllamt:
I thought changes were made in commit 7944a85, but I recommitted the changes and it looks like they are reflected on github.

computron:
There are still instances of ``touples``, e.g. ``neighbor_touples``

dyllamt:
They should all be sorted out now. I had only updated the doc spelling before.

computron:
The tests still use ``touples``.

dyllamt:
I'm sorry about all the spelling mistakes in the code. I think they are all fixed now.

computron:
Thanks!

PR #140, opened 2018-02-17
Updated PlotlyFig

ardunn:
### Summary

- Fix `histogram` bin size problem
- Fix `heatmap `
- Update docs
- Added `parallel_coordinates` plotter

### Todo soon:

- Add `font_scale`
- Fix dict representation of layout in init with `None` types
- Fix violin, so there are no multiple colors by default
- Make `plot_mode` set through each method
- Major cleanup + docs


computron:
blindly merging, thanks!

PR #155, opened 2018-02-19
Bit of cleanup/standardization for data retrieval

montoyjh:
In the process of refactoring tutorials, found that there was some inconsistency in the data retrieval tools.  This PR resolves some of that, and includes a bit of cleanup.  The primary thing is that each data retrieval tool has a "get_dataframe" method that interfaces directly to the data retrieval API.  I think there could still be more uniformity on these methods, but that'll have to be a TODO for now.

### DIFF
--- a/matminer/data_retrieval/retrieve_Citrine.py
+++ b/matminer/data_retrieval/retrieve_Citrine.py
@@ -48,7 +48,10 @@ class CitrineDataRetrieval:
                       from_record=None, data_set_id=None, max_results=None,
                       show_columns=None):
         """
-        Converts list of json/pifs to a Pandas dataframe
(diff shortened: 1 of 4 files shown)


computron:
Thanks - would be worth taking a look at the Codacy issues tho. I will fix them after merging

WardLT:
Should we make all of the classes inherit from a single base class? For now, it seems like the main thing the base class would have is the `get_dataframe` abstract method.

computron:
@WardLT Sure

(note the comment is probably best as a new issue referring to the PR, otherwise I have trouble tracking down all "open" things to address)

WardLT:
Cool. I'll open up an issue (now, and in the future)

PR #156, opened 2018-02-19
WIP: Refactor BaseFeaturizer

WardLT:
Introduces a draft BaseFeaturizer class (discussion in #135) that follows the API for scikit-learn's Transformer classes. 

I also implemented PartialRadialDistributionFunction as an example, and wrote some documentation on how to implement a BaseFeaturizer.

A few questions and issues that I already have:

- Is it a problem that we have both `transform` and `featurize_many`? I kept both because featurize_many provides the parallelism and error-handling options, and the `transform` API from scikit-learn does not support any additional `kwargs`. `transform` is a restricted wrapper over `featurize_many`.

- Parallelism and `transform`: Sklearn classes generally support parallelism by including a `n_jobs` class attribute. Adding `n_jobs` to the initialize means all new subclasses should include `n_jobs` in their kwargs and call the base classes's constructor. I'm in favor of doing this. 

- Does it make sense to put the implementation guide I wrote in the class documentation for BaseFeaturizer? 

To do list:

- Adjust all classes that override `featurize_dataframe` to use this new class. (They probably benefit from the `fit` operation)
- Write a guide for documenting the features which each featurizer generates. I think we should have a standard way for doing this.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -96,6 +96,14 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
     able to reproduce the feature, just to get an idea what it is.
     """
 
+    def set_n_jobs(self, n_jobs):
+        """Set the number of threads for this """
+        self._n_jobs = n_jobs
+
+    @property
+    def n_jobs(self):
+        return self._n_jobs if hasattr(self, '_n_jobs') else 1
+
     def fit(self, X, y=None, **fit_kwargs):
         """Update the parameters of this featurizer based on available data
 
@@ -108,8 +116,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
         return self.featurize_many(X, ignore_errors=True)
 
-    def featurize_dataframe(self, df, col_id, ignore_errors=False,
-                            inplace=True, n_jobs=1):
+    def featurize_dataframe(self, df, col_id, ignore_errors=False, inplace=True):
         """
         Compute features for all entries contained in input dataframe.
 
@@ -122,10 +129,6 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
                 exceptions are thrown if True. If False, exceptions
                 are thrown as normal.
             inplace (bool): Whether to add new columns to input dataframe (df)
-            n_jobs (int): Number of parallel processes to execute when
-                featurizing the dataframe. If None, automatically determines the
-                number of processing cores on the system and sets n_procs to
-                this number.
 
         Returns:
             updated dataframe.
@@ -144,7 +147,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
                 raise ValueError('"{}" exists in input dataframe'.format(col))
 
         # Compute the features
-        features = self.featurize_many(df[col_id].values, n_jobs, ignore_errors)
+        features = self.featurize_many(df[col_id].values, ignore_errors)
 
         # Create dataframe with the new features
         res = pd.DataFrame(features, index=df.index, columns=labels)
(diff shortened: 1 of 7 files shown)


computron:
I'd say:

1. For now I'd say keeping ``transform`` and ``featurize_many()`` is OK.
2. If I understand correctly, this means every single featurizer (i.e. dozens so far) will need to (i) add n_jobs as a parameter in the constructor and (ii) need to call the superclass.__init__(n_jobs=n_jobs) in the constructor. This seems just onerous to me given that most people writing featurizers don't want to worry about n_jobs at all. I'd prefer something else than going through every featurizer and adding n_jobs and making sure the constructor calls the superclass (in my experience, this is the type of developer friction that leads to frustration and mistakes). We've done the work of making n_jobs painless from the developer perspective since once you implement ``featurize()``, ``featurize_many()`` basically needs to be parallelizable. This might be bending the rules of sklearn's BaseEstimator, but we could for example add a ``set_njobs`` method to BaseFeaturizer itself that sets an internal variable. Then all downstream featurizers will automatically inherit it and the ability to set n_jobs without the constructor param. If a featurizer wants to enforce n_jobs=1, it can call set_njobs=1 in the constructor and then override the set_njobs() function with an error message.
3. Yes, I think the implementation guide is fine there!

computron:
Upon further thought I am not sure that my suggestion for (2) is the best way forward, but I would still like to avoid having every featurizer need to add n_jobs and call the superclass in the constructor if at all possible.

WardLT:
What do you not like about the solution for 2? I do agree with your argument about avoiding `n_jobs` in all of the constructors. One problem I could see with 2 is that we'd need some logic to check whether the number of threads used in `featurize_many` is read from the default value of the kwarg pass to that argument, the `n_jobs` class attribute, or the value a user passes to that function. Having three ways for the number of threads to be determined does sound problematic.

Should we just do nothing about `n_jobs`? All that we lose for not having the ability to set `n_jobs` as a class attribute is that `transform` will be serial, which does not seem like a big loss. Users are still free to featurize in parallel using `featurize_dataframe` and `featurize_many`, and the main use for `transform` is if users put the featurizer in a `Pipeline`. The main reason I'd put my featurizer in a Pipeline is to optimize the feature generation using `GridSearchCV` (having separate calls to compute features and run a model is not a big enough deal for me to make a `Pipeline`). GridSearchCV can test multiple parameterizations in parallel, so we don't necessarily need to parallel feature generation for that use case. Overall, having `transform` be serial and maintaining the ability for users to parallelize `featurize_dataframe` could keep many people happy.

computron:
Maybe a revised version of (2) is best:

- we create a private var called ``_n_jobs``
- we have a getter function called ``n_jobs()`` with the standard ``@property`` decorator. All the logic of determining the actual number of jobs goes in this function. That way, the logic doesn't need to be duplicated everywhere that is just trying to access the number of jobs.
- like before, we have a setter function called ``set_n_jobs()`` that is the only approved way to change the state of the private ``_n_jobs`` variable.

Thus, with BaseFeaturizer having an explicit ``n_jobs`` attribute (function) and ``set_n_jobs()`` function everything feels pretty "official". 

I see your point about having transform just be serial, but I'm afraid someone will expect a certain performance based on running the featurizers "normally" and then be surprised when that same set of features gives a completely different performance when part of a simple pipeline (i.e., one that doesn't multiprocess some other way, e.g. via Grid Search).

WardLT:
What you propose sounds reasonable to me. Does 4acfbe6fa22182b7214af29ef993308444455b90 capture your idea?

computron:
Yes! I think there is a separate discussion on what the default n_jobs should be, but I think we can leave that for a separate issue if you want.

Is the PR ready for review and merge?

WardLT:
I think it's ready for review.

WardLT:
Oh, and I'm fine with moving a discussion of the default `n_jobs` to another issue.

computron:
Thanks!

Issue #154, opened 2018-02-18
Should featurize_dataframe have an exceptions_encountered() variable?


computron:
Let's say you run featurize_dataframe with ignore_errors=True. After finishing featurization, you might want to just see what errors were encountered. If the BaseFeaturizer had an attribute/function called exceptions_encountered() perhaps it could return all the exceptions encountered during the previous featurize_dataframe() call.

Implementation would be simple - just create the variable, reset it at the beginning of the featurize_dataframe, and use traceback module to add the string version of each Exception to a list.

Another way to do it would be to add another parameter called ``return_errors`` to featurize_dataframe(). If ``ignore_errors`` and ``return_errors`` are both true, then featurize_dataframe would change its return type to be a tuple of the new dataframe and a list of errors encountered. This would avoid having a state variable for the previous errors.

Maybe other ways to do it - but the point would be to be able to see a list of errors encountered during featurize_dataframe with ignore_errors = True.

ardunn:
Sure, I think this would be tremendously useful! 

I would personally lean towards the second option (a `return_errors` arg). However instead of changing the return type, what do you think about adding a "(name of featurizer) errors " column to the dataframe? Then you can just print the relevant rows of the dataframe to see what rows did not featurize well (in an organized fashion).

computron:
Sounds good to me! But I would wait until the BaseFeaturizer changes are finished and merged before implementing this, otherwise we're gong to end up with merge conflicts all over.

computron:
Closed via https://github.com/hackingmaterials/matminer/pull/165

PR #165, opened 2018-02-22
Add return_errors option to BaseFeaturizer

ardunn:
- When used in `featurize_dataframe`, adds a column holding formatted strings of tracebacks
- When used in `featurize_many`, adds a 'feature' holding each formatted string traceback for each entry (e.g., `[nan, nan, nan, 'Traceback (most recent call last):\n ...']`)
- If no error is encountered, the error becomes `nan`

Changing the return type for `(features, errors)` for `featurize_many` is also an option, however it may be simpler for the time being to keep the return type the same.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -1,6 +1,6 @@
 from __future__ import division, unicode_literals
 
-import sys
+import sys, traceback
 import warnings
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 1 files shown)


ardunn:
Fixes #154.

computron:
@ardunn Note that you can put a message like 

Closes #154

in the PR and it will close the issue upon merging

computron:
nevermind looks like the "Fixes 154" message was there, surprised it didn't work

PR #168, opened 2018-02-23
Updated Bag of Bonds

ardunn:
- Uses `fit` instead of overriding featurize_dataframe. BoB now requires `fit` to be run before featurizing, and will throw an error if it is not. 
- Clear up docs + update to be in line with new BaseFeaturizer standards

- Added `no_oxi` arg, to optionally make BoB agnostic to oxidation states (ie, turn off discrimination between `Al3+ - Ni-` and `Al - Ni` .
- Added `token` arg, to optionally change the string representation of bonds in the dataframe (e.g., `Cs - Cl bond frac.` vs. `Cs Cl bond frac.` etc.)
- Updated BoB tests
- PEP8-ified BaseFeaturizer documentation


computron:
@ardunn why put allowed_bonds in fit() instead of init()

Note that I am trying to follow a convention of putting most arguments in init() unless really necessary to put elsewhere. That way someone sees most of the options up front and the fit() function is a bit more predictable in usage.

ardunn:
`allowed_bonds` is a list of bonds (as strings) which can be used as a fitting parameter (in lieu of a list of structures), *not* a parameter that should be set alongside bbv or approx_bonds (imho). I thought it would be most clear to put all available fitting parameters in the `fit` function. 

If allowed_bonds is in __init__() but you could still fit by passing a list of structures to fit(), it might be confusing for the user (e.g., "Wait, I thought I already fit BoB!")

That being said, in the tests I've ran with BoB, I've almost exclusively used a list of structures to fit, not a list of `allowed bonds`. So removing `allowed_bonds` altogether is an option. Maybe an enthusiastic user of BoB, (@albalu) could offer an opinion?

computron:
From skimming the code (not having a complete understanding) the way I see it is:

- if someone sets ``allowed_bonds`` in the constructor, there is no need to call fit() at all. They can just directly use featurize.
- Otherwise, if they are going to be calling fit, it should be with a list of structures only

computron:
(i.e., fit() is skipped if allowed_bonds was set)

albalu:
As a user if I have to call fit anyway, wouldn't make a big difference for me but for the sake of clarity it makes more sense to me to pass that when I instantiate the class (it can still be optional: i.e. one can set it either in fit or __init__ but if both are set raise error)

If calling the fit is not required I would definitely vote to have allowed_bonds in __init__

ardunn:
@computron You are correct. What I am saying is that using `allowed_bonds` is essentially just a way to fit. We could equivalently move the `structures` arg of fit() to init() and prevent calling fit() as well. But I thought the purpose of fit was to establish a protocol to use some featurizers. Ie "always call fit() before featurize().". If there are some scenarios where calling fit is not needed, then it seems we are breaking that protocol.

Anyway if the vote is for `allowed_bonds` in init it is no problem to move it there. Whatever is most clear for everyone is best

computron:
You can still call fit(), but fit() will just skip if allowed_bonds is set

computron:
the structures are *needed* to perform the fitting. the allowed bonds are not needed to perform the fitting

WardLT:
I also think `allowed_bonds` in the initializer makes sense, the list of structures does not. A pre-defined list of bonds to include as features is a permanent decision from a user, and the list of structures is something that would change when a training set is changed. 

PS. Does that help better differentiate what makes sense for `fit`? If so, I could update the BaseFeaturizer guidance about `fit`

ardunn:
`allowed_bonds` as an arg to `fit` defines `self.allowed_bonds_`
`structures` as an arg to `fit` also defines `self.allowed_bonds_`

Neither are necessary to perform fitting, but at least one is required. 

If you define `allowed_bonds` in the initializer, then `fit` won't change anything during training and doesn't really have a use if `allowed_bonds` is passed in init().

WardLT:
That's a good point @ardunn . And, would `fit` even overrwite what is provided in the constructor?

Would a solution like what we have for PartialRDF work? The argument for the constructor would take a list of bonds to include/exclude, and the `fit` operation actually sets `self.allowed_bonds_`.

This would require `fit` is called before running the featurizer. Maybe we could call `fit` with no structures from the initializer, if we want `fit` to be unnecessary in some cases.

ardunn:
@WardLT would PRDF-type solution require a variable number of arguments for fit? 

Scenario 1: I have a list of `allowed_bonds`
```python
bob = BagofBonds(allowed_bonds=['Li-O', 'Li-Li'])
bob.fit()
bob.featurize_dataframe(test_df, 'structure')
```


Scenario 2: I have a list of structures
```python
bob = BagofBonds()
bob.fit(train_df['structure'])
bob.featurize_dataframe(test_df, 'structure')
```

ardunn:
Another option is to just change the syntax to just have fit accept one argument, so that this is what the user would see:

Scenario 1:
```python
bob = BagofBonds()
bob.fit(['Li-O', 'Li-Li'])
bob.featurize_dataframe(test_df, 'structure')
```

Scenario 2:
```python
bob = BagofBonds()
bob.fit(train_df['structure'])
bob.featurize_dataframe(test_df, 'structure')
```

Scenario 3: Changing allowed bonds 
```python
bob = BagofBonds()
bob.fit(train_df['structures'])
bob.featurize_dataframe(test_df1, 'structure') # only shows bonds from train_df
bob.fit(['Li-Li', 'Li-O'])
bob.featurize_dataframe(test_df2, 'structure') # only shows Li-Li and Li-O bonds
```

computron:
Go back to scenario 1 and 2 from 2 two comments ago. In both scenarios you can call fit with a list of structures. But in acenario 1 the fit will not do anything since you already set allowed bonds. Or scenario 1 could instead take the intersection of allowed bonds set by the user and allowed bonds based on the structures

ardunn:
I like intersection however I think that adds some complexity to the usage. ie The expectation is that when a user 'fits' a Featurizer they should get back exactly the values they fit it for? Whereas with intersection the following situation could happen:


```python
bob = BagofBonds(allowed_bonds=['Li-Li', 'Li-O'])
bob.fit(train_df['structure'])         # if no columns in common between train_df and allowed_bonds
bob.featurize(test_df, 'structure')    # returns no new columns...
```

computron:
I don't see any problem with the example you put. Based on the constructor, the user wants to train the effect of Li-Li and Li-O bonds on a certain output (e.g. band gap). The training data set is telling you that you have no pertinent data on how Li-Li and Li-O might affect the band gap. So, calling featurize essentially does nothing because even though your new data set might contain such bonds, the training set is telling you that you're not going to get any good information from those bonds

ardunn:
Ok that is fine by me. Intersection it is, although it might be good to throw a warning in the case of the intersection having no bonds

computron:
Sure, a warning is OK by me

ardunn:
@computron The behavior is now:

- The allowed bonds can be set through `__init__` with `allowed_bonds`
- The allowed bonds can also be set through `fit` with a list of structures
- If `allowed_bonds` is set **and** `fit` is called, the allowed bonds are set to the intersection of the allowed bonds and the bonds in the list of structures. A BoB object can be refit multiple times, and uses the intersection of the allowed bonds passed into the initializer and the *current* list of structures to determine which bonds to output. So you can refit as much as you want and always make sure you are only featurizing according to allowed bonds.

matminer/featurizers/structure.py
```diff
@@ -1172,34 +1209,44 @@ def from_preset(preset):
...
+                objects which will be used to determine the allowed bond
+                types.
 
         Returns:
-            (DataFrame) BagofBonds-featurized dataframe
+            None
         """
+        listlike = (tuple, list, np.ndarray, pd.Series)
+        if not isinstance(structures, listlike):
+            structures = [structures]
```
computron: don't fix it for the user. Otherwise, the user might try to send a single Structure into fit() for this featurizer, have it work out, and think that's an OK usage of the code. Just throw an error if they give the wrong argument. Otherwise *every* featurizer should fix Structure -> [Structures], and I don't think we should do that.

computron: I will just merge and change that, it's probably the easiest

PR #171, opened 2018-02-26
Added example that uses the MDF

WardLT:
Introduces a new example. Showcases training a machine learning model that predicts formation enthalpy given composition, which uses OQMD data taken from the MDF search index.

We discuss this example in the paper, should we include it in the repo?


computron:
I think it's good to have the example but not sure where to put it.

- main matminer repo is good for discoverability/maintainability, but unfortunately it really bloats the repo. Every change to Jupyter notebook is a huge diff, and the files are also just very large. We tried to clear most Jupyter notebooks from the repo for this reason (have some more to go still I think).
- separate gists prevents bloating the repo, but is bad for discoverability/maintainability
- the "git-lfs" seemed to be a good middle ground - lets you put it in the main repo without having it bloat it. But when Kyle and I tried it, we found it was still a little hairy. In frustration I ended up trying to remove all trace of git-lfs. But if someone can really get it working cleanly it might be a good option.
- other ideas?

WardLT:
Have you considered having a separate git repository (e.g., "matminer_examples" or "matminer_notebooks") to store a collection of notebooks? Having them under a git repo would make them maintainable, and it is probably possible to automate testing them with new changes to matminer. It would be slightly less discoverable and still have huge diffs, but at least it would be away from the matminer source code.

computron:
Let's put all examples in a separate repo as you suggest. I just created one:

https://github.com/hackingmaterials/matminer_examples

Please submit the PR there and I will merge. Feel free to create any directory structure or other files you think are needed in that repo, I just put something bare-bones for now.

PR #172, opened 2018-02-26
FLLA Dataset and SCM/OFM example script

kylebystrom:


kylebystrom:
I fixed a couple of Codacy style issues, I think the two remaining style issues that got marked aren't real issues in context, but I'll change them if it's preferable.

PR #173, opened 2018-02-27
Adjusted ChemicalSRO with fit

Qi-max:
Adjusted ChemicalSRO to work with `fit`:

- use `fit` to identify elements to be included in the following featurization (`self.el_list_`), by intersecting the elements present in the structures with those explicitly included (or excluded) by `self.includes` and `self.excludes` in `__init__`. Compositions of the passed structures will also be stored in a dict `self.el_amt_dict_`, avoiding repeated calculation of compositions when featurizing multiple sites in the same structure.

- use `_check_is_fitted` (as named in sklearn) to verify if `fit` is called before `featurize` or `featurize_labels`, and raises a `RuntimeError` if not.

Concerning this issue, shall we move `_check_is_fitted` to utils and maybe create a custom exception like the `NotFittedError` in sklearn? Besides, `_check_is_fitted` is now called in `featurize` or `featurize_labels`, would it be better to call it in `featurize_dataframe` or `transform` to avoid repeated checking? Or as the user might call `featurize` directly, so it still has to be called in `featurize`?

Other minor changes:

- add `**kwargs` to from_preset to allow passing args to the NearNeighbor class.

- update comments, etc.


Qi-max:
Codacy problem: 
`def fit(self, structs)` parameters differ from overridden `fit` method.
Should I change this? Thanks~

computron:
You can ignore the codacy

Your "check_is_fitted" function is way too complicated. It is trying to be very general (any message, any parameter check, etc) for no reason. Plain and direct code is better than abstract code unless there is really a good reason to make it abstract and general. Currently, you are always calling it with the same parameters and message, and there is really no reason for those to be settable. So you should change check_is_fitted to be very simple, maybe one line of code that explicitly checks the features you want.

Qi-max:
Yeah, the current `check_is_fitted` is coded more like a general version (similar to that in sklearn). I thought this function might be better to be migrated to utils or elsewhere, so allowing to change checked attributes, error message etc can be useful.

computron:
So why don't people just use the sklearn version? I see no reason to reimplement it yourself within a single estimator (this makes no sense at all, since other estimators can't easily call it) or even putting it in matminer utils (the same function already exists in sklearn, which is a requirement of matminer!)

Honestly I'd just put a simple "if not <parameter exists>, throw error". Or if you don't want to do that, call the sklearn function

ardunn:
Maybe it would be good to eventually write a parent class for these structure/site heterogeneous featurizers like Max was doing? For example BoB also requires a _check_fitted method.

WardLT:
Regarding the Codacy error about `fit`: I think sklearn's Pipeline module needs the `y` argument in `fit` to run properly. Also, the `fit` operation should take the same inputs as `featurize` so that you do not pass different data for `fit` than `transform`. Would it make sense to revise the function call for `fit` to be `def fit(self, X, y=None):` where `X` is a list of structures and sites?

computron:
@WardLT I agree with the revision for fit

@Qi-max @ardunn You should just call the sklearn function if you really don't want to use a direct if-else. Most coding sins stem from people trying to develop fancy solutions to simple problems.

ardunn:
@computron roger that. BoB's _check_fitted is just a direct 2-liner if else currently

@WardLT Yes! That would fix BoB codacy sins as well

Qi-max:
@computron @WardLT  Ok, I see. I will make the change soon. Thanks!

ardunn:
@WardLT "Also, the fit operation should take the same inputs as featurize so that you do not pass different data for fit than transform" Do you mean `featurize_many`?

WardLT:
@ardunn Yeah, that's right: same as featurize_many

Qi-max:
In fixing `fit`, it seems that the current `transform` (that directly calls `featurize_many`) does not support dataframe as input `X`. So if `fit` takes the same input as `transform` and does not support dataframe, the user need to use `.values ` to transform the dataframe first. Will it be a problem? Is it possible that we also support dataframe as input in `fit` and `transform` and transfer dataframe to list within? Then the user can easily put dataframe as input in `fit` as in `featurize_dataframe`. Thanks! @computron @WardLT @ardunn

computron:
``fit`` is not supposed to take a dataframe as an input. It is supposed to take a list of data, e.g. a list of structures as ``X``. Only ``featurize_dataframe`` takes in a dataframe as input AFAIK

Qi-max:
I don't know if I made myself clear... For instance, is it possble to support (at least tolerates) this: 
`.fit(df[['struct', 'site']])`
I think I used to feed a dataframe to `fit` and `transform` in sklearn, and it doesn't raise error. Maybe this will make it more flexible? Thanks for all the clarifications.

WardLT:
Thanks for pointing this out Qi. It is a little confusing to have some
operations  be dataframes friendly and others require conversation to numpy
array, especially when both pandas/numpy hold array data.

We could build something simple to handle converting pandas arrays into
transform. Maybe this is better a different issue/PR?

On Feb 27, 2018 9:15 PM, "Qi Wang" <notifications@github.com> wrote:

I don't know if I made myself clear... For instance, is it possble to
support (at least tolerates) this:
.fit(df[['struct', 'site']])
I think I used to feed a dataframe to fit and transform in sklearn, and it
doesn't raise error. Maybe this will make it more flexible? Thanks for all
the clarifications.

—

<https://github.com/hackingmaterials/matminer/pull/173#issuecomment-369097340>,
or mute the thread

Qi-max:
@WardLT Hi Logan, sorry for the late reply! Yeah, I think making `transform` dataframe friendly would be a minor modification, such as using `np.atleast_2d(X) ` before calling `featurize_many`. 

I have created a new commit that already made fit in ChemicalSRO compatible with dataframe, is it ok at this moment, to finish this PR? A few additional tests on `fit` + `transform` and `fit_transform` are also included in unittest. 

Btw, the Codacy issue is still there due to no` **fit_kwargs` in `def fit(self, X, y=None)`. So just ignore it? 

Thanks!

WardLT:
Sounds great!

I would ignore the warning about the fit_kwargs, I don't think we need them for compatibility reasons.

matminer/featurizers/site.py
```diff
@@ -804,15 +801,19 @@ def fit(self, structs):
         in a dict of self.el_amt_dict_, avoiding repeated calculation of
         composition when featurizing multiple sites in the same structure.
         Args:
-            structs (pandas series): series of pymatgen Structures.
+            X (pandas series): series of pymatgen Structures.
```
computron: this should just be a list-like object of pymatgen structures (not necessarily a pandas series)

matminer/featurizers/site.py
```diff
@@ -804,15 +801,19 @@ def fit(self, structs):
...
+            y : to be compatible with the overridden fit method.
```
computron: for docstring of y, put:

y : unused (added for consistency with overridden method signature)

matminer/featurizers/site.py
computron: also is this correct? do you need a 2D array of structures and sites? if the latter the docstring should read:

```
X (2D array): 2D array of pymatgen Structures and Sites. One can also pass two dataframe columns, e.g. df["structures", "sites"]
```

Not sure if what I wrote above is correct, but please double-check your docs

computron: Also it's unclear to me what the format of X should be if one is not using dataframes - the tests don't have any examples. Can you clarify the non-dataframe format of X?

computron:
See just my minor comments on documentation on:

https://github.com/hackingmaterials/matminer/pull/173/commits/814b4d046d7b92fb4b56b159a73bf00fe0615ac9

computron:
Actually it looks like the comments show up as part of this PR above

matminer/featurizers/site.py
Qi-max: Forgot to cleanup the docs... I will fix this soon. 

The input X now has multiple choices: 
1. X can be 2D list-like object: `[['structure', 'site'], ['structure', 'site'], …]` for instance
2. X can be dataframe:  `df[['structure', 'site']]`
3. X can be tolerant to 1D object: `[structure, site]`.

They will all be transformed to `np.array` within `np.atleast2D()`. Is this behavior desirable?

The test of `df[['structure', 'site']].values` is embedded in the test of `fit_transform` currently. I will update the tests and also add test of a 1D object later.

Qi-max:
I think they are now ready for review. @computron

computron:
Thanks! Note that in the future, you shouldn't worry so much about "fixing" user inputs - e.g., converting string to array of string. Usually it's better to just be clear about what is expected and enforce a consistent usage of the code rather than trying to fix bad input from the user. For example, just say that you expect a list of string and don't worry about converting on behalf of the user.

I will merge and just make the minor change to not return self at the end of fit (I can't think of any reason why this would be needed, so let me know if you have some specific reason for it).

computron:
Nevermind my last comment, I didn't realize that fit() needs to return self for the fit_transform() to work due to method chaining

PR #175, opened 2018-02-27
FLLA Dataset (No Kernel Ridge Example)

kylebystrom:
This is the same PR as last time except I removed the KRR example script, i.e. the one change I made is to add the dataset used by Faber et al in their 2015 paper.


computron:
Thanks!

PR #176, opened 2018-02-28
PlotlyFig updates

ardunn:
# Summary

- Complete a full set of examples as .py files: 1-4 examples each for [xy, scatter_matrix, bar, histogram, violin, parallel_coordinates, heatmap] and 2 examples showing the plot modes and advanced formatting
- Tests that won't break every time a default is changed
- Various bugfixes and improvements

I am guessing that the examples will be eventually ported to matminer_examples in notebook form, but I think they are still useful here for the time being


computron:
Can you just put all examples in matminer_examples? Otherwise there is going to be schism of some examples in matminer vs other examples in matminer_examples and people are going to get confused.

computron:
Thanks!

PR #180, opened 2018-03-02
Update njobs to cpu_count()

### LINKED ISSUE #162: Discussion: default n_jobs for Featurizers
Currently the default n_jobs for featurizers is 1, i.e., serial jobs.:

matminer/featurizers/base.py:105

This choice is in-line with what sklearn featurizers do, e.g.:

http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html

I'd suggest being more aggressive and setting this equal to cpu_count() - i.e., to use full parallelization by default. For example, the multiprocessing module uses cpu_count when the number of processes is set as ``None`` (default is full parallelization).

Regardless of the default, I can see use cases for both and it would be nice to have a "set-it-once-and-forget-it" way to configure n_jobs. e.g., a config variable one can set at the beginning of a Jupyter notebook.

ardunn:
This breaks 2 site tests, AFS and GRDF @dyllamt . This is not the fault of BaseFeaturizer, but the fact that the tests use default parallelization (requires pickle) and AFS/GRDF use lambda functions. AFS/GRDF do not work with njobs>1 (unless I am doing something wrong), because their lambda functions are not pickleable.

Before, when the default was njobs of 1, the issue was not apparent, but with default njobs = cpu_count, the issue is apparent (ie the site tests use default njobs)

Closes #162

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -5,7 +5,7 @@ import warnings
 import pandas as pd
 import numpy as np
 from six import string_types
-from multiprocessing import Pool
+from multiprocessing import Pool, cpu_count
 from sklearn.base import TransformerMixin, BaseEstimator
 
 
@@ -113,7 +113,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
     @property
     def n_jobs(self):
-        return self._n_jobs if hasattr(self, '_n_jobs') else 1
+        return self._n_jobs if hasattr(self, '_n_jobs') else cpu_count()
 
     def fit(self, X, y=None, **fit_kwargs):
         """Update the parameters of this featurizer based on available data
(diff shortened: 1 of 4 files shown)


computron:
@ardunn I would suggest

1. Modifying the constructor of those two featurizers so that if the njobs is set >1, to call set_njobs(1) and use warnings.warn("") to print a message saying that the featurizer doesn't support n_jobs > 1. After completing, the tests will pass
2. Creating an issue about n_jobs > 1 not working for those 2 featurizers and tagging @dyllamt 

This strategy allows me to pull the change and the updated cpu_count(), retains the previous behavior, and creates an issue for the new problem

computron:
I guess you don't need the warning in the constructor, maybe just fit() and/or featurize() then? i.e., the place where it would actually be used

computron:
nevermind what you have is fine, I misread the diff based on the way github formatted it

PR #190, opened 2018-03-08
AFS Presets

dyllamt:
Added presets for AFS. User can choose between two common bin functions:
1. Gaussians
2. rectangular functions (histogram mode).

The user can also control the bin widths, bin spacing, and cutoff radius easily with the `from_preset` function.

New tests for AFS ensure that `featurize` can be called with the preset bin functions without errors and that the correct set of bins are generated.


dyllamt:
Added same functionality for `GeneralizedRadialDistributionFunction` featurizer

PR #191, opened 2018-03-12
added u_o example back to docs

montoyjh:
Just updates the examples list in docs_rst for the u-o bondlengths example.

Out of curiosity, should we be updating the html docs using invoke when we issue PRs like this?  I'm not sure of the protocol.


computron:
Thanks!

You can either update or not update the HTML docs. Probably prefer updating the HTML.

- don't update - less likely to have strange merge conflicts if someone else also working on it. If you use this method, just include a note in the PR requesting that I update the HTML after merging.
- update - OK as well, so long as there are not merge conflicts. Then there is no second step.

computron:
(note that if you update the docs, use the "make_doc" task and not "update_doc". The latter seems to be tied to releasing a new version of matminer)

Issue #91, opened 2017-12-15
Voronoi-tessellation-based features (i.e., de Jong, Isayev, Ward) 


computron:
talk to @WardLT if interested

https://www.nature.com/articles/srep34256
https://www.nature.com/articles/ncomms15679
https://journals.aps.org/prb/abstract/10.1103/PhysRevB.96.024104

computron:
(maybe something for @Qi-max to contribute?)

WardLT:
I'm gradually working on this issue. If anyone else is working on it or wants to, let me know and we can figure out how to combine efforts.

WardLT:
Thanks, @Qi-max! I'm about halfway done with the features from the PRB article, which will cover most of the Sci Rep features. So, it is probably easiest for me to finish those two papers myself.

Would you want to start on the Nat. Comm features?

Also, the areas that I haven't figured out yet:

1. Getting 2nd-nearest neighbors from the NearNeighbor pymatgen classes, which are needed for some of the PRB attributes. I was thinking of building an additional method on to the NearNeighbor classes for this, but am not sure if that is already done or if there is an easier way.
2. Whether we can/should combine these features with your existing VoronoiFingerprint class. I am in favor of keeping them as separate featurizers to avoid having classes with too many options. 
3. How to cache the Voronoi tessellation, if it is used by multiple featurizers. When implementing the PRB features in Magpie, I found that running the tessellation comprised a significant amount of the featurization time. While I'm not doing anything about caching now, I'm worried that making many different featurizers (see 2) will exacerbate any performance issues.

Any thoughts about these issues?

Qi-max:
Hi Logan,

Thank for your reply! 

for 1, I think adding an additional method to the NearNeighbor classes to get 2nd-nn is a good way to go. I am only a bit concerned about the performance issue, as it may includes some repeated calculation of 1st-nn in real applications. eg., if we want to get 2nd-nn of each atom in the structure and we already calculated the 1st-nn of all atoms (which is a more common case I think), we can just combine the 1st-nn of the 1st-nn atoms of each atom to get the 2nd-nn atoms, without further calculation. This requires solving the caching issues. But still, I think adding a standard way to get 2nd-nn is quite useful.

[...]

PR #195, opened 2018-03-27
Adding Features from Ward PRB 2017

WardLT:
This PR adds features from https://journals.aps.org/prb/abstract/10.1103/PhysRevB.96.024104

Partially addresses issue #91 

This will be ready to merge once the next pymatgen is released, but it is ready for review for the time being.


matminer/featurizers/site.py
```diff
@@ -1677,3 +1708,109 @@ def citations(self):
...
+                data_source=MagpieData(),
+                properties=["Number", "MendeleevNumber", "AtomicWeight",
+                            "MeltingT", "Column", "Row", "CovalentRadius",
+                            "Electronegativity", "NsValence", "NpValence",
+                            "NdValence", "NfValence", "NValance", "NsUnfilled",
+                            "NpUnfilled", "NdUnfilled", "NfUnfilled",
+                            "NUnfilled", "GSvolume_pa", "GSbandgap",
+                            "GSmagmom", "SpaceGroupNumber"]
+            )
+        else:
```
computron: technically you don't need the "else" here, if you haven't "returned" by now it's time for an error

Hard to say what's the more "proper" code here. Leaving the else "in" might be clearer than removing it? up to you

matminer/featurizers/structure.py
```diff
@@ -952,24 +956,30 @@ def featurize(self, s):
...
-                ' parameter feature vectors}, journal={in progress}, author={'
-                'Zimmermann, N. E. R. and Jain, A.}, year={2017}}']
+        return self.site_featurizer.citations()
 
     def implementors(self):
-        return ['Nils E. R. Zimmermann', 'Alireza Faghaninia', 'Anubhav Jain']
+        return ['Nils E. R. Zimmermann', 'Alireza Faghaninia',
+                'Anubhav Jain', 'Logan Ward']
 
+    # TODO: Should we make these classmethods? They call should return an initialized version of the class -LW
```
computron: I don't have a strong opinion, in my experience:

1. the class method adds the complication of remembering to put "cls" as the first argument
2. the above is useful if you need the class information for something (e.g., you write a subclass that inherits from this class, and want the same function to work for both the superclass and subclass but initializing a class of the same type).

I don't really see a case for point (2) above here, let me know if there are other reasons to prefer a class method

matminer/featurizers/structure.py
```diff
@@ -1550,3 +1574,297 @@ def citations(self):
...
+                "in machine learning models of formation energies "
+                "via Voronoi tessellations}},"
+                "url = {http://link.aps.org/doi/10.1103/PhysRevB.96.014107},"
+                "volume = {96},year = {2017}}"]
+
+    def implementors(self):
+        return ['Logan Ward']
+
+
+class ChemicalOrdering(BaseFeaturizer):
```
computron: Can you clarify (for my knowledge) the difference between this and the ChemicalSRO site descriptor?

matminer/featurizers/structure.py
```diff
@@ -1550,3 +1574,297 @@ def citations(self):
...
+                "in machine learning models of formation energies "
+                "via Voronoi tessellations}},"
+                "url = {http://link.aps.org/doi/10.1103/PhysRevB.96.014107},"
+                "volume = {96},year = {2017}}"]
+
+    def implementors(self):
+        return ['Logan Ward']
+
+
+class StructureComposition(BaseFeaturizer):
```
computron: I think this should be deleted - the user can just create a composition column rather than pass through this wrapper object.

See for example the ``structure_to_composition`` function in ``conversions.py`` which allows for:

``df["composition"] = structure_to_composition(df["structure"])``

It is a *little* repetitive in the data since the composition is embedded in the structure, but I don't think it's so bad.

computron: Maybe Im being too hasty in suggesting deletion - there is some discoverability value in having this here. Lets discuss more

matminer/featurizers/site.py
WardLT: I could go either way. I'll keep the 'else' in, because - to me - it makes it clearer that this is a case/select block.

matminer/featurizers/structure.py
WardLT: Thanks for your thoughts on this. I agree, case (2) is the main advantage of a class method and we haven't yet seen a need for that.

I'll remove this Todo

matminer/featurizers/structure.py
WardLT: It appears that ChemicalSRO breaks down the degree of ordering to each particular element. ChemicalOrder is - in some ways - a coarse average over all the ChemicalSROs for each type of element in the structure.

Does that help clarify? 

I'm currently polishing the documentation of each class to make the difference clearer.

WardLT:
Thanks! 

No, it won't pass with the new version of pymatgen - yet. The changes to pymatgen actually break the OFM features (I changed the output of a method it uses). I'll fix that too.

matminer/featurizers/structure.py
WardLT: I was concerned about this featurizer being a redundant with `structure_to_composition`, so I'm glad you brought this up.

My rationale for adding it was to allow for training a structure-based model with a single `MultiFeaturizer`. This way, I can write scripts that compare different structure-based featurizers easier - I do not need to have extra code for featurizers that use composition-based information and those which do not. 

The other advantage is that this allows for running featurizer on single entries (or non-pandas lists of entries) much more easily. Without this adapter, I might end up with code like this:

```python
x = Structure(...)
features = np.hstack(strc_f.featurize(s), comp_f.featurize(x.composition))
```

Which I think is tedious to write.

Do these use cases seem reasonable enough to warrant this wrapper class?
```

computron: Ok let's keep it

matminer/featurizers/site.py
computron: ok

matminer/featurizers/structure.py
computron: ok

computron:
This is OK to merge from my side once the unit tests pass

PR #197, opened 2018-04-05
add DopingFermi: a Dos featurizer + unittest

albalu:
this featurizer returns the fermi levels that result in the input doping levels (at a given temperature). This fermi is calculated solely based on pymatgen DosFermi object or any pmg Dos object that has the structure attribute.

the tests are expected to fail until a new version of pymatgen is released. I did not try  @unittest.expectedFailure decorator as I am importing DopingFermi at the beginning of the test and DopingFermi class is not available in the latest pymatgen so the test would fail even with the decorator.

### DIFF
--- a/matminer/featurizers/dos.py
+++ b/matminer/featurizers/dos.py
@@ -1,15 +1,15 @@
 import numpy as np
 from collections import OrderedDict
 from matminer.featurizers.base import BaseFeaturizer
+from matminer.featurizers.composition import BandCenter
 from pymatgen import Spin
-from pymatgen.electronic_structure.dos import CompleteDos
+from pymatgen.electronic_structure.dos import CompleteDos, FermiDos
 
 
 class DOSFeaturizer(BaseFeaturizer):
     """
     Featurizes a pymatgen density of states, CompleteDos, object.
     """
-
     def __init__(self, contributors=1, significance_threshold=0.1,
                  energy_cutoff=0.5, sampling_resolution=100, gaussian_smear=0.1):
         """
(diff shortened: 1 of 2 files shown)


computron:
I am confused about the unit test

>> DopingFermi class is not available in the latest pymatgen

Do you mean FermiDos? DopingFermi is in matminer

albalu:
oh yes sorry I meant FermiDos from pymatgen. I don't understand the Codacy issue...

WardLT:
FYI: I just merged a commit that updates pymtagen to the latest version.

WardLT:
Also, the Codacy issue is a result of `featurize` FermiDos having different arguments than just `*x`. It is our standard practice in matminer to overload `featurize` to have different arguments for different types of featurizer, so feel free to ignore the Codacy error.

A few comments/questions about your featurizer:

- Can you make your docstring for the class explain what these features are? I'm trying to create a common standard for how to read the docstrings, which is explained in BaseFeaturizer (also see [ChemicalOrdering](https://github.com/hackingmaterials/matminer/blob/4ceaa3aa08a60bde8f7d9df07d293b7ce8d86ff3/matminer/featurizers/structure.py#L1746))
  - While you're at it, could you clarify the difference between DopingFermi and DOSFeaturizer in their docstrings?
- Why would a user to pass the structure or bandgap to featurize? Could you given an example in the unit tests showing/testing their effect?

albalu:
How does it look now?

matminer/featurizers/dos.py
```diff
@@ -101,6 +101,97 @@ def implementors(self):
         return ['Maxwell Dylla', 'Alireza Faghaninia', 'Anubhav Jain']
 
 
+class DopingFermi(BaseFeaturizer):
+    """
+    Futurizes a pymatgen density of states (dos) object. structure is needed;
+    it can either be fetched automatically from dos.structure(e.g. CompleteDos)
+    or passed into featurize/featurize_dataframe. This featurizers return the
```
WardLT: Could you revise this doc string to make the 3rd sentence ("This featurizer returns...") the first? That way, the featurizer summary table Alex made (#200) will display more meaningful information.

matminer/featurizers/dos.py
```diff
@@ -101,6 +101,97 @@ def implementors(self):
...
+            if not self.return_eref:
+                fermi -= eref
+            feats.append(fermi)
+        if self.return_eref:
+            feats.append(eref)
+        return feats
+
+    def feature_labels(self):
+        """
+        Returns ([str]): list of names of the features generated by featurize
```
WardLT: Could you move this description of the features to the main class documentation? This change would make it more consistent with the other Featurizers.

WardLT:
Outside of the two comments about the docstrings, this looks ready to merge to me.

matminer/featurizers/dos.py
ardunn: @albalu Preferably just put the first line of the docstring as:

Fermi level associated with input carrier concentration and temperature w.r.t.  reference energy

ardunn:
@albalu @WardLT let's merge this before we merge #200, so #200 includes FermiDos in the generated  docs.

albalu:
Ok I take care of these now.

albalu:
ok nevermind seems like @computron is fixing them. Thanks.

computron:
You also shouldn't require structure

1. It is only needed if the user wants to adjust the midgap energy based on electronegativity. Don't force this requirement on people that don't want to make that adjustment
2. My preference is probably just to keep it simple. If you use midgap adjustment, the assumption is that DOS object has a structure attribute. Otherwise, you need to do the midgap adjustment manually (which is not hard)
3. If you *do* keep an optional parameter for midgap adjustment, all you need is composition, not structure. But make it clear that this is *only needed for midgap adjustment*. It is confusing why one needs a structure in order to get a Fermi level for carrier concentration

computron:
@albalu I think you should handle the remaining ones, I did the first one @WardLT mentioned

albalu:
@computron the structure is also needed for FermiDos class. It is used there for getting the volume and total number of electrons both of which is required to get a concentration units is 1/cm3.

To write FermiDos in pymatgen I had a choice of requiring Composition and volume separately or just Structure and I decided having structure is more straight forward and more likely given the volume already has to be known

computron:
Ok ignore my comment about structure, I misunderstood how it was being used

computron:
@albalu please use the "ready to merge" tag when finished making changes

albalu:
@WardLT I kept the labels documentation as I think it needs some documentations. featurize, return a list of numbers for which I provided a better doc

albalu:
Seems like I have to be a "contributor" to label issues and PRs? https://stackoverflow.com/questions/13829466/how-to-put-a-label-on-an-issue-in-github-if-you-are-not-a-contributor-owner

Anyway, I do old fashioned for this one: it is ready to merge

PR #198, opened 2018-04-08
Caching Nearest Neighbor Results

WardLT:
This PR introduces a caching mechanism for computing nearest neighbors. In tests for the Ward PRB features (see #91), caching speeds up featurization by about a factor of 2-3x. 

There are already a few things that I'd like to get someone's opinion about:

1. Is automatically caching neighbors for all sites a good idea? I currently do this to create a single cache for site and structure featurizers, and to minimize the size of the cache. 
2. Should we make caching optional? Does anyone have a use case (preferably for site featurizers) they want me to benchmark? 
3. Anything else worth caching?

### DIFF
--- a/matminer/utils/tests/test_caching.py
+++ b/matminer/utils/tests/test_caching.py
@@ -41,9 +41,10 @@ class TestCaching(PymatgenTest):
         self.assertEquals(2, _get_all_nearest_neighbors.cache_info().misses)
         self.assertEquals(2, _get_all_nearest_neighbors.cache_info().hits)
 
-        # Perturb the structure, make sure it induces a miss
+        # Perturb the structure, make sure it induces a miss and
+        #  a change in the NN weights
         x.perturb(0.1)
-        get_nearest_neighbors(method, x, 0)
+        nn_2 = get_nearest_neighbors(method, x, 0)
+        self.assertNotAlmostEqual(nn_1[0]['weight'], nn_2[0]['weight'])
         self.assertEquals(3, _get_all_nearest_neighbors.cache_info().misses)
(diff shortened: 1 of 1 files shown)


computron:
Btw do caching features work if you change a structure?

e.g

- load a structure
- featurize it
- change the same structure object (perhaps by calling perturb() on it, deleting a site, or something else)
- featurize it again

That second featurization should give different values than than the first. From my inspection of the code it seems like it would work, but just want to make sure.

Even if it works, it might be a good unit test to have to make sure this doesn't break in the future. Also, maybe we want to encourage having IStructure objects to begin with (e.g., in utils/conversions.py have a method to convert all Structures to IStructure). Then, in the code that converts to Structures to IStructure, you want to test if something is already an IStructure instance (or contains a hash method) and avoid conversion to IStructure if that's the case to save the time of the structure conversion to IStructure, assuming that time to convert a large amount of structures is non-negligible.

computron:
@WardLT any thoughts regarding if caching a mutable object will work correctly?

WardLT:
I think it would cause problems if we cache mutable objects (e.g., Structure), but I’m not sure. I’ll implement those tests you suggest and find out. 

Sorry for not having addressed these comments yet. I’m hoping to get some more time to work on matminer this weekend or early next week.

WardLT:
Sorry about rebasing. I did not realize it would render the PR unreadable. Should I try to fix this?

WardLT:
I just added in a utility operation for converting a Series of Structures to IStructures. Over 100 iterations, it does have a minor, but noticeable 5% improvement for the featurization time (~80s to 76s on a test case I've been using).

As requested, I've also added tests to ensure perturbing a Structure will lead to different results even with caching and ensure caching is working as expected.

PR #199, opened 2018-04-11
Add a proper BagofBonds featurizer

ardunn:
## Summary
- Rename old BoB to BondFractions
- Add a BagofBonds featurizer based on [original paper](https://pubs.acs.org/doi/pdf/10.1021/acs.jpclett.5b00831), which can accept CoulombMatrix or SineCoulombMatrix (or similar) to generate the CM values
- Assorted fixes to CoulombMatrix
- A set of tests for new BoB

## Notes

A featurizer like this might benefit from a 'fit_featurize' method or something similar, as the act of fitting _requires_ featurization for all entries (unlike, for example BondFractions or PRDF).

Also it is up for grabs whether the default Coulomb matrix should be CoulombMatrix or SineCoulombMatrix


matminer/featurizers/structure.py
```diff
@@ -1590,6 +1594,182 @@ def citations(self):
...
+            for j in range(nsites):
+                bond = bonds[i, j]
+                cmval = cm[i, j]
+                bags[bond].append(cmval)
+
+        # We must sort the magnitude of bonds in each bag
+        return {bond: sorted(bags[bond]) for bond in bags}
+
+
+    def featurize(self, s):
```
computron: Can you add a docstring here like you did for bag()?

Some of the docstrings in matminer are inconsistent but we'll at least try to be more consistent for new featurizers and fix the old ones later

computron:
I added a minor comment about docstring

Is the description of these features in the matminer paper up to date? If not, can you also work w/Logan to update the matminer draft so we don't ship a paper w/incorrect info?

Note that new featurizers (like the DosFermi Alireza is working on) will come in during the paper process. We might want to update all of them in one go during the review. For now, I just want to make sure that existing information (e.g. on BagofBonds) is not incorrect.

WardLT:
The description of the features in matminer is not update, but I can do that while editing the paper today/tomorrow.

Also, who made the figure listing all of the featurizers? I'd like to update that as well, so could someone send me the code for it?

matminer/featurizers/structure.py
ardunn: For featurize? Sure.

computron:
I think @nisse3000 did the figure and no one else has the plot

Unfortunately he is away from work until the end of the month. 

I think it is OK if we update the plot during the revision rather than initial submission. Probably we'll need to revise the plot at that point anyway since the number of featurizers will likely have increased from initial submission to revision time

matminer/featurizers/structure.py
computron: yes for featurize

computron:
@ardunn - note that I added a new label called "ready to merge!" Please use it when you think everything is good to go

ardunn:
@computron "Note that new featurizers (like the DosFermi Alireza is working on) will come in during the paper process. We might want to update all of them in one go during the review. For now, I just want to make sure that existing information (e.g. on BagofBonds) is not incorrect."

Ok sounds good to me. Also whenever a release is made (via tasks.py --> release) the table will auto update

computron:
Yah I meant the table in the *paper*, not the table in the online docs (the table in the paper doesn't auto-update!)

Maybe @WardLT can mention in the paper that an updated table of featurizers is available in the online docs. So that when the paper does go out of date, people know where to go for updated info

ardunn:
Whoops, yep understood. Probably it would be good though if @WardLT can correct the BagofBonds entry (for correctness, because people might see "BagofBonds" thinking it is implemented like the paper, then look at the description which says something about bond fractions, and be confused...) and add the newer featurizers (e.g. StructuralHeterogeneity)

WardLT:
I’m actually working on updating the table of featurizers in the paper right now. 😊

I was reading over your implementation of BagOfBonds to understand how this featurizer works. To address your question about “fit_featurize,” could we implement the “fit” function in a way that avoids calling the CoulombMatrix featurizer? We don’t actually need the bond strengths during fitting, so couldn’t we just determine the necessary bag lengths from the composition of the structure?

ardunn:
@WardLT that is a good point, with one correction: we need more than the composition to determine the maximum bag lengths, we need the structure. We basically need to 

(1) calculate the number of each kind of bond in each _structure_ (the "bag length")
(2) for each bond, select the largest bag length from among all fitted structures

So I suppose that could be a todo for BoB: determine the bag lengths without doing the CM computation.

However I also think just due to the nature of BoB, something in the spirit of 'fit_featurize' might be nice, because (a) the *number of features is variable* with the set of structures to be featurized but (b)   'fitting' on one set of structures and 'featurizing' on another set is probably going to screw up any BoB featurizer implementation. You'll probably always want to fit and featurize on the _same dataframe_ in one operation. This is because, unlike BondFractions, where if a bond type was missing from the training set you could just ignore it or approximate it, with BoB there are also problems if any structure in the test set has a larger bag than the corresponding bag in the training set. 

Basically what I am saying is that in some circumstances, if you fit on one dataset and featurize a different one, you run into trouble whether you decide to correct it or not. If you don't correct it, you have errors often. If you do correct it, you will either have an incomplete BoB vector or you will have to refit BoB!

WardLT:
I think fit_featurize is a good idea, and BoB is a good motivating case.
Since we already have fit_transform via the TransformerMixin, would we just
need to add a fit_featurize_dataframe?

ardunn:
Would fit_featurize_many work through fit_transform?

WardLT:
Yea, I think fit_featurize_many would be the same as fit_transform. Maybe
we create both? I'd vote for only fit_transform

ardunn:
I would vote for only fit_transform as well. But I have yet to test if it works as is in leiu of fit_featurize_many...

ardunn:
Ok so fit_transform does work in place of fit_featurize_many (neat!). For fit_featurize_dataframe, something like:

If we were to put in BaseFeaturizer
```python
def fit_featurize_dataframe(df, col, *args, **kwargs):
    return self.fit(df['col']).featurize_dataframe(df['col'], *args, **kwargs)
```

Could work

Issue #193, opened 2018-03-23
Add summary of featurizers to documentation


WardLT:
The draft of our paper contains a list of all of the featurizers as a nice summary table. We should add something similar to our documentation, and keep it up to date.

computron:
I think this would be useful, but I also think these things can be difficult to maintain and keep up to date. It takes a lot of discipline and remembering to do it which people don't usually have - i.e., to keep the list of features, their descriptions (e.g. if we add more features, etc), and potentially even citations up to date).

 It might be easier and more sustainable to just:

1. link to the various featurizer base module doc pages, e.g.:
https://hackingmaterials.github.io/matminer/matminer.featurizers.html#module-matminer.featurizers.composition
or 
https://hackingmaterials.github.io/matminer/matminer.featurizers.html#module-matminer.featurizers.structure

2. Pretty-ify those docs above (somehow) and ensure that all docstrings are actually there (e.g., I just noticed DensityFeatures is not there).

computron:
(this would also help make sure that docstrings are accurate, e.g., I just noticed that DensityFeatures doesn't have one by browsing the above)

WardLT:
I agree, maintaining a table by hand would be impractical.

What I was thinking of was something like option 2: Writing a script that will retrieve the first sentence of the docstring for each featurizer and combine them into table. We'd have to ensure that the first line is an adequate summary of the featurizer, but that high-level summary would hopefully remain accurate even with changes to featurizer (e.g., adding new features). 

The main advantage I see for this is being able to quickly view all of the featurizers for a certain type of data without having to scroll through the detailed documentation page. 

Does that sound like a good approach?

computron:
That sounds good to me, as long as it is leveraging the existing docstring and not trying to create a new description

ardunn:
@WardLT do you have a WIP for this? If not, I will try my hand at writing a script. I think it would give a huge usability boost to matminer. Especially if we could have a readable table for each 'type' (structure, site, composition, etc.) somewhere on the matminer docs pages ...  for example, a 'featurizers summary' page where the auto-generated tables would reside.

WardLT:
I have not started on this and agree a table for each type would be awesome.

I look forward to seeing what you come up with!

ardunn:
Alright, does the below example image (only of structure and compositon) look acceptable? if so, I can make a PR. 

[...]

PR #200, opened 2018-04-11
Added a feature table/summary to docs + regen docs

ardunn:
## Summary
- Added script matminer/utils/featurizer_summary.py called by `invoke make-doc` to automatically generate featurizer table. This will account for any new BaseFeaturizer subclasses added to any existing featurizer module. _If new featurizer modules are added to matminer, this file will need to be updated by importing the new module and adding it's high level description in featurizer_summary.py._
- Added feature table/summary to a new page in docs
- Put link to featurizer page on main (index) page
- Added a featurization graphic bc index.rst was text heavy
- Regenerated sphinx docs to include all new featurizers (incl. new structure ones @WardLT wrote, + new BoB)
- Added suitable first line descriptions to each featurizer. 
- Added guideline for writing new featurizer docs. The first line of a new featurizer doc has to be a short, high-level description of the featurizer for this script to work correctly. 
- Added elastic dataset static html to embed in index. It's interactive in the webpage, which is nice.
 
## Screenshots
image
image
image

## Notes
Clicking on the [more] links to the module page as per @computron suggestion
image


ardunn:
Also added the elastic dataset Figrecipes example in interactive form to the docs via RST iframe. 
Here is screenshot:

image

ardunn:
Closes #193

WardLT:
I think this looks pretty sharp! 

Specially, I like that the table will automatically rebuild with the new documentation (if I understand Sphinx right), that the table links to detailed documentation, abd that you've updated the BaseFeaturizer documentation guide. 

The interactive plot is also a nice addition to the webpage.

One question: Should we move the table generation script to the `doc` folder? It seems a little out-of-place in the main module, since it will likely never be called by users.

ardunn:
@WardLT sure, that makes more sense logistically.

computron:
I know there are some other Featurizers up in the air right now, but I don't want to wait for those to merge this. We'll need to update the docs from time to time anyway

PR #210, opened 2018-04-20
debug heatmap_df + add color_range arg

albalu:
yaxis title handling had a bug that was fixed + the range of data (or z-axis or colorbar) can now be set via color_range argument.

These changes are backwards compatible and small enough that don't need new tests. Finally, heatmap.py in matminer_examples run as expected.


ardunn:
Looks good my be.

PR #211, opened 2018-04-23
MultiFeaturizer handle 1D arrays w/ and w/o singlton dimensions

WardLT:
Makes MultiFeaturizer more robust to how the underlying featurizers represent 1D arrays. If the classes each use a different method (e.g., shape of (4,1) vs shape == (4,)), MultiFeaturizer errors with a vague `ValueError: all the input arrays must have same number of dimensions`. This PR prevents this error, and includes a new unittest for catching this bug.


computron:
Thanks!

PR #214, opened 2018-04-27
Sort out reference for cohesive energies

### LINKED ISSUE #145: Add more descriptive comments to citations for CohesiveEnergy
@saurabh02 

See this todo:
```
# TODO: @sbajaj unclear whether cohesive energies are taken from first ref, second ref, or combination of both
```
matminer.featurizers.composition.CohesiveEnergy#citations

You can add some code comments that explains what is going on. This could be part of the main doc for the CohesiveEnergy featurizer, e.g.
```
        Class to get cohesive energy per atom of a compound by adding known
        elemental cohesive energies from the formation energy of the
        compound. <<SOME MORE DESCRIPTION HERE ABOUT KNOWN COHESIVE ENERGIES SOURCE>>
```

saurabh02:
To resolve #145 . 
I used KnowledgeDoor to get the cohesive energies, which in turn got the data from Kittle. 
Since Kittle is already reference on KnowledgeDoor, I remove it from the citation here.

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -938,9 +938,7 @@ class CohesiveEnergy(BaseFeaturizer):
         return ["Saurabh Bajaj", "Anubhav Jain"]
 
     def citations(self):
-        # TODO: @sbajaj unclear whether cohesive energies are taken from first ref, second ref, or combination of both
-        return [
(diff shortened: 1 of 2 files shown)


computron:
Thanks! I think it is good to have both citations since the Kittel one is more original and also a more permanent citation. I have pushed a commit 	481f5b66b18a7704f0327b1ddc26b8cb40de72a4 to do that.

Issue #217, opened 2018-04-29
functionfeaturizer doesn't work with MultipleFeaturizer


montoyjh:
Because of the column dependence of FunctionFeaturizer (which we solved previously by making feature_labels() None and postprocessing the dataframe in featurize_dataframe), it doesn't play nicely with MultipleFeaturizer.  This seems not ideal, since I'd imagine function featurizer would often be used in a multiplefeaturizer context, but we're going to need to revisit how to flexibly handle feature_labels when feature_labels depend on input columns if we want to fix it.

computron:
@montoyjh should I close this?

montoyjh:
Yep, I believe it's been addressed.

PR #218, opened 2018-05-01
Function featurize

montoyjh:
A few changes to address #217.

Main change is in the addition of a `generate_feature_labels` method which takes column_ids as input.  Most Featurizers will not use column identity in generating labels, so the default is to return feature_labels, but the addition of this method in the base class allows developers of custom featurizers the ability to customize column names if desired without altering the implicit documentation role of the feature_labels method.

Also added a unittest to ensure FunctionFeaturizer works with MultipleFeaturizer.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -365,8 +365,10 @@ class MultipleFeaturizer(BaseFeaturizer):
     def featurize_dataframe(self, df, col_id, ignore_errors=False,
                             return_errors=False, inplace=True):
         for f in self.featurizers:
+            f.set_n_jobs(1)
             df = f.featurize_dataframe(df, col_id, ignore_errors,
                                        return_errors, inplace)
+            df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
         return df
 
     def citations(self):
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -368,10 +368,24 @@ class MultipleFeaturizer(BaseFeaturizer):
 
     def featurize_dataframe(self, df, col_id, ignore_errors=False,
                             return_errors=False, inplace=True):
-        for f in self.featurizers:
-            df = f.featurize_dataframe(df, col_id, ignore_errors,
-                                       return_errors, inplace)
-            df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
+        """
+        Featurize dataframe is overloaded in order to allow
+        compatibility with Featurizers that overload featurize_dataframe
+        """
+        # Detect if any featurizers override featurize_dataframe
+        override = ["featurize_dataframe" in f.__class__.__dict__.keys()
+                    for f in self.featurizers]
+        if any(override):
+            warnings.warn(
+                "One or more featurizers overrides featurize_dataframe, "
+                "featurization will be sequential and may diminish performance")
+            for f in self.featurizers:
+                df = f.featurize_dataframe(df, col_id, ignore_errors,
+                                           return_errors, inplace)
+                df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
+        else:
+            df = super(self, MultipleFeaturizer).featurize_dataframe(
+                df, col_id, ignore_errors, return_errors, inplace)
         return df
 
     def citations(self):
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -181,7 +181,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
             col_id = [col_id]
 
         # Generate the feature labels
-        labels = self.feature_labels()
+        labels = self.generate_feature_columns(col_id)
 
         # Check names to avoid overwriting the current columns
         for col in df.columns.values:
@@ -340,6 +340,21 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
         raise NotImplementedError("implementors() is not defined!")
 
+    def generate_feature_columns(self, col_id):
+        """
+        Function that returns new feature columns ids, exists primarily
+        so any method which needs to modify new column names based on
+        input column identity can do so without altering feature_labels,
+        by default this will simply return feature_labels()
+
+        Args:
+            col_id ([str]): input column ids
+
+        Returns:
+            list of column labels
+        """
+        return self.feature_labels()
+
 
 class MultipleFeaturizer(BaseFeaturizer):
     """
@@ -361,6 +376,9 @@ class MultipleFeaturizer(BaseFeaturizer):
 
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
+
+    def generate_feature_columns(self, col_id):
+        return sum([f.generate_feature_columns(col_id) for f in self.featurizers], [])
 
     def citations(self):
         return list(set(sum([f.citations() for f in self.featurizers], [])))
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -384,7 +384,7 @@ class MultipleFeaturizer(BaseFeaturizer):
                                            return_errors, inplace)
                 df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
         else:
-            df = super(self, MultipleFeaturizer).featurize_dataframe(
+            df = super(MultipleFeaturizer, self).featurize_dataframe(
                 df, col_id, ignore_errors, return_errors, inplace)
         return df
 
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -362,9 +362,6 @@ class MultipleFeaturizer(BaseFeaturizer):
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
 
-    def generate_feature_columns(self, col_id):
-        return sum([f.generate_feature_columns(col_id) for f in self.featurizers], [])
-
     def citations(self):
         return list(set(sum([f.citations() for f in self.featurizers], [])))
 
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -181,7 +181,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
             col_id = [col_id]
 
         # Generate the feature labels
-        labels = self.generate_feature_columns(col_id)
+        labels = self.feature_labels()
 
         # Check names to avoid overwriting the current columns
         for col in df.columns.values:
@@ -339,21 +339,6 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
         """
 
         raise NotImplementedError("implementors() is not defined!")
-
-    def generate_feature_columns(self, col_id):
-        """
-        Function that returns new feature columns ids, exists primarily
-        so any method which needs to modify new column names based on
-        input column identity can do so without altering feature_labels,
-        by default this will simply return feature_labels()
-
-        Args:
-            col_id ([str]): input column ids
-
-        Returns:
-            list of column labels
-        """
-        return self.feature_labels()
 
 
 class MultipleFeaturizer(BaseFeaturizer):
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -359,13 +359,16 @@ class MultipleFeaturizer(BaseFeaturizer):
     def featurize(self, *x):
         return np.hstack(np.squeeze(f.featurize(*x)) for f in self.featurizers)
 
+    def set_n_jobs(self, n_jobs):
+        for f in self.featurizers:
+            f.set_n_jobs(n_jobs)
+
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
 
     def featurize_dataframe(self, df, col_id, ignore_errors=False,
                             return_errors=False, inplace=True):
         for f in self.featurizers:
-            f.set_n_jobs(1)
             df = f.featurize_dataframe(df, col_id, ignore_errors,
                                        return_errors, inplace)
             df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -185,7 +185,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
         # Check names to avoid overwriting the current columns
         for col in df.columns.values:
-            if labels and col in labels:
+            if col in labels:
                 raise ValueError('"{}" exists in input dataframe'.format(col))
 
         # Compute the features
@@ -361,6 +361,13 @@ class MultipleFeaturizer(BaseFeaturizer):
 
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
+
+    def featurize_dataframe(self, df, col_id, ignore_errors=False,
+                            return_errors=False, inplace=True):
+        for f in self.featurizers:
+            df = f.featurize_dataframe(df, col_id, ignore_errors,
+                                       return_errors, inplace)
+        return df
 
     def citations(self):
         return list(set(sum([f.citations() for f in self.featurizers], [])))
--- a/matminer/featurizers/function.py
+++ b/matminer/featurizers/function.py
@@ -119,10 +119,9 @@ class FunctionFeaturizer(BaseFeaturizer):
         as the base class
 
         Args:
-            df (Pandas dataframe): Dataframe containing input data.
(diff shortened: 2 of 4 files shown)


computron:
I don't really like this and I'm sure you don't either.

I don't think it's clear to me why this can't be incorporated into the existing ``fit()`` framework. The goal of ``fit()`` is just to set all internal variables that depend on your data choice.

Yes, ``fit()`` right now only takes in ``X`` and ``y`` data - but it also has the option to add ``fit_kwargs`` like col_ids or whatever you need for FunctionFeaturizer to set the internal label variables.  If the FunctionFeaturizer used fit() to get the right labels set internally, we'd be all set - is that correct?

Next, ``fit_featurize_dataframe()`` in BaseFeaturizer doesn't yet allow for throwing in ``**kwargs**`` to fit. But I think this is also fixable.

Are there other issues to using the existing fit() framework?

montoyjh:
Yeah, I'm definitely not crazy about it, but I have two issues with using fit.  The first is that using `fit` to determine qualitative identity of output data seems semantically counterintuitive to me.  If I'm "fitting" something, I'd imagine I'm determining quantitative model parameters from data.  It's not obvious to me that I'd need to use it to determine qualitative labels for output.  

The other issue is that it makes any featurizer that relies on input column identity to describe feature labels stateful in a way that makes it prone to both soft and hard errors.  For example, I might generate labels that correspond to a certain set of columns, and then reuse the featurize_dataframe method with different columns.  If the number of columns is different, this probably throws an error telling the user something vague about how two numbers aren't the same 5 layers into a pandas traceback, but if they use the correct number of columns (but different columns), then featurize dataframe works fine but labels all of the columns incorrectly.

ardunn:
@montoyjh 

regarding the 2nd issue:  I think some of the label mishap can be avoided by explicitly declaring what to do when a fittable featurizer is fit on a dataframe (generating one feature label set) and then featurizes another dataframe (assigning values based on a different potential feature label set). As far as I can tell, there are 4 scenarios where problems can occur:

1. Fitted set of labels has some labels not in the featurized set
2. Featurized set has some labels not in the fitted set
3. A combination of 1 and 2
4. No overlap between featurized set labels and fitted set labels

In BondFractions, for example, the fitted set of labels is always returned. 1 is taken care of by setting bond values in fitted set but not in featurized set to zero. 2 is taken care of by optionally approximating values for featurized labels to values for fitted labels, or dropping them otherwise. 3 is taken care of by using 1 and 2 in sequence. And 4 is taken care of by throwing an exception. Not saying a BondFractions-like implementation is ideal (or applicable to FunctionFeaturizer), but I haven't (yet) run into bad label assignments or vague 5-layer pandas tracebacks when fitting and featurizing on different sets.

montoyjh:
Alright, I think this is moving in the prescribed direction.  What I really want to avoid is forcing the _user_ to use fit when using FunctionFeaturizer, because I don't want them to have to provide the same information (i. e. input column names) in multiple places.  In this implementation I put all of the logic for generating the labels in fit, and call it appropriately in both featurize_dataframe and fit_featurize_dataframe (because in this case it should be used in both).  The user could still use fit if they really wanted, but it wouldn't mess with the operation of featurize_dataframe.  I still need to resolve an issue in MultiFeaturizer compatibility related to the parallel use of featurization, which I'll tackle tomorrow.

computron:
Btw @WardLT mentioned to me that he had some comments but is currently on travel. I was planning to wait for that before merging

Let me know however if you think this needs to be merged sooner rather than later

montoyjh:
Sure, no reason to rush.  This is working now.  In `MultipleFeaturizer`, I've overriden featurize_dataframe to featurize the dataframe using each featurizer sequentially, which I think should allow any featurizer that overrides featurize_dataframe (as FunctionFeaturizer does) to be compatible.

computron:
Ok great! I'll wait another week in case there are more reviews, otherwise I'll merge at that time and we can modify more after merging.

matminer/featurizers/function.py
```diff
@@ -73,15 +75,53 @@ def exp_dict(self):
...
+        Sets the feature labels.  Not intended to be used by a user,
+        only intended to be invoked as part of featurize_dataframe
+
+        Args:
+            X ([str]): list of feature labels
+
+        Returns:
+            Set of feature labels corresponding to expressions
+        """
+        self._feature_labels = self.generate_string_expressions(X)
```
WardLT: X should be a list of entries, and a not a list of column names.

I'd imagine the column names could be passed to this argument a few ways:
1. An optional keyword argument
2. The code detects if X is a dataframe and, if so, uses the column labels.

If the code does not detect label names, we could always make them up (e.g., "x0", "x1").

My reasoning for these suggestions are to make this featurizer follow a similar style to other TransformerMixin classes (which take data as input, not labels), and to allow people to  easily use this class without dataframes (I want to do that)

matminer/featurizers/base.py
```diff
@@ -359,9 +359,21 @@ def __init__(self, featurizers):
     def featurize(self, *x):
         return np.hstack(np.squeeze(f.featurize(*x)) for f in self.featurizers)
 
+    def set_n_jobs(self, n_jobs):
+        for f in self.featurizers:
+            f.set_n_jobs(n_jobs)
```
WardLT: Good call

matminer/featurizers/function.py
```diff
@@ -73,15 +75,53 @@ def exp_dict(self):
...
+            list of functional outputs corresponding to input args
+        """
+        return list(self._exp_iter(*args, postprocess=self.postprocess))
+
+    def feature_labels(self):
+        """
+        Returns:
+            Set of feature labels corresponding to expressions
+        """
+        return self._feature_labels
```
WardLT: Should we add in a "not fitted" error here?

matminer/featurizers/function.py
```diff
@@ -91,46 +131,31 @@ def featurize_dataframe(self, df, col_id, ignore_errors=False,
...
-        return list(self._exp_iter(*args, postprocess=self.postprocess))
-
-    def feature_labels(self):
-        """
-        Returns:
-            Set of feature labels corresponding to expressions
+            updated dataframe based on featurizer fitted to that dataframe.
         """
-        return None
+        return self.featurize_dataframe(df, col_id, *args, **kwargs)
```
WardLT: Do we need fit_featurize_dataframe if all it does is call featurize_dataframe?

matminer/featurizers/function.py
```diff
@@ -73,15 +75,53 @@ def exp_dict(self):
...
+        """
+        Returns:
+            Set of feature labels corresponding to expressions
+        """
+        return self._feature_labels
+
+    def fit(self, X, y=None, **fit_kwargs):
+        """
+        Sets the feature labels.  Not intended to be used by a user,
+        only intended to be invoked as part of featurize_dataframe
```
WardLT: This warning seems out of place, if you agree with my next comment.

matminer/featurizers/base.py
```diff
@@ -359,9 +359,21 @@ def __init__(self, featurizers):
...
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
 
+    def featurize_dataframe(self, df, col_id, ignore_errors=False,
+                            return_errors=False, inplace=True):
+        for f in self.featurizers:
+            df = f.featurize_dataframe(df, col_id, ignore_errors,
+                                       return_errors, inplace)
+            df[f.feature_labels()] = df[f.feature_labels()].applymap(np.squeeze)
+        return df
```
WardLT: This change is going to wreck havoc with caching. If I understand multiprocessing correctly, each time featurize_dataframe completes the processes holding the cache will exit - thereby preventing data cached from one featurizer to be employed by another.

So, I think we need to be really sure we want this change to featurize data frame. Can you convince me? 

Fwiw, I'm not concerned FunctionFeaturize need work with MultiFeaturizer. I picture it being used in a Pipleline after primary features are computed but not in parallel (as assumed by MultiFeaturizer)

matminer/featurizers/tests/test_function.py
```diff
@@ -82,6 +87,21 @@ def test_helper_functions(self):
...
+        self.assertEqual(len(test_combo_4), 18)
+
+    def test_multi_featurizer(self):
+        ff1 = FunctionFeaturizer(expressions=["x ** 2"])
+        ff2 = FunctionFeaturizer(expressions=["exp(x)", "1 / x"])
+        mf = MultipleFeaturizer([ff1, ff2])
+        new_df = mf.featurize_dataframe(self.test_df, ['a', 'b', 'c'],
+                                        inplace=False)
+        self.assertEqual(len(new_df), 11)
+
```
WardLT: Can we add a test for this class behaving well in a Pipeline? That's where I want to use it (and figure that may be a common use case)

matminer/featurizers/tests/test_function.py
```diff
@@ -82,6 +87,21 @@ def test_helper_functions(self):
...
+                                                         combo_function=np.sum)
+        self.assertEqual(len(test_combo_4), 9)
+        test_combo_4 = generate_expressions_combinations(["1 / x", "x ** 2", "exp(x)"],
+                                                         combo_function=lambda x: x[1]-x[0])
+        self.assertEqual(len(test_combo_4), 18)
+
+    def test_multi_featurizer(self):
+        ff1 = FunctionFeaturizer(expressions=["x ** 2"])
+        ff2 = FunctionFeaturizer(expressions=["exp(x)", "1 / x"])
+        mf = MultipleFeaturizer([ff1, ff2])
```
WardLT: Ok. This might sell the need for FF to have MultiFeaturizer compatibility, but I still worry about giving up caching to save this. 

A few questions:

- Could we achieve the ability to make features based on functions of features that are functions of other features another way? Is this not what the depth argument provides? 
- Could we get it both ways by making a MultiFunctionFeaturizer? 
- Any other thoughts?

matminer/featurizers/function.py
montoyjh: Agree and prefer 2.  I tried 1 in another PR, and my initial approach to that was pretty unwieldy.  Latest commits have this.

matminer/featurizers/function.py
montoyjh: Added, tested.

matminer/featurizers/tests/test_function.py
montoyjh: I don't know what a Pipeline is.  Someone who will use it that way should probably add that test (and any of the necessary functionality).

matminer/featurizers/base.py
montoyjh: Hmm... this seems like an important problem but I don't know how I'd fix it off of the top of my head.  Any featurizer that overrides featurize_dataframe will end up needing something like this, though (I don't know if any of the other featurizers do this).

To your last point, my main reason for trying to make this work is more practical, which is to say that someone actually tried to use it that way, it didn't work, and they e-mailed us to ask why. I think new users encountering a lack of interoperability of the code is bad for its perceived robustness, but I also think you're right that the more important supported use case should be in sequential applications of featurization, rather than parallel.

montoyjh: Also, can you give me an example where this causes the problem you mention in a significant way?  I'm confident there is one, I just don't know the ins and outs of featurizers that use caching well enough to come up with an example on my own.

matminer/featurizers/tests/test_function.py
montoyjh: To (1) - this doesn't achieve that, this test isn't really the intended use case, it's just to ensure that it works with MultiFeaturizer.  Thus, I don't think (2) is necessary.

One thing we could do is to just to create two modes of MultiFeaturizer operation, one "fast" mode that uses the prior method if FunctionFeaturizer (or other featurizers that override featurize_dataframe) is not present, and one "slow" mode that warns the user if a MultiFeaturizer-caching-incompatible featurizer is present.  This way, it still works if you use FunctionFeaturizer (and shouldn't be too slow if they're not using a bunch of features that are accelerated by caching) but issues a relevant warning that users can pay attention to if they're concerned about efficiency.

computron: I haven't played around too much with sklearn Pipeline, but perhaps someone can explain what MultiFeaturizer achieves that Pipeline can't. i.e., should we just encourage everyone to do multi-featurization in Pipelines?

WardLT: Scikit-learn's Pipeline tool is used to construct models that are composed of multiple, sequential operations. A simple example would be a model that first converts a pmg Composition into a feature vector and then trains a linear regression model . The distinguishing feature of a Pipeline is that the outputs from one step are used as the inputs into the next one. In contrast, MultiFeaturizer is provides the same inputs to all of the suboperations.

An example that uses MultiFeaturizer, FunctionFeaturizer, and Pipeline together is:

```python
model = Pipeline([
  ('featurize', MultiFeaturizer([ElementProperty(), IonProperty()])),
  ('expand', FunctionFeaturizer()),
  ('model', Lasso())
])
model.fit([Composition('Al')], [0])
```

This Pipeline will first call MultiFeaturizer transform the list of Compositions into a feature array, then use FunctionFeaturizer to expand those primary features, and then use the expanded list of features to fit a Lasso model. 

Sorry for the long reply, but does this help clarify how all of these tools could fit together?

matminer/featurizers/base.py
WardLT: The Voronoi notebook in our [examples repo](https://github.com/hackingmaterials/matminer_examples/blob/master/notebooks/machine-learning/voronoi-ward-prb-2017.ipynb) uses caching with MultiFeaturizer.

matminer/featurizers/tests/test_function.py
computron: I see, but why can't someone do it this way?

```
model = Pipeline([
  ('featurize1', ElementProperty()),
  ('featurize2', IonProperty()),
  ...
  ('expand', FunctionFeaturizer()),
  ('model', Lasso())
])
```

If the above is valid, then the main use case I see is that MultiFeaturizer is just more intuitive than sklearn Pipelines. Technically, MultiFeaturizer could also parallelize across featurizers (instead of across data within featurizers), but I don't think that's being exploited yet.

But given that Pipelines do more, my question is whether we should encourage people to bite the bullet and learn Pipelines - e.g., by giving them an easy example notebook to work from. Since this is a more complex discussion I'll add it as a separate issue.

computron:
@montoyjh @WardLT 

I'm a bit lost on the status of this PR.

Is this ready to merge?

If not, what steps are remaining?

matminer/featurizers/tests/test_function.py
computron: Or are you saying that the above *doesn't* work since ElementProperty doesn't just return an array of Compositions (rather a matrix of Composition + other features), so the IonProperty() step would fail?

WardLT: Yes, the above code you just provided would not work because ElementProperty returns a list of features and not a list of Compositions, as needed by IonProperty.

WardLT:
I don't think this is ready to merge. 

The one critical issue, for me, is that overloading `featurize_dataframe` kills our caching. If we could fix that, then I'm satisfied. The other things we can address later.

montoyjh:
We need to decide on how we want multifeaturizer to work, we have three options currently proposed

(1) - What's in this pull requests changes `MultiFeaturizer.featurize_dataframe` to run the featurizers sequentially, rather than in aggregate using the `MultiFeaturizer.featurize` method and the inherited `BaseFeaturizer.featurizer_dataframe`.  FunctionFeaturizer is compatible with MultiFeaturizer (and presumably any Featurizer that overrides featurize_dataframe is also), but we are left with potential issues with caching.
(2) - I can revert the changes so we keep what's been changed in FunctionFeaturizer (which I think is an improvement, even if #217 isn't addressed).  FunctionFeaturizer will not be compatible with MultiFeaturizer.
(3) - We could adopt a default mode for multifeaturizer if none of the included featurizers override featurize_dataframe which would do everything in parallel, and then a sequential mode if any of the included featurizers override featurize_dataframe, issuing a warning that performance may be hindered by the sequential operation.

montoyjh:
@WardLT How do you feel about option 3 above?  I'm not crazy about making MultiFeaturizer more complicated, but it presumably wouldn't change anything about how MatMiner performs now.

WardLT:
Options 3 sounds OK to me. I share your concern about making MultiFeaturizer complicated, but it would allow us to have the best of options 1 and 2 (i.e., caching and MF compatibility). 

How do you detect if a function has been overloaded?

WardLT:
Also, in the interest of closing this pull request, let's tackle the test for Pipeline compatibility in another PR. I might be able to work on that early next week.

montoyjh:
Okay, I modified such that option 3 is implemented as indicated above (with override detection in the class dict), let me know how that looks.

WardLT:
This looks good to me. Thanks @montoyjh !

PR #221, opened 2018-05-10
Update Citrine data retrieval to citrination-client v4.0

saurabh02:


computron:
Thanks! Can you also update the requirement in setup.py?

computron:
Thanks!

PR #222, opened 2018-05-11
added the capability of data=list of lists even with different lengths (for PlotlyFig histogram)

albalu:
also updated the doc


ardunn:
looks ok by me

PR #224, opened 2018-05-12
Added cluster packing efficiency attributes

WardLT:
Added features related to the cluster packing efficiency model of [Laws et al.](http://www.nature.com/doifinder/10.1038/ncomms9123)

These are some of the features used in a recent paper by [Ren et al.](http://advances.sciencemag.org/content/4/4/eaaq1566)

Sorry for the extra changes. I did not realize until after I committed that I ran "reformat" code on the whole file, which ensured that no lines are longer than the specified maximum. Would you like me to go and change those lines back?

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -2,14 +2,16 @@ from __future__ import division
 
 import itertools
 import os
-from functools import reduce
+from functools import reduce, lru_cache
 import collections
 
 import numpy as np
 import pandas as pd
 from pymatgen import Element, MPRester
+from pymatgen.core.composition import Composition
 from pymatgen.core.periodic_table import get_el_sp
 from pymatgen.core.molecular_orbitals import MolecularOrbitals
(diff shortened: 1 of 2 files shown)


computron:
For the reformat, it's generally better to do as a separate commit but we can work with it. I'd just make sure that any auto-reformat is readable, sometimes it can make things worse and a manual reformat is needed.

computron:
Also a note that I typically prefer a 100 char max line length (against PEP guidelines) but don't have a strong preference for 80 vs 100 so long as it's in the interest of code readability.

WardLT:
I'll set my editor to 100. I try to avoid running it on the whole file, I just happened to miss it this time (if I don't highlight text when calling re-format, it runs the whole file rather than my highlight).

The changes do affect readability, IMO. I'm currently reverting them.

WardLT:
Alright, I got rid of the most ugly changes. I'll be more careful next time.

computron:
Thanks!

PR #225, opened 2018-05-14
add optional args weights and atol in BranchPointEnergy

albalu:
- added the optional arg "weights" for example for when reading the BandStructure from pymatgen Vasprun and weights are already available (hence making the BranchPointEnergy calculations much faster).

- Also added atol argument to be able to directly access/set this arg for get_kpoint_weights for when weights is None and they are supposed to be calculated automatically.

### DIFF
--- a/matminer/featurizers/bandstructure.py
+++ b/matminer/featurizers/bandstructure.py
@@ -28,12 +28,13 @@ class BranchPointEnergy(BaseFeaturizer):
         calculate_band_edges: (bool) whether to also return band edge
             positions
     """
-    def __init__(self, n_vb=1, n_cb=1, calculate_band_edges=True):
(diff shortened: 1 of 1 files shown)


computron:
Thanks!

Do you think the KPoint object in pymatgen should have a ``weight`` property in the future? That way bs.kpoints might have the weights embedded in them rather than having to get them again via SpacegroupAnalyzer

computron:
(merging for now) - added an issue to pymatgen: https://github.com/materialsproject/pymatgen/issues/1148

albalu:
Yes. I can't think of any reason why bs.kpoints can't have weights already that defaults to None or sth. I think I checked for this first but didn't find it.

PR #228, opened 2018-05-21
small debug

albalu:
limit=0 instead of None does the same thing cleaner. In this way there is no need for total = min(limit, r.count()) and check for limit is None, etc.


computron:
Thanks!

Note that I find "limit=None" (no limit) to be a little clearer than "limit=0" (what does this mean). But given that MongoDB supports 0 limit (and incidentally also negative limit) I think this is OK

PR #229, opened 2018-05-21
apply shift when calculating bpe + return unchanged vbm/cbm

albalu:
- Previously if target_gap was provided, the branch_point_energy (bpe) calculation wouldn't change.

- Now the unchanged vbm/cbm is returned. The users can subsequently subtract bpe (or vbm/cbm or anything) themselves if they want (i.e. the reference is not decided to be bpe by default).


albalu:
@computron , I know we talked bout this once. I still find this definition of "absolute" more intuitive and easier to work with: it doesn't subtract anything from vbm/cbm by default so I think it's also less surprising.

PR #230, opened 2018-05-22
Add stacked featurizer

WardLT:
This featurizer let you use the output of a machine learning as a feature.


matminer/featurizers/base.py
```diff
@@ -393,3 +392,65 @@ def citations(self):
...
+        if self._is_classifier():
+            output = self.model.predict_proba(features)[0]
+            return output[:-1]
+        else:
+            return self.model.predict(features)
+
+    def feature_labels(self):
+        name = self.name or ''
+        if self._is_classifier():
+            if self.class_names is None:
```
computron: should this check go in the __init__() ?

Pro: avoid a situation where you spent time featurizing the data and then get an error only at the end when calling feature_labels(), forcing you to go back to the beginning
Pro: throw the error right away as soon as you know there might be a problem so the user fixes it right away
Con: *Maybe* you want to allow not setting ``class_names`` for a classifier if someone doesn't care about feature_labels? If this is the case, you can still consider throwing a warning that feature_labels won't work if you don't set class_names

computron: (the warning would come at initialization)

matminer/featurizers/base.py
```diff
@@ -393,3 +392,65 @@ def citations(self):
...
+    """
+
+    def __init__(self, featurizer=None, model=None, name=None, class_names=None):
+        """Initialize featurizer
+
+        Args:
+            featurizer (BaseFeaturizer): Featurizer used to generate inputs to the model
+            model (BaseEstimator): Fitted machine learning model to be evaluated
+            name (str): [Optional] name of model, used when creating feature names
+            class_names (str): Required for classification models, used when creating
```
computron: is class_names an array? doc lists as str

Also if an array, does it need to be in any special order?

WardLT: Oops. It is an array, I'll clarify that and the order.

matminer/featurizers/base.py
WardLT: I like the idea of printing a warning for init() that class names are required for `featurize_dataframe` and `feature_labels` to function, but not throwing an error.

PR #231, opened 2018-05-23
Initial Draft of a Design Module

WardLT:
This commit adds a module with tool generate search spaces over which to apply an ML model.

As a start, I added in a class which generates equally-spaced compositions.

A few questions that I already have for you all:

1. Should matminer contain this feature?
2. Do you like the name "design" for the module? 
3. Would any other helper functions in the base class be useful?
4. Should I implement any other generates (e.g., crystal structure decorator) as examples?

Note: This PR contains a commit from #230, so we should hold until that one is approved. (Sorry, I needed to use features from both of these PR in a project).


computron:
Hi Logan

How does the enumeration compare to SMACT for example?

https://github.com/WMD-group/SMACT

I haven't used SMACT but I think it was intended to enumerate stoichiometric chemical spaces with constraints like charge balance built in (see also https://www.cell.com/chem/abstract/S2451-9294(16)30155-3)

I think some of the responses to the other questions might depend on that one. At the moment I would be leaning  towards "sure put it in matminer if you're willing to support it, but probably it's the topic of a different codebase that perhaps has matminer as a dependency"

ardunn:
@WardLT cool idea! I would offer one warning of this implementation though. In generating the search space with itertools product (in `CompositionGenerator`) it is a pretty common occurrence to generate huge objects in memory from surprisingly small search spaces. Not sure if you have run into that issue yet but it is something to note

computron:
@WardLT any update on this PR?

If no, I would suggest to close it and re-open if there are plans in the future.

PR #242, opened 2018-06-04
Need a deep-copy for cn_target_motif_op in site.py

nisse3000:
There was an undesirable behavior. Consider this snippet:

```
from matminer.featurizers.site import CrystalNNFingerprint, OPSiteFingerprint

cnnfp = CrystalNNFingerprint.from_preset('ops')
opsfp = OPSiteFingerprint()

cnnfp.featurize(s, 0)
opsfp.featurize(s, 0)
```

It returned following error:

```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    opsfp = OPSiteFingerprint()
  File "matminer/matminer/featurizers/site.py", line 224, in __init__
    self.ops[cn].append(LocalStructOrderParams([ot], parameters=[p]))
  File "pymatgen/pymatgen/analysis/local_env.py", line 1397, in __init__
    t + ")!")
ValueError: Unknown order parameter type (wt)!
```

The problem was that we were using shallow copies of the dictionary. I've also tested the 
`dict2 = dict(dict1)`
copying mechanism and the update function.
```
dict2 = dict()
dict2.update(dict1)
```
Neither worked. Only, the deepcopy from the copy module worked.

### DIFF
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -204,7 +204,7 @@ class OPSiteFingerprint(BaseFeaturizer):
     def __init__(self, target_motifs=None, dr=0.1, ddr=0.01, ndr=1, dop=0.001,
                  dist_exp=2, zero_ops=True):
         self.cn_target_motif_op = copy.deepcopy(cn_target_motif_op) \
-            if target_motifs is None else target_motifs.copy()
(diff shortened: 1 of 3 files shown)


computron:
Can you just add a quick unit test?

You can use the code snippet from the top of the file. So if anyone breaks the copy mechanism, the test will fail.

PR #250, opened 2018-06-08
 Finished first version CrystalNNFingerprint augmented by chemical information.

nisse3000:
The main idea is that the user provides a dictionary that maps elemental symbol to value for any given property that s/he wants to add (e.g., atomic mass, electronegativity). Using a dictionary seems to provide largest flexibility so that you could get the Pauling scales/electronegativities from whatever source/calculation that you like. Per property, the average absolute difference between central site property and neighbors is added that is averaged using the coordination likelihoods (wt) as weights. That is, if wt is zero, a given coordination number environment doesn't contribute. The underlying "calculation" formula is (for mass, m):
screen shot 2018-06-08 at 3 44 34 pm


matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
 
         cn_fingerprint = []
 
+        if self.chem_info is not None:
+            absdeltas = {}
```
computron: suggest refactor this variable to be named:

``prop_absdelta`` since it is a dictionary of properties to delta values

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
+            for prop in self.chem_props:
+                absdeltas[prop] = 0
+            Nwt = 0
```
computron: suggest rename ``Nwt`` to ``total_weight``

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
+            elem_cent = struct.sites[idx].species_string
```
computron: suggest rename ``elem_cent`` to ``elem_central``

computron: Note also that the species string can be problematic

Often the species is something like ``Al3+`` but your property dict might have values for elements only (e.g., a value for ``Al``). Ideally if you don't have a value for the specific species, the code would fall back on just using the element specific value if you have it. e.g., if you didn't define a value for ``Al3+`` but did define a value for ``Al``, it will just use the ``Al`` value.

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
+                        if self.chem_info is not None and wt != 0:
+                            Nwt += wt
+                            for prop in self.chem_props:
+                                delem = self.chem_info[prop]
+                                tmp = 0
+                                for d in nndata.cn_nninfo[cn]:
+                                    tmp += delem[d["site"].species_string] - \
+                                        delem[elem_cent]
+                                absdeltas[prop] += wt * tmp / cn
+                    elif wt == 0:
```
computron: I don't understand this ``elif wt == 0`` statement ...

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
+            if cn in self.ops:
                 for op in self.ops[cn]:
-                    if op == "wt" or wt == 0:
+                    if op == "wt":
+                        cn_fingerprint.append(wt)
+                        # Compute additional chemistry-related features
+                        if self.chem_info is not None and wt != 0:
+                            Nwt += wt
+                            for prop in self.chem_props:
+                                delem = self.chem_info[prop]
```
computron: suggest rename ``delem`` to ``elem_prop``

computron: or ``elem_propval``

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
+                                tmp = 0
+                                for d in nndata.cn_nninfo[cn]:
+                                    tmp += delem[d["site"].species_string] - \
```
computron: see again my comment about ``species_string`` vs just the element symbol for this portion of the code.

matminer/featurizers/site.py
```diff
@@ -464,14 +472,31 @@ def featurize(self, struct, idx):
...
                 for op in self.ops[cn]:
-                    if op == "wt" or wt == 0:
+                    if op == "wt":
+                        cn_fingerprint.append(wt)
+                        # Compute additional chemistry-related features
+                        if self.chem_info is not None and wt != 0:
+                            Nwt += wt
+                            for prop in self.chem_props:
+                                delem = self.chem_info[prop]
+                                tmp = 0
```
computron: generally this kind of ``tmp`` variable is discouraged. You'd instead do something like

```
my_var = sum([delem[d["site"].species_string] - delem[elem_cent] for d in nndata.cn_nninfo[cn]])
```

matminer/featurizers/site.py
```diff
@@ -481,15 +506,23 @@ def featurize(self, struct, idx):
                                             range(1, len(neigh_sites) + 1)])[0]
                         opval = opval or 0  # handles None
                         cn_fingerprint.append(wt * opval)
-
-        return cn_fingerprint
+        chem_fingerprint = []
+        if self.chem_info is not None:
+            for val in absdeltas.values():
+                chem_fingerprint.append(val / Nwt)
+        return cn_fingerprint+chem_fingerprint
```
computron: use Pycharm, etc. to get code formatting right (in this case, spaces surrounding the + sign)

matminer/featurizers/site.py
```diff
@@ -481,15 +506,23 @@ def featurize(self, struct, idx):
...
-
+        max_cn = sorted(self.op_types)[-1]
+        for k in range(max_cn):
+            cn = k + 1
+            if cn in list(self.ops.keys()):
+                for op in self.op_types[cn]:
+                    labels.append("{} CN_{}".format(op, cn))
+        if self.chem_info is not None:
+            for prop in self.chem_props:
+                labels.append("cn-wt-weighted av. abs. diff {}".format(prop))
```
computron: this title is too long, it's going to get cut off in the table headers and if you have multiple properties, you will only see "cn-wt-weighted" (and nothing else) for every property.

Maybe "{} local abs. diff" (?)

matminer/featurizers/site.py
```diff
@@ -413,24 +413,32 @@ def from_preset(preset, **kwargs):
...
 
-    def __init__(self, op_types, **kwargs):
+    def __init__(self, op_types, chem_info=None, **kwargs):
         """
         Initialize the CrystalSiteFingerprint. Use the from_preset() function to
         use default params.
         Args:
             op_types (dict): a dict of coordination number (int) to a list of str
                 representing the order parameter types
+            chem_info (dict): a dict of chemical properties (e.g., atomic mass)
```
computron: In practice, I think people are going to have a hard time passing in the chem_info dict and find this to be a pain to use. While flexible, normally you have a dataset of 1000 compounds and don't want to manually construct a dictionary of all possible species to some value like electronegativity.

It might be better if this was able to easily work with an "AbstractData" class instead, since those already have a bunch of built-in dictionaries for various properties. Or, some easy way for the user to leverage those classes rather than construct the dictionary themselves.

Or, maybe instead AbstractData itself could implement a method that turns its properties into a dictionary. That way you could do ``PymatgenData.dict_output("X")`` that will turn all the PymatgenData electronegativities to a dictionary format.

matminer/featurizers/site.py
```diff
@@ -413,24 +413,32 @@ def from_preset(preset, **kwargs):
...
+            return CrystalNNFingerprint(op_types, chem_info=None, **kwargs)
 
         else:
             raise RuntimeError('preset "{}" is not supported in '
                                'CrystalNNFingerprint'.format(preset))
 
-    def __init__(self, op_types, **kwargs):
+    def __init__(self, op_types, chem_info=None, **kwargs):
         """
         Initialize the CrystalSiteFingerprint. Use the from_preset() function to
```
computron: This is not your mistake, but might as well fix it - change ``CrystalSiteFingerprint`` to ``CrystalNNFingerprint``.

computron:
@nisse3000 please see the code review

nisse3000:
@computron, I'm done with your suggestions. The only thing that I've not addressed yet is the dictionary/AbstractData suggestion. Can we raise on issue, and I'll address that later? @shyamd was fine with using dictionaries in the beginning.

computron:
Please go ahead and add the issue about the AbstractData

There is also the part about falling back on element properties -

Here is a common situation:
- We decorate a structure to have oxidation states, e.g., has sites with Mn3+ and O2- species
- We want to use a chemical site property that doesn't have oxidation state dependence - e.g., Pauling electronegativity, which might only be defined for "Mn" and "O"

From my reading of the code, it seems this would trip things up. Either you'd have to remove the oxidation states from the structure as a preprocessing step before doing the analysis (which seems wasteful if you have 50,000 structures) , or you'd have to manually hack your dictionary to try and repeat the Pauling electronegativity for all possible oxidation states of that element (e.g., the dictionary has the same electronegativity value repeated for Mn2+, Mn3+, Mn4+, Mn5+).

matminer/featurizers/site.py
nisse3000: @computron, this is for efficiency. If wt == 0 and you don't have this elif part, you'd still compute the OPs for this CN. I believe this could save some computation time.

nisse3000:
@computron, your last comment should be addressed now. Let me know what you think. I made the element/species dictionaries property-specific to be most robust.

PR #260, opened 2018-06-25
Bugfix of lambdas in GRDF's and AFS's presets in site.py

ldwillia:
Updated lambda functions for GeneralizedRadialDistributionFunction.from_preset and AngularFourierSeries.from_preset to use a local variable instead of grabbing an outer scope variable when defining lambda functions in a loop. Previous behavior caused all bins to be identical and using the final value of 'center' instead of creating a series. Also updated default values in GeneralizedRadialDistributionFunction.from_preset from width=0.5, spacing=0.5 to width=1.0, spacing=1.0 in order to avoid error: "ValueError: Numerical integration does not play well with the chosen bin functionals, choose new ones."

## Summary

* Fixed lambda functions in GRDF and AFS .from_preset methods
* updated default width and spacing in GRDF.from_preset to avoid error
* Added self to implementors list for GRDF and AFS (Is that okay? If not, remove.)

## TODO (if any)

* GRDF.from_preset('histogram') sometimes produces a value of 'inf' for some bins. Cause currently unknown.

### DIFF
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -1494,7 +1494,7 @@ class GeneralizedRadialDistributionFunction(BaseFeaturizer):
                                      'determine the correct feature labels.')
 
     @staticmethod
-    def from_preset(preset, width=0.5, spacing=0.5, cutoff=10, mode='GRDF'):
+    def from_preset(preset, width=1.0, spacing=1.0, cutoff=10, mode='GRDF'):
         '''
         Preset bin functionals for this featurizer. Example use:
             >>> GRDF = GeneralizedRadialDistributionFunction.from_preset('gaussian')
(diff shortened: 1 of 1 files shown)


ldwillia:
I think the fail on the ci/circleci test might be due to updating the bin defaults for GRDF and having half as many bins as expected by the test. I am not certain though.

dyllamt:
You are right about the failure, it's in lines 492-499 of `test_site.py`. I originally created a test to make sure that the correct bins were generated by the `from_presets` method, but since the default bins are changed they no longer match.

If you take out the half-integer bins from the reference array, this pull request should pass the tests.

computron:
All, for now I will merge this and update the tests as it seems an improvement over the current code. However, I do think more work on this class is needed.

PR #263, opened 2018-06-30
minor bugfix: set accuracy to int, None is not accepted.

albalu:
I set it to 12 that is the default for acc_factor in pymatgen's EwaldSummation.

### DIFF
--- a/matminer/featurizers/structure.py
+++ b/matminer/featurizers/structure.py
@@ -1073,7 +1073,7 @@ class EwaldEnergy(BaseFeaturizer):
     Features:
         ewald_energy - Coulomb interaction energy of the structure"""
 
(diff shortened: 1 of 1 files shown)


computron:
@WardLT does this class need an ``accuracy`` parameter at all?

I would guess that the default (12 decimal places) is not really worth fiddling with, have you found otherwise?

computron:
Is 12 the number you'd set as default?

WardLT:
I suspect 12 is unnecessarily accurate. Maybe 4 could be a better default? I use an accuracy of 4 without seeing too much noise in the forces from Ewald summation.

computron:
@albalu I would suggest making the default either 4 or 5 in this PR

computron:
Thanks!

PR #267, opened 2018-07-07
Faster gaussym

WardLT:
## Summary

Accelerates the GaussSymFunc calculation by about 40x. I did this by by caching some results (e.g., distance, angle calculations) and vectorizing a few operations.

I've also accidentally included another, minor commit that handles some edge cases for the YangSolidSolution and AtomicPackingEfficiency classes.

## TODO (if any)

Potential: Cache the "get_neighbors" function.


computron:
Thanks!

PR #268, opened 2018-07-10
featurize_dataframe debug in PartialRadialDistributionFunction

albalu:
## Summary

PartialRadialDistributionFunction featurize method works fine. However, when featurize_dataframe is called, "zeros" will be used which has unequal length to prdf. This was very confusing so I added featurize_dataframe test for this featurizer as well.

dist_bins modified to be the start of each bin as is supposed to be based on the docs. 

@WardLT , could you please check to see the way I resolve this issue is correct?


WardLT:
Nice catch! This looks good to me

computron:
Thanks!

Issue #246, opened 2018-06-06
make the DOS featurizers more "smooth" and less "hard"


computron:
See discussion on:

https://github.com/hackingmaterials/matminer/issues/240#issuecomment-395229082

dyllamt:
I would like to work on this. I can think of two ways that this can be accomplished, but would like others opinions on the two methods.

1. The first method applies to both `DOSFeaturizer` and `BandEdge`. In both, the partial DOS is weighted by an exponentially decaying function. For `BandEdge`, the decay constant will depend on the DOS mass, while the decay constant will be the same for all compounds in `DOSFeaturizer`.
2. The second method would add another dimension to the features generated. In this method you track the contributions to the partial DOS in sub-energy windows. This is an "energy distribution function", analogous to a radial distribution function, but for density of states. You'll still have a hard cutoff where your last bin is, but it's "smoother" since you track the change in DOS composition with energy.

Maybe its best to implement option 2 in `DOSFeaturizer` and option 1 in `BandEdge`?

albalu:
I like the idea of using DOS mass for decay constant by default. I am having a hard time imagining how option 2 works, so we will have more features like cbm_s_0.0-0.05, cbm_s_0.05-0.10 , etc; i.e. contribution in energy windows? Option 1 is simpler.

dyllamt:
Option 2 adds some level of spectral information to the features, but for now, I'll go ahead and implement option 1 along with an entropy-type score in the `DOSFeaturizer`. If we like it, it will be easy to extend these changes to `BandEdge`.

computron:
The "spectral" type approach has been somewhat explored before, see for example:

Isayev, O. et al. Materials Cartography: Representing and Mining Material Space Using Structural and Electronic Fingerprints. Chem. Mater. 141215151057004 (2014). doi:10.1021/cm503507h

It's probably somewhat useful for describing a distance metric between DOSes and might be useful in kernel methods. But I think it would take a little fiddling to make it useful. The first method of a single number is probably good enough to start with

ardunn:
Old issue, was closed by #246

PR #272, opened 2018-07-24
Softening the density of states featurizers

dyllamt:
## Summary

* An exponential decay function was introduced to sample the band edge density of states smoothly.
* This affects both `DOSFeaturizer` and `Hybridization` (formerly `BandEdge`). The name change is consistent with the output of the featurizer.
* The number of significant features in `DOSFeaturizer` was changed to an entropy score to reflect the degree of hybridization at the band edge.
* Updated tests. The orbital scores before and after implementing the exponential decay function were similar, but the precise digits changed.


dyllamt:
Linking to issue #246

dyllamt:
Failing style test is for redefining the `featurize` method in `Hybridization` (formerly `BandEdge`).

I don't think that it affects the use of the featurizer.

albalu:
Hi @dyllamt , yes I had that Codacy issue before too, you can ignore that one. Thank you for implementing this. Regarding the name Hybridization, it makes it clear that we can get hybridization but doesn't cover features such as cbm_p or Pb_d as opposed to cbm_sp (hybridization) that was why I thought about BandEdge or maybe BandEdgeCharacter; regardless, since it is under dos.py it implies that it looks at dos information.

dyllamt:
Hmmm that's a good point that there is more than just hybridization features. Maybe we should break that up into two featurizers?

Then there would be:
1. DosFeaturizer - returns site and orbital projected DOS
2. DosOrbitals - returns orbital projected DOS and hybridizations
3. DosSpecie - returns s/p/d/f DOS for a particular species

albalu:
I think "Hybridization" is better than having two separate classes.

matminer/featurizers/dos.py
```diff
@@ -59,25 +56,30 @@ def featurize(self, dos):
...
                                        self.gaussian_smear)
 
         feat = OrderedDict()
         for ex in ['cbm', 'vbm']:
-            orbscores.sort(key=lambda x: x['{}_score'.format(ex)], reverse=True)
+            orbscores.sort(key=lambda x: x['{}_score'.format(ex)],
+                           reverse=True)
             scores = np.array([s['{}_score'.format(ex)] for s in orbscores])
-            feat['{}_nsignificant'.format(ex)] = len(scores[scores > self.significance_threshold])
+            print(scores)
```
albalu: debug print leftover

dyllamt:
Good catch, removed the print statement from the pull request!

albalu:
@computron , I don't have permission to merge this. Could you please if everything looks good?

computron:
@albalu I made some permission changes, can you see if you can merge now?

albalu:
Yes thank you both!

PR #274, opened 2018-07-24
Update plot.py for hoverinfo

ardunn:
## Summary

Previously, hoveinfo did not work for most of the plots (it was ignored)
Hoverinfo can now be used by setting `hoverinfo` in the constructor of PlotlyFig

* Added hoverinfo specification to bar, parallel_coordinates, heatmap_basic, and heatmap_df
* Regenerated test plot json accordingly


computron:
Thanks!

Issue #273, opened 2018-07-24
Have progress bar during featurization


ardunn:
It might be nice to have a progress bar (which could be turned on/off) during featurization, so you'd have an idea of how long a featurizer will run. `tqdm` seems like a good package to do this

computron:
Agree, we use tqdm to do progress bars for some of the data retrieval already

computron:
Also I guess this will mainly be for featurize_many?

ardunn:
Yeah I imagine featurize_many is the lowest level at which the progress bar would be present. Although it would be good if it worked with the higher level methods (featurize_dataframe, fit_featurize_dataframe, MultiFeaturizer.featurize_dataframe, MultiFeaturizer.fit_featurize_dataframe)

computron:
Sure, I think we are on the same page here

ardunn:
closed via #275

PR #275, opened 2018-07-28
Improvements to basefeaturizer (+pbar)

ardunn:
## Summary

Add a progress bar, fix a few minor BaseFeaturizer issues

####  Progress bar
* Added a `tqdm` progress bar (`pbar`)for single featurizers and multiplefeaturizers
* Works with featurize_many, featurize_dataframe, and fit_featurize_dataframe
* Progress bar enabled by default
* To my knowledge, works with all valid combinations of arguments to all `*_featurize_*` methods

#### Other
* Changed input signatures of featurize_dataframe for MultipleFeaturizer + FunctionFeaturizer to prevent future errors when args/kwargs added to BaseFeaturizer
* Fixed common return_errors bug which occurred when a featurizer returned a numpy array rather than a list
* Moved a print statement routed to stderr to a warning statement
* Reformatted base.py
* Removed `FunctionFeaturizer` need to overload featurize_dataframe

# Example
```
df = load_elastic_tensor().iloc[:100]
bf = BondFractions()
xrd = XRDPowderPattern()
de = DensityFeatures()
mf = MultipleFeaturizer([bf, de, xrd])
mf.fit_featurize_dataframe(df, "structure", multiindex=True, pbar=True)
```

Output:
```
BondFractions: 100%|██████████| 100/100 [00:09<00:00, 10.52it/s]
DensityFeatures: 100%|██████████| 100/100 [00:00<00:00, 628.06it/s]
XRDPowderPattern: 100%|██████████| 100/100 [00:07<00:00, 14.23it/s]
```

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -196,6 +196,20 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
             if col in labels:
                 raise ValueError('"{}" exists in input dataframe'.format(col))
 
+        # Multiindexing doesn't play nice with other options!
+        if multiindex:
+            if inplace:
+                warnings.warn("Multiindexing enabled with inplace=True! The "
+                              "original dataframe index has changed.")
+            if pbar:
+                warnings.warn("Progress bars with multiindexing are not"
+                              "currently supported. Progress bar is being "
+                              "disabled.")
+        elif isinstance(df.columns, pd.MultiIndex):
+            # If input df is multi, but multi not enabled...
+            raise ValueError("Please enable multiindexing to featurize an input"
+                             " dataframe containing a column multiindex.")
+
         # Compute the features
         features = self.featurize_many(df[col_id].values,
                                        ignore_errors=ignore_errors,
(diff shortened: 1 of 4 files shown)


ardunn:
Closes #273

ardunn:
The difference in parameters Codacy mentions is intentional

WardLT:
This looks very cool! 

A comment not completely related to this PR: A call to the `super` of `featurize_dataframe` was removed by a previous commit [link](https://github.com/ardunn/matminer/commit/269fa53885f0fee37f5ba14a980f84dbdfa7bb45), which prevents the caching operations for VoronoiNN from working. If re-add the call to `super`, we loose the nice progress list for which featurizer is being run but get the caching back. But, do you think that's a worthwhile tradeoff? [I'm in favor in putting it back, but am not sure why the call to super was removed and don't know what other problems it was causing]

ardunn:
Yes that makes sense. 

But how many features overload featurize_dataframe? To me it appears only FunctionFeaturizer does it, and (at a glance) it looks like it could be converted over to _not_ overriding featurize_dataframe.

computron:
@ardunn 

You should do  the doctrings properly (PyCharm can help auto-generate them), not just put a comment that says "same things as before". Meaning, each variable should have its own entry in the "Args" key.

PR #276, opened 2018-07-31
Added BOOP site features

WardLT:
## Summary

Add Bond Orientational Ordering Parameters. Used by the feature set of [Seko PRB 2017](http://link.aps.org/doi/10.1103/PhysRevB.95.144110). 

Question: Does this tool already exist in pymatgen? It seems like a few different methods use BOOP parameters, but I couldn't find a general implementation.

## TODO (if any)

- Might benefit from a caching operation (noted in a line comment)


WardLT:
Waiting for new release of pymatgen

computron:
Great, thanks!

PR #277, opened 2018-07-31
Added test for MultiFeaturizer supporting caching

WardLT:
## Summary

This PR adds a unit test for MultiFeaturizer to verify it works with the caching methods.

For now, it will fail. Sorry, I should have made this one earlier.

## TODO (if any)

None.


WardLT:
I made some changes to allow for the caching to work and maintain the multiindex by adding some utility operations for determining which column labels to generate. 

This update will change the status bar for MultiFeaturizer, as it no longer loops over featurizers and then entries. But, I think that is the only behavior change we should notice.

computron:
Thanks! I won't really have a chance to review in detail - feel free to merge the PR yourself if you feel it's ready to go.

WardLT:
I think it's ready to go.

@ardunn , do you want to look it over?

ardunn:
looks ok to me @WardLT

ardunn:
@WardLT this PR partially breaks the progress bar.

```python
df = load_elastic_tensor().iloc[:100]
bf = BondFractions()
bob = BagofBonds()
xrd = XRDPowderPattern()
de = DensityFeatures()
mf = MultipleFeaturizer([bf, de, xrd])
mf.fit_featurize_dataframe(df, "structure", multiindex=True, pbar=True)
print(df)
```

Used to produce:
```
BondFractions: 100%|██████████| 100/100 [00:09<00:00, 10.52it/s]
DensityFeatures: 100%|██████████| 100/100 [00:00<00:00, 628.06it/s]
XRDPowderPattern: 100%|██████████| 100/100 [00:07<00:00, 14.23it/s]
```

Now it produces;
```
MultipleFeaturizer: 100%|██████████| 100/100 [00:06<00:00, 15.55it/s]
```

Not a big deal but it would be good if we can fix it

WardLT:
Unfortunately, I couldn't figure out how to fix the status bars as it directly conflicts with the caching. Multiple progress bars requires the outer loop to be featurizers, and caching requires the outer loop to be entries.

PR #281, opened 2018-08-10
Add seko

WardLT:
## Summary

I'm gradually adding in the features needed to recreate the features from [Seko PRB 2017](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.95.144110). This PR includes the tools for the "elemental representation" portion of the features:

1. New lookup tables for MagpieData. Some are using data taken from PymatgenData, and duplicated in MagpieData for convenience
2. New site featurizer for the elemental properties of each site
3. The ability to compute the covariance between site features
4. The ability to compute skewness and kurtosis of features

## TODO (if any)

None


computron:
A few comments:

- why not use scipy's skew and kurtosis functions?
- I am not sure the cleanest way to do this, but I feel like if someone uses the Seko preset, then the Seko citation should be there. If someone manually sets the features (even if they overlap the Seko preset), it's OK by me if the citation doesn't show up, but if someone is just re-using the feature set that's published in a paper it would be nice to cite that paper.

WardLT:
Thanks for the comments, @computron !

- I do use scipy's skew and kurtosis for unweighted data. Unfortunately, I had to re-implement them for data with weights.
- I like your idea about having citations if somone uses the Seko present, and implemented a simple way for making the citation appear.

PR #288, opened 2018-08-21
Average Bond Length and Bond Angle implementations

atan14:
## Summary

- Calculating mean bond length and bond angle using VoronoiNN class
- Based on De Jong's 2016 paper (https://www.nature.com/articles/srep34256)

## TODO (if any)
- None

### DIFF
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -1991,8 +1991,6 @@ def _iterate_wigner_3j(l):
 
 class AverageBondLength(BaseFeaturizer):
     '''
-    Voronoi based average bond length.
-
     Determines the average bond length between one specific site
     and all its nearest neighbors using one of pymatgen's NearNeighbor
     classes. These nearest neighbor calculators return weights related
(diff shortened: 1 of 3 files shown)


WardLT:
Rui is working with me this summer. I've already looked over this PR, and think it's good to go. 

It also raises a few points that, perhaps, warrant opening an issue:

1. Does SiteStatsFingerprints need a mechanism for associating publications with certain presents? An example of this would be how SiteElementProperty returns a citation for Seko's paper if you use the preset.
2. Do we need a utility module that holds the citations for papers that frequently appear in our featurizers? Those bibtex citations are adding a fair amount of length to each module.

WardLT:
Also, mention #91 (almost fixed!)

computron:
Hi all

Thanks for the PR - for the AverageBondLength and AverageBondAngle, the description in the code seems to say you can use any of pymatgen's NearNeighbor classes but the implementation looks specifically geared towards VoronoiNN.

We have some tests on the various NN algorithms as part of an upcoming publication and I think our team would mostly want to use CrystalNN for various bond length, etc. determination going forward.

I would suggest that AverageBondLength and AverageBondAngle just take in some nearneighbor instance (of any type) and just use that rather than tailor the code to VoronoiNN. As for weights, note that the get_nn_info() method of an arbitrary NearNeighbor class will return both site and weights needed. So you don't need to input the weighting scheme inside the matminer class, the NearNeighbor class already has the information on how to do the weighting.

computron:
@WardLT @atan14 any comment on this PR?

I typically don't like to leave PRs open for too long

computron:
Thanks!

PR #290, opened 2018-08-29
Fix Issues with MultiFeaturizer and Multiple dtypes

### LINKED ISSUE #289: MultipleFeaturizer with mixed dtypes
When output data is mixture of numbers and str (pandas object) MultipleFeaturizer converts all of them to object which is pretty bad. For example, the following test under "test_composition.py" would fail.

```
def test_multifeat_dtypes(self):
    featzers = [
        Stoichiometry(),
        AtomicOrbitals()
    ]
    multifeat = MultipleFeaturizer(featzers)
    df = multifeat.featurize_dataframe(self.df, col_id='composition')
    self.assertEqual(set(df.dtypes.astype(str)), {'float64', 'int64', 'object'})
```

if I override get_dataframe inside MultipleFeaturizer, the problem is fixed (also the progress bars work for each featurizer as opposed to only for MultipleFeaturizer). However, it seems like it fails some caching tests. @WardLT , this seems to be related to your code? Unfortunately, at the moment, I don't have a solution to both these issues:

```
def featurize_dataframe(self, df, col_id, **kwargs):
    multiindex = kwargs.get("multiindex", False)
    for i, featzer in enumerate(self.featurizers):
        if i>0 and multiindex:
            col_id = ('Input Data', col_id)
            multiindex = False
        df = featzer.featurize_dataframe(df, col_id, **kwargs)
    return df
```

WardLT:
## Summary

Should fix #289, where @albalu pointed out that the current implementation of MultiFeaturizer changes the dtypes of features if multiple types are present.

- Adds a unit test that illustrates this problem
- Preserve dtypes by using `dtype=object` when concatenating features

## TODO (if any)

None

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -421,7 +421,8 @@ class MultipleFeaturizer(BaseFeaturizer):
         self.featurizers = featurizers
 
     def featurize(self, *x):
-        return sum([list(f.featurize(*x)) for f in self.featurizers], [])
+        return np.hstack(np.squeeze([np.array(f.featurize(*x), dtype=object)
+                                     for f in self.featurizers]))
 
     def feature_labels(self):
         return sum([f.feature_labels() for f in self.featurizers], [])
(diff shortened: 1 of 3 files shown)


albalu:
great! Thank you @WardLT

PR #291, opened 2018-08-29
Better Support for Pandas for featurize_many

WardLT:
## Summary

Improves compatibility of `featurize_many` with Pandas objects. Fixes two kinds of errors:

- Using `pd.Series` crashes if there is no entry with index of `0`
- Using `pd.DataFrame` crashing if you do not first retrieve the values (e.g., `data.values`)

These errors are likely confusing to newer users, and I often ran into them. 

This PR adds a unit test that fails before the fix, and a fix for the issue.

## TODO (if any)

None

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -286,7 +286,9 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
             return []
 
         # If the featurize function only has a single arg, zip the inputs
-        if not isinstance(entries[0], (tuple, list, np.ndarray)):
+        if isinstance(entries, pd.DataFrame):
+            entries = entries.values
+        elif isinstance(entries, pd.Series) or not isinstance(entries[0], (tuple, list, np.ndarray)):
             entries = zip(entries)
 
         # Add a progress bar
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -255,12 +255,14 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
     def featurize_many(self, entries, ignore_errors=False, return_errors=False,
                        pbar=True):
-        """
-        Featurize a list of entries.
+        """Featurize a list of entries.
+
         If `featurize` takes multiple inputs, supply inputs as a list of tuples.
 
+        Featurize_many supports entries as a list, tuple, numpy array, Pandas Series, or Pandas DataFrame.
+
         Args:
-            entries (list): A list of entries to be featurized.
+            entries (list-like object): A list of entries to be featurized.
             ignore_errors (bool): Returns NaN for entries where exceptions are
                 thrown if True. If False, exceptions are thrown as normal.
             return_errors (bool): If True, returns the feature list as
@@ -278,7 +280,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
                              " return_errors.")
 
         # Check inputs
-        if not hasattr(entries, '__getitem__'):
+        if not isinstance(entries, (tuple, list, np.ndarray, pd.Series, pd.DataFrame)):
             raise Exception("'entries' must be a list-like object")
 
         # Special case: Empty list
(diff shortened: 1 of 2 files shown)


computron:
In general I find that "isinstance" is considered less "Pythonic" than other methods for checking, e.g. 
- try it a certain away and handle exceptions
- check for the interface rather than the class - e.g. if you are actually checking for "a single arg" like the code comments, than the check should be for that directly rather than the class

ardunn:
@WardLT another way to do it without explicitly type checking might be:

```python
try:
    entries = entries.tolist()
except AttributeError:
    pass # or handle?
```

WardLT:
Ok, I might be able to do this without any type checking. But, apparently, that way was not it.

WardLT:
Nevermind, there are some issues involving calling `numpy.array` of a list of pymatgen `Composition` objects that will make this numpy-based approach difficult. I'll be rolling back these commits.

Would you mind going in to more detail on the issues behind `isinstance`? For this case, using `isinstance` seems like a more reliable approach than relying on `try/except`. For example, pd.Series and pd.DataFrame will both throw `KeyError`, but have different resolutions. Also, my issue is that all classes implement `__getitem__`  differently, which complicates checking whether the interface supports an operation. For that reason, it seems like the `isinstance` approach would be cleaner, but I want to understand your point of view better.

computron:
I'd say:

* It's not clear to me that ``featurize_many`` should natively support pd DataFrame or Series in the first place. Maybe it should be expected that the user can follow the API and just throw in "pd.values" to ``featurize.many`` rather than throwing in a DataFrame. What if there is some other matrix-like object but not quite a matrix that people use, do we need to support that too?
* If you really do think it's useful to add pd.DataFrame and pd.Series as supported inputs for featurize_many, the docstring for the function should reflect that.
* If you are adding explicit hacks intended for pd.Dataframe and pd.Series, then I guess it's fine to use isinstance() as in the original solution.
* Finally, there seems to be different ways of checking whether something is "iterable" within a few lines of code. There is the ``hasattr(entries, '__getitem__')`` and also ``isinstance(entries[0], (tuple, list, np.ndarray))``. I don't really like either solution - the first will consider a string as iterable, which is not something you want I think, and the second might not be a complete list. But other than something like "isinstance(entries, collections.Iterable) and not isinstance(entries, str)" I'm not sure what else to do. But at least it should be the same method since I think you are doing the same check.

WardLT:
All good points:

- It's not clear to me that featurize_many should natively support pd DataFrame or Series in the first place. 

I have found I use Pandas as inputs as `featurize_many` somewhat frequently. For all its benefits, employing `featurize_dataframe` results in me needing to keep track of the feature labels. E.g., 

```python
data = f.featurize_dataframe(data, 'composition')
model.fit(data[f.feature_labels()], f['energy`])
```

which is just verbose enough to make it common for me to just get the numpy array with the features

```python
X = f.featurize_many(data['composition'])
model.fit(X, data['energy'])
```

The latter snippet works 90% of the time, except for when there is no entry with an index of `0`. (Hence my motivation for making this PR) 

- What if there is some other matrix-like object but not quite a matrix that people use, do we need to support that too?

I would limit special cases to Pandas, as many components of matminer are designed to be Pandas-friendly.

- If you really do think it's useful to add pd.DataFrame and pd.Series as supported inputs for featurize_many, the docstring for the function should reflect that.

Oops, I'll make that change.

- Finally, there seems to be different ways of checking whether something is "iterable" within a few lines of code. There is the hasattr(entries, '__getitem__') and also isinstance(entries[0], (tuple, list, np.ndarray)). I don't really like either solution - the first will consider a string as iterable, which is not something you want I think, and the second might not be a complete list. But other than something like "isinstance(entries, collections.Iterable) and not isinstance(entries, str)" I'm not sure what else to do. But at least it should be the same method since I think you are doing the same check.

I agree, the type checking in this class is muddled and I'm not sure the best route forward. `Iterable` isn't quite what we need, as sets are Iterable but lack the `__getitem__` needed later on. `dict` satisfies both, but is not suitable for use in `featurize_many` because it is does not have order. Maybe I'll just stick with `isinstance(..., ('list', 'tuple', 'ndarray', 'Series', 'DataFrame'))` and specify those are the accepted "list-like" objects in the docstring.

computron:
Ok sounds good, let's go with this

PR #297, opened 2018-09-11
Add flexible example dataset retrieval and handling

Doppe1g4nger:
## Summary

Added feature to fetch datasets from online storage. Each example dataset currently available in matminer is stored in a project on figshare. Added direct import of these functions from the datasets module so they can be used easily in other projects. Modified behavior of dataset path determination so that users can set a MATMINER_DATA environment variable to look for and place datasets in a user defined path rather than the module directory, default is still module directory.

* The fetch_external_dataset function downloads a file from a given url to a specified path, then ensures that the MD5 hash of the file matches a provided hash that the data should match. The load_* functions have been modified to use these if the dataset can't be found on the local disk
* The datasets \_\_init\_\_ file was modified so users can import necessary external dataset fetching functions directly from matminer.datasets
* Added functionality was made to allow users to set a MATMINER_DATA environment variable that can be used to change where matminer looks for and stores example datasets

* Cosmetic updates: sorted import ordering, updated docstrings, prettified list blocks and spacings, etc.
* Updated tests

## TODO (if any)

* Before dataset downloads will work, the figshare project must be made public so the download attempt doesn't return a 403 error. I have tested the functionality with private download links but did not feel they should be included in the PR. Before making the project public various metadata needs to be filled out for the project and the individual datasets on figshare. I feel I don't know enough about the datasets to handle that entirely unsupervised. @computron I sent you an invite to collaborate on the project, where you should be able to make any edits you feel necessary prior to making the project public.


Doppe1g4nger:
Note that the PR fails the circleCI test due to the aforementioned 403 error problem

computron:
@Doppe1g4nger can you comment on the Codacy tests? Should those be fixed?

I will look at the figshare stuff relatively soon ...

Doppe1g4nger:
@computron for the Codacy tests two issues popped up:

The first seems to be an issue with the way I'm requesting data using urllib. There seems to be a fix on stack overflow I can try but I'm not sure if it will work as it looks like bandit blacklists any data retrieving calls from urllib. If that doesn't work I could switch over to using requests for getting data but that would add another third party dependency. I can try the alternative method but if that doesn't fix the issue I don't think it's worth adding another dependency to the project over.

The second issue is with the way I check the checksum for the data when it's downloaded. I used MD5 hashing since figshare already displays that hash to users on the data download page. I could switch over to a more secure hashing method and use that instead but I don't see much use as it's not being used to store sensitive data, just check that the data hasn't been changed or corrupted.

computron:
@Doppe1g4nger I didn't seem to get the Figshare invite

@ardunn Can you take charge of this PR?
- get added to the Figshare project that Daniel is talking about
- make sure everything looks OK, and especially that things are properly cited
- give it the "OK to pull in" vote

ardunn:
@computron Sure.

ardunn:
@Doppe1g4nger looking at the unittest, I think it is important that we test all datasets being downloaded, not just the elastic tensor.

Also @Doppe1g4nger : Is your intention to remove the csv files from the repo once they are on FigShare?

Doppe1g4nger:
@ardunn agreed, I'll add additional tests. I don't plan to remove the existing csv files initially, but in the next major release (do we use any major.minor.tiny style versioning scheme?) once we're sure they work properly/consistently.

What are your thoughts on the Codacy issues with the use of MD5 and urllib?

ardunn:
@Doppe1g4nger re versioning style: [pretty much](https://github.com/hackingmaterials/matminer/releases)

Since we are coming up on v0.4.2, the next minor release won't be for a while anyway. Unless someone else has qualms about it, I think it is ok to rm the csvs from the repo in this PR.

To clarify my comment about tests: I think we should have two sets of tests. The first set downloads all datasets and checks the integrity of each via hash (you have already done this for elastic tensor). The second set uses the downloaded data to run the previous test suite. 

Re MD5 and urllib: Just do whatever sklearn does for this. Neither are big issues

Doppe1g4nger:
@ardunn alright, I updated the unit tests to first check to see that downloads work, then check against the data being valid. I also removed the csvs from the repo. I switched the hashing method over to SHA256 like sklearn uses and it turns out plotly already depends on the requests library so it's an implicit requirement anyway; I swapped the download method over to that library.

Fingers crossed once the files are made public it should pass CircleCI tests and be ready for use.

Doppe1g4nger:
@ardunn alright I factored out some of the code so now it's just a process of setting up the path and metadata and calling a validate function for each instance. If we REALLY wanted to generalize it we could in the future make a dataset loader base class with subclasses for each dataset. That would change the current functionality though so it aught to be held off on. Will do the test cases next Monday.

ardunn:
@Doppe1g4nger  I actually just read over the issue corresponding to this PR and for the time being we should keep the csv files in the repo. Would you mind adding them back?

computron:
FYI adding and removing many large files really adds to the overhead of the Git repo. I'd like to avoid doing this too many times. If the plan is to switch everything to Figshare with this PR, then it's fine to leave the CSV removed. Just make sure things *work* for *everyone*, e.g., by actually going through all the published examples and making sure they work as intended (all the way through the entire notebook, not just initial steps) without any code modifications, etc.

Doppe1g4nger:
@ardunn ok, I added some extensive test cases to the error conditions. To test everything properly I switched the Figshare urls from the public links to the private links. I will switch these back and deactivate the private links once the Figshare project is made public. 

I went through all the notebooks that use the example datasets and ran them to make sure they worked. The notebooks all ran the same when the files were downloaded or already present on the machine. Some examples had issues running but as far as I can tell they are results of other changes to matminer, not the dataset loading portion. I verified that the downloaded files have identical checksums to the original files currently on the github repo so they haven't changed.

I'm rather confident in the integrity of the download process at this point so I have the csv files left out of the PR for the reasons Anubhav mentioned.

ardunn:
@Doppe1g4nger Ok thank you. For the issues not related to dataset loading, could you raise issues on the matminer_examples repo if they are not already there?

ardunn:
Do I need to make the FigShare project public?

Doppe1g4nger:
@ardunn Will do. The project can't be made public until all the required metadata is filled out for each dataset. If you could look over the metadata and make sure it looks alright that would be great. So far I have something for each field in each dataset save for authors and keywords. Once the project is made public things can still be updated, but it's probably best to make sure things are up to snuff before then. I'm particularly not certain about the license for each dataset since the license for the github page seems to be a BSD licence and Figshare only has MIT, GPL, and CC variants.

ardunn:
@Doppe1g4nger I am not quite sure where the missing metadata is, besides missing DOIs. Regardless, I am not able to edit any of it. Can you change the settings to allow me to edit it?

Doppe1g4nger:
@ardunn Hmmm, I had assumed that adding collaborators to a project would allow them to edit existing files. It seems only "managers" of given files can edit their metadata and the only way I can find to make someone a manager is by being the person that created the file...

 I emailed Figshare to see if there is a way to make you a manager. If that doesn't work you CAN upload new files so we could just have you upload copies that you would be the manager of. I'm not a fan of that as it makes long term management more complicated but it would be a workable solution...

computron:
@Doppe1g4nger @ardunn - is this ready to merge now? If no, can you give me an ETA? Either is OK, just want to know whether I should include this in an upcoming release or wait until a later release to include this change.

Doppe1g4nger:
@computron Alex and I just need to finalize the datasets on figshare and then it will be ready. I moved the datasets over to a shared figshare account - login info is posted on the group Google Drive - and added what I felt was the appropriate metadata if either of you would like to take a look. I'm meeting with @ardunn tomorrow or Wednesday to go over things so it should be ready to merge by Wednesday afternoon.

ardunn:
@computron Do you have a preference for the license these datasets are released under? Figshare allows for Creative Commons, GPL (multiple versions), MIT, etc. but does not have BSD, which is the license currently used for matminer seems to be based on.

computron:
I'd say either MIT or CC

Doppe1g4nger:
@ardunn I don't see them changing any time soon. The only instances where they should change would be if figshare closed down or they made a major overhaul to the way they host data. Both of which would most certainly come with advance warning.

ardunn:
@Doppe1g4nger ok I am going to merge then

Issue #298, opened 2018-09-15
MultipleFeaturizer does not work with featurizers returning matrices


ardunn:
The following code illustrates the problem:

```python
from matminer.datasets.dataframe_loader import load_elastic_tensor
df = load_elastic_tensor()[:5]
featurizers = [sf.GlobalSymmetryFeatures(), sf.CoulombMatrix()]

# Fails
mf = MultipleFeaturizer(featurizers)
mf.fit_featurize_dataframe(df, "structure", ignore_errors=True)

# Works
# for f in featurizers:
#     df = f.featurize_dataframe(df, "structure", ignore_errors=True)
```

The stacktrace is:
```
MultipleFeaturizer: 100%|██████████| 5/5 [00:00<00:00, 320.58it/s]
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/Users/ardunn/alex/lbl/projects/common_env/dev_codes/matminer/matminer/featurizers/base.py", line 442, in featurize_wrapper
    for f in self.featurizers])
  File "/Users/ardunn/alex/lbl/projects/common_env/common_env3/lib/python3.6/site-packages/numpy/core/shape_base.py", line 286, in hstack
    return _nx.concatenate(arrs, 0)
ValueError: all the input arrays must have same number of dimensions
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/ardunn/alex/lbl/projects/matbench/matbench/matbench/featurize.py", line 564, in <module>
    mf.fit_featurize_dataframe(df, "structure", ignore_errors=True)
  File "/Users/ardunn/alex/lbl/projects/common_env/dev_codes/matminer/matminer/featurizers/base.py", line 160, in fit_featurize_dataframe
    **kwargs)
  File "/Users/ardunn/alex/lbl/projects/common_env/dev_codes/matminer/matminer/featurizers/base.py", line 217, in featurize_dataframe
    pbar=pbar)
  File "/Users/ardunn/alex/lbl/projects/common_env/dev_codes/matminer/matminer/featurizers/base.py", line 320, in featurize_many
[...]

Issue #213, opened 2018-04-27
Structure featurizers should return equal-length vector of features


computron:
A condition we would like to maintain for a structure featurizer is that regardless of the number (or order) of sites in the structure, it returns a consistent number and order of features. At the very least, the featurizer should have an *option* (probably default) to operate in this way. Otherwise the featurizer cannot be used as black-box.

The ``GRDF`` and ``AFS`` featurizers were identified as two that do not meet this requirement. The ``CoulombMatrix`` is another.

We explored options such as having ``VectorStats`` (formerly ``PropertyStats``) and ``MatrixStats`` classes that can take n-dimensional arrays of arbitrary dimension (e.g., 1 x 10, 4 x 4, etc) and collapse them to a finite number of features (e.g., trace/N for all N x N matrices). More details in issue #159 .

ardunn:
I would consider the "equal length" versions of `CoulombMatrix` and `SineCoulombMatrix` taken care of by `BagofBonds`. Perhaps we could keep the functionality of these "heterogeneous" featurizers but move them and label them as something other than featurizers?

computron:
Another thing I would like to do:

For a structure with N sites, compute the the fingerprints (of length M) for each site. So now we have an N x M matrix of numbers of the structure.

Personally, I'd like to take the max across the rows (maximum fingerprint value for a single site), which will result in N maximum values for fingerprint. Then I'd like to take the average of those N numbers, yielding a single number for the structure.

Basically, the first step (taking the max of a site fingerprint) corresponds to seeing how "perfect / symmetric" the bonding environment is around a single site. Then, averaging across all sites, you get an average measure of this. For a very ordered / perfect material, this number should be close to 1.0. For a very disordered material, this number will be closer to 0.

But currently there is no way to start with this N x M feature matrix and then reduce it down the way I want.

albalu:
Just FYI: list of structure featurizers that can't be used as black box that I found so far:
CoulombMatrix, SineCoulombMatrix, OrbitalFieldMatrix, MinimumRelativeDistances

ardunn:
these are all fixed  by #634

Issue #159, opened 2018-02-21
Site-based featurizers in the structure module


dyllamt:
Some of the featurizers in the _structure_ module return site-dependent features:
- CoulombMatrix
- SineCoulombMatrix
- MinimumRelativeDistances
- AngularFourierSeries and GeneralizedRadialDistributionFunction (pending in pull request #138) 

Should they be moved to the _site_ module?

There does exist a featurizer `SiteStatsFingerprint` in _structure_ that can generate the features for all sites if the user desires.

dyllamt:
@nisse3000 since you implemented `SiteStatsFingerprint`, do you think this solution would be appropriate for featurizers other than the presets?

@computron In a way, this may also pertain to issue #135 because you would only need to implement the `fit` method in `SiteStatsFingerprint` for some of these site-based featurizers

Depending on what you think, I can make the changes.

computron:
@dyllamt these are structure features, e.g. the Ewald matrix is a function of the entire structure.  Same with the others, read the docstring and notice that they take in just a Structure (not a Structure and a site index, which the site.py featurizers do) as input.

I am not sure exactly what you are suggesting for ``SiteStatsFingerprint``  but it already works the way you said. The main constructor (``__init__``) takes in any site featurizer, not just the preset ones.

dyllamt:
One of the comments Logan made to the GRDF/AFS pull request was that since these two featurizers return a list of features for each site (e.g., in the case of GRDF, RDFs for each site), maybe they should be site featurizers.

This is different behavior than the `RadialDistributionFunction` or `ElectronicRDF`, which sums the respective RDFs from each site.

I realize that most structural features are composed of site features, but there are currently two types of behavior in the structure module: featurizes which sum the feature vectors from each site (resulting in a feature vector) and featurizers which concatenate the feature vectors from each site (resulting in a feature matrix).

dyllamt:
I guess both behaviors could be achieved through the `SiteStatsFingerprint`

WardLT:
Besides the inputs to the functions, I think another good requirement for "structure" featurizers should be that they return the same number of features regardless of the number of sites in the structure.

The featurizers that return a concatenated list of site features are problematic. For one, the number of features changes with the number of sites, which causes problems for many ML algorithms. Also, the order of inputs to the model depend on the order of the sites - meaning that reordering the sites could lead to different inputs/outputs for an ML model. Together, these factors make these features not easily "ML-ready."

[...]

PR #299, opened 2018-09-15
Ability to flatten OrbitalFieldMatrix

ardunn:
## Summary

Adds an option to flatten the OFM of fixed size (either 32x32 or 39x39) to a fixed length vector (1024 or 1521). This might make it easier to input into ML models. The difference is:

**Matrix**
```python
>>> df = load_elastic_tensor()[:2][["structure", "formula"]]
>>>OrbitalFieldMatrix(flatten=False).featurize_dataframe(df)
structure    formula                               orbital field matrix
0  [[0.94814328 2.07280467 2.5112    ] Nb, [5.273...    Nb4CoSi  [[0.5890106964309796, 0.5152901102725697, 0.0,...
1  [[0. 0. 0.] Al, [1.96639263 1.13529553 0.75278...  Al(CoSi)2  [[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,...
```

vs.

**Flat**
```python
>>> df = load_elastic_tensor()[:2][["structure", "formula"]]
>>>OrbitalFieldMatrix(flatten=True).featurize_dataframe(df)
                                           structure    formula  OFM: s^1 - s^1  OFM: s^1 - s^2  OFM: s^1 - p^1  OFM: s^1 - p^2  OFM: s^1 - p^3  OFM: s^1 - p^4  OFM: s^1 - p^5  OFM: s^1 - p^6  \
0  [[0.94814328 2.07280467 2.5112    ] Nb, [5.273...    Nb4CoSi        0.589011         0.51529             0.0        0.244162             0.0             0.0             0.0             0.0   
1  [[0. 0. 0.] Al, [1.96639263 1.13529553 0.75278...  Al(CoSi)2        0.000000         0.00000             0.0        0.000000             0.0             0.0             0.0             0.0 

...

   OFM: f^14 - f^6  OFM: f^14 - f^7  OFM: f^14 - f^8  OFM: f^14 - f^9  OFM: f^14 - f^10  OFM: f^14 - f^11  OFM: f^14 - f^12  OFM: f^14 - f^13  OFM: f^14 - f^14  
0              0.0              0.0              0.0              0.0               0.0               0.0               0.0               0.0               0.0  
1              0.0              0.0              0.0              0.0               0.0               0.0               0.0               0.0               0.0 
```

Also includes some tests to support this flattened case.

### DIFF
--- a/matminer/featurizers/structure.py
+++ b/matminer/featurizers/structure.py
@@ -646,12 +646,14 @@ class OrbitalFieldMatrix(BaseFeaturizer):
 
     Args:
         period_tag (bool): In the original OFM, an element is represented
-                by a vector of length 32, where each element is 1 or 0,
-                which represents the valence subshell of the element.
-                With period_tag=True, the vector size is increased
-                to 39, where the 7 extra elements represent the period
-                of the element. Note lanthanides are treated as period 6,
-                actinides as period 7. Default False as in the original paper.
+            by a vector of length 32, where each element is 1 or 0,
+            which represents the valence subshell of the element.
+            With period_tag=True, the vector size is increased
+            to 39, where the 7 extra elements represent the period
+            of the element. Note lanthanides are treated as period 6,
+            actinides as period 7. Default False as in the original paper.
+        flatten (bool): Flatten the avg OFM to a 1024-vector (if period_tag
+            False) or a 1521-vector (if period_tag=True).
 
     ...attribute:: size
         Either 32 or 39, the size of the vectors used to describe elements.
@@ -660,7 +662,7 @@ class OrbitalFieldMatrix(BaseFeaturizer):
         `Pham et al. _Sci Tech Adv Mat_. 2017 <http://dx.doi.org/10.1080/14686996.2017.1378060>_`
     """
 
-    def __init__(self, period_tag=False):
+    def __init__(self, period_tag=False, flatten=False):
         """Initialize the featurizer
 
         Args:
@@ -682,6 +684,7 @@ class OrbitalFieldMatrix(BaseFeaturizer):
             my_ohvs[Z] = self.get_ohv(el, period_tag)
             my_ohvs[Z] = np.matrix(my_ohvs[Z])
         self.ohvs = my_ohvs
+        self.flatten = flatten
 
     def get_ohv(self, sp, period_tag):
         """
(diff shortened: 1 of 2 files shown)


ardunn:
@WardLT IIRC CoulombMatrix produces matrices of varying sizes with specific (and varying) meanings for the labels -- ie the meaning of the [0, 1] element in Coulomb Matrix A might mean Li-O while the [0, 1] element in Coulomb Matrix B means N-Cu or something. This makes it hard to do an easy "flatten" operation.

The reason it was easy to do it for OFM was that (1) all OFMs for a given dataframe have the same size and (2) since structures have averaged OFMs, the meaning of [0, 0] element in OFM A (the s^1-s^1 average over the entire structure) is essentially the same thing as the [0, 0] element in OFM B.

Luckily, we already have the BagofBonds class which essentially takes care of this problem, with one caveat -- it produces a lot of features. It uses CoulombMatrix or SineCoulombMatrix as input, fits to a dataframe by identifying all possible labels, and then gives a fixed length vector for each structure.

ardunn:
@WardLT However, it would be nice to add some flattened forms for the remaining featurizers which can only produce matrices or non-ML friendly formats, which at the moment are:

- `MinimumRelativeDifferences` (gives a list of varying length for 1 feature)
- `RadialDistributionFunction` (gives a dict w/ 2 keys, and array values that vary in length)
- `ElectronicRadialDistributionFunction` (same as as `RadialDistributionFunction`)

I am not sure the best way to go about "flattening" these while preserving their physical meanings. Flattening a featurizers output is not good if the output features are uninformative. 

Any ideas?

WardLT:
@ardunn , these are all good points.

Maybe for the two CM classes, we could add functionality to do the "compute eigenvalues and pad with zeros" operation used in many of the CM papers. Determining the length of the vector would require adding in a fitting function to determine the maximum unit cell size or having that as a class attribute. Maybe that is a little too much to confuse / derail this PR, but I think I'd be worth adding.

Those three classes you mention might need some more attention. I'm not sure the best route for moving forward on the MinimumRelativeDistances. We could shape the outputs of the two RDF classes to look more like the PRDF and GRDF, which both produce flattened arrays.

My vote is to go ahead and merge this PR, and I can open issues for these points later (if you agree they're worth doing).

computron:
There have been multiple previous issues on this flattening:

https://github.com/hackingmaterials/matminer/issues/159
https://github.com/hackingmaterials/matminer/issues/213
https://github.com/hackingmaterials/matminer/issues/298

Earlier, we thought of having a general way to turn matrices into equal length vectors using the stats package, and then having the various featurizers that return matrixes reuse these methods.

Any thoughts on this? Would this be separate from the current PR?

ardunn:
I think that would be separate from this PR. Flattening OFM is a bit of an easy special case since equal sized matrices are returned for each structure, which is not true of CM and SCM. So it was easy enough to retain all 1024 elements of the matrix just by flattening it, as opposed to taking eigenvalues and 0-padding etc.

Although I see no reason why we couldn't also separately apply whatever flattening method we use for CM and SCM (e.g.,  eigenvalues as @WardLT suggested) to OFM and then provide an option for generating whichever flattened feature set is desired (the full 1024-vector vs. just the eigenvalues).

computron:
I will merge for now, but still think we should separate out flattening methods (even simple flattening to take the first N elements of a matrix to a vector)

WardLT:
An argument in favor of keeping flattening methods as part of classes: maybe it makes sense to do so for representations that have fixed sizes and where each scalar in the matrix is useful without the other scalars. OFM would fall under this category, but - for the reasons @ardunn pointed out - CM would not. In the cases where the scalars are meaningful, it is nice to be able to generate feature labels for each of them, which requires a modification of the Featurizer. Perhaps separating out the flattening methods from the featurizer is most useful for cases where features are not individually meaningful.

Issue #301, opened 2018-09-17
AtomicOrbitals Fails for Fractional Compositions


WardLT:
AtomicOrbitals gives an `IndexError` when you provide it a fractional composition (example below). Is this a bug, or me not understanding how AtomicOrbitals works? 

If it is not the result of some unit cell dependence (e.g., "NaCl" is different than "Na0.5Cl0.5"), we should make this class return the same value for any compositions with the same element fractions.

```
>>> AtomicOrbitals().featurize(Composition('NaCl'))
['p', 'Cl', -0.32038, 's', 'Na', -0.103415, 0.216965]
>>> AtomicOrbitals().featurize(Composition('Na0.5Cl0.5'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "c:\users\ward6\documents\code\python\matminer\matminer\featurizers\composition.py", line 322, in featurize
    homo_lumo = MolecularOrbitals(string_comp).band_edges
  File "c:\users\ward6\documents\code\python\pymatgen\pymatgen\core\molecular_orbitals.py", line 60, in __init__
    self.band_edges = self.obtain_band_edges()
  File "c:\users\ward6\documents\code\python\pymatgen\pymatgen\core\molecular_orbitals.py", line 102, in obtain_band_edges
    homo = partial_filled[-1]
IndexError: list index out of range
```

computron:
The method should give you the same result whether you put in NaCl or Na0.5Cl0.5

@dyllamt can you comment?

Also, when fixed, a quick unit test should be added to make sure this feature doesn't break in the future

dyllamt:
It looks like fractional compositions were not supported for this featurizer at all, because the underlying Pymatgen class `MolecularOrbitals` assumes that your composition contains integer numbers of atoms (i.e. Na2Cl2 would give the same result as NaCl, but Na0.5Cl0.5 is not supported at all).

I added `get_integer_formula_and_factor` to the beginning of the `featurize` function to sanitize the input of the featurizer. This ensures that the composition passed to `MolecularOrbitals` has integer subscripts.

I also added a test to ensure that two different fractional compositions (NaCl v.s. Na0.5Cl0.5) return the same features. 

Pull request #303

WardLT:
Fixed by #303

PR #303, opened 2018-09-18
fractional compositions in AtomicOrbitals

dyllamt:
## Summary

- add "sanitation" step to `AtomicOrbitals` composition featurizer
- fractional compositions (Na0.5Cl0.5) are now supported
- added test to ensure equivalent compositions (NaCl v.s. Na0.5Cl0.5) return the same features

## TODO
- possibly update pymatgen.core.MolecularOrbitals to require integer subscripts

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -317,9 +317,9 @@ class AtomicOrbitals(BaseFeaturizer):
                 the estimated bandgap from HOMO and LUMO energeis
         """
 
-        string_comp = comp.reduced_formula
+        integer_comp, factor = comp.get_integer_formula_and_factor()
 
-        homo_lumo = MolecularOrbitals(string_comp).band_edges
+        homo_lumo = MolecularOrbitals(integer_comp).band_edges
 
         feat = collections.OrderedDict()
         for edge in ['HOMO', 'LUMO']:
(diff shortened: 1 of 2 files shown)


dyllamt:
## Update
pymatgen's `MolecularOrbitals` were updated to raise a `ValueError` if non-integer subscripts are passed (`AtomicOrbitals` calls this pymatgen class to generate features). The sanitation step of this featurizer ensures that only integer value subscripts are passed to pymatgen.

WardLT:
Thanks @dyllamt !

This looks good to me. Anyone want to review before I merge? I have a user who wants to use this feature.

PS. Adding that this Fixes #301

computron:
The default max_denominator for ``get_integer_formula_and_factor`` is 10,000. @dyllamt does that sounds reasonable? Or do you want to make it something like 1000. I don't know how well the code performs with for example a few thousand atoms in the Composition.

dyllamt:
The performance is not much of an issue (~7ms per featurize call with 10,000 atoms), but the discussion does raise an issue that using `get_integer_formula_and_factor` will truncate a composition when there is a **very** dilute element (e.g. FeC0.00001 --> Fe).

I think this is ok behavior from a scientific standpoint, since if the element is that dilute, it is not contributing significant orbital character. But it seems like the user should still be aware of this behavior, so I added a warning in the docstring and a warning message in the code.

## Summary
* dilute compositions (FeC0.00001) are truncated (becomes Fe) by `get_integer_formula_and_factor`
* updated docstring to alert user to this behavior
* added warning message for when this occurs (using `warnings.warn`)
* added test for the warning message

computron:
Thanks!

Issue #302, opened 2018-09-18
structure_to_oxidstructure should be parallelised 


utf:
The `matminer.utils.conversions.structure_to_oxidstructure` function takes a very long time when processing many structures. Currently the implementation runs only on a single thread. A quick and easy way to improve performance would be to introduce an `n_jobs` argument that enables parallelisation. Similar to the behavior of `n_jobs` in scikit-learn, providing `n_jobs=-1` would set the number of jobs equal to the number of processors.

computron:
The easiest way to do this would probably to make this into a featurizer, which would automatically support parallelism. Similar for some of the other conversion functions

e.g. we can have a module of "conversion" (or other name) featurizers that do these kinds of transformations, the only thing is that the "in_place" option would probably no longer be supported.

computron:
(It would also allow using these conversion tools as part of a Pipeline)

utf:
Seems like a nice idea. Naively, looking through the list of functions in `matminer.utils.conversions`, it looks like every function in there *could* be converted into a featurizer. If anyone has any objections as to why this might be a bad idea I'd like to hear their thoughts. Otherwise, I'll go ahead and refactor them all.

WardLT:
I support making all of these based on the Featurizer API. Is the idea to keep in the `utils.conversions` module?

WardLT:
:+1: I think it's reasonable to put them in a special module under `featurizers.conversions`, as long we document clearly that they do not produce ML-ready features but are instead utilities.

computron:
I think you can just call it ``{}_object``

I would say it shouldn't overwrite anything in the dataframe by default, that should be optional if there at all

PR #305, opened 2018-09-20
Implement conversion functions as Featurizers

utf:
## Summary

This pull request reimplements conversion utilities as Featurizers, allowing them to take advantage of the parallelization in Scikit-Learn. This pull request address #302.

The new conversion Featurizers support all the functionality of the old utility functions, including the ability for "in place" conversion of data. A new `ConversionFeaturizer` abstract class was created which handles this process. The docstring for this class provides the best explanation for how the functionality was implemented:

```python
 """Abstract class to perform data conversions.

    Featurizers subclassing this class do not produce machine learning-ready
    features but instead are used to pre-process data. As Featurizers,
    the conversion process can take advantage of the parallelisation implemented
    in ScikitLearn.

    Note that `feature_labels` are set dynamically and may depend on the column
    id of the data being featurized. As such, `feature_labels` may differ
    before and after featurization.

    ConversionFeaturizers differ from other Featurizers in that the user can
    can specify the column in which to write the converted data. The output
    column is controlled through `target_col_id`. ConversionFeaturizers also
    have the ability to overwrite data in existing columns. This is
    controlled by the `overwrite_data` option. "in place" conversion of data can
    be achieved by setting `target_col_id=None` and `overwrite_data=True`. See
    the docstring below for more details.

    ConversionFeaturizers do not currently support multi-indexing.

    Args:
        target_col_id (str or None): The column in which the converted data will
            be written. If the column already exists then an error will be
            thrown unless `overwrite_data` is set to `True`. If `target_col_id`
            begins with an underscore the data will be written to the column:
            `"{}_{}".format(col_id, target_col_id[1:])`, where `col_id` is the
            column being featurized. If `target_col_id` is set to None then
            the data will be written "in place" to the `col_id` column (this
            will only work if `overwrite_data=True`).
        overwrite_data (bool): Overwrite any data in `target_col_id` if it
            exists.
    """
```

## Example usage

The new conversion featurizers can be used as follows:

```python
from pandas import DataFrame
from pymatgen import Structure, Lattice

coords = [[0, 0, 0], [0.75, 0.5, 0.75]]
lattice = Lattice([[3.8401979337, 0.00, 0.00],
                   [1.9200989668, 3.3257101909, 0.00],
                   [0.00, -2.2171384943, 3.1355090603]])
struct = Structure(lattice, ["Si"] * 2, coords)

df = DataFrame(data={'structure': [struct]})

# Simple conversion; converted data written to df['composition']
df = StructureToComposition().featurize_dataframe(df, 'structure')

# Write converted data to df['composition_red"]
stc = StructureToComposition(reduce=True,
                             target_col_id='composition_red')
df = stc.featurize_dataframe(df, 'structure')

# Convert data in place (eg. converted data written to df['structure']
stc = StructureToComposition(target_col_id=None, 
                             overwrite_data=True)
df = stc.featurize_dataframe(df, 'structure')
```

## List of changes

- Refactor conversions as Featurizers.
- Added deprecation warnings to `matminer.utils.conversions`.
- Added tests for new conversion Featurizers.
- Fixed bug to make `PlotlyFig` histogram plot compatible with Plotly v3
- Tidied and fixed typos in some docstrings.

## ~TODO~

~Currently the implementation does not support multi-indexing as sometimes the column id is changed dynamically. If a user tries to use multi-indexing with the conversion featurizers an error is thrown. From what I can tell, the old implementation also didn't support multi-indexing, so this shouldn't be an issue.~


ardunn:
Awesome! Is there any way we could get multiindexing to work with ConversionFeaturizer? I am not sure about this but you might be able to use the `homogenize_multiindex` util function to help with this

utf:
@ardunn An issue with multi-indexing is that the new column id can be automatically generated.

For example in the case of `DictToObject`: if the column id of the data being converted is `dict_formula`, the converted data will be written to `dict_formula_object` (by default, this can be changed).

If you have a multi index then this problem is not well defined. For example, if your data is in `('dict_data', 'structure')`, should the converted data be written to `('dict_data_object', 'structure')` or `('dict_data', 'structure_object')`?

I'm tempted to make it so that it's the deepest level of the multi-index that is changed. If you think it should be the other way round let me know.

utf:
I've implemented multi-indexing support (w. tests) so this should be good to go.

PR #306, opened 2018-09-21
Add serialization of Pandas DataFrames to JSON

utf:
## Summary

Add functions to serialize Pandas dataframes as JSON. Pymatgen objects within dataframes are serialized and decoded using monty. There is additional support for loading and saving compressed json files to save space.

While not as quick as serialization using `pickle`, the json files produced by these methods can be transferred across different computing platforms, whereas `.p` files are architecture specific.

## Example usage

```python
import pandas as pd
from pymatgen import Structure, Lattice
from matminer.utils.io import load_dataframe_from_json, store_dataframe_as_json

diamond = Structure(
    Lattice([[2.189, 0, 1.264], [0.73, 2.064, 1.264], [0, 0, 2.528]]),
    ["C0+", "C0+"], [[2.554, 1.806, 4.423], [0.365, 0.258, 0.632]],
    validate_proximity=False,
    to_unit_cell=False, coords_are_cartesian=True,
    site_properties=None
)
df = pd.DataFrame(data={'structure': [diamond]})

store_dataframe_as_json(df, 'dataframe.json.gz', compression='gz')
df_new = load_dataframe_from_json('dataframe.json.gz')
```


WardLT:
Thanks! I think this will be a useful feature. 

Noting your comment about integer indices, have you considered using `orient=='list'` to get around that problem? You might need to take the transpose of the dataframe when loading it back in, but maybe it will avoid this problems.

utf:
I'll give that a go. `DataFrame.from_dict()` also has an `orient` option so maybe they will play nicely.

Thanks for the heads-up.

utf:
I'm not too familiar with CSV as it just reminds me of Excel, however for starters I don't think CSV will deal with serialization of pymatgen objects well.

I just tested it and it doesn't look good.

```python
In [1]: import pandas as pd
   ...: from pymatgen import Structure, Lattice
   ...: from matminer.utils.io import load_dataframe_from_json, store_dataframe_as_json
   ...:
   ...: diamond = Structure(
   ...:     Lattice([[2.189, 0, 1.264], [0.73, 2.064, 1.264], [0, 0, 2.528]]),
   ...:     ["C0+", "C0+"], [[2.554, 1.806, 4.423], [0.365, 0.258, 0.632]],
   ...:     validate_proximity=False,
   ...:     to_unit_cell=False, coords_are_cartesian=True,
   ...:     site_properties=None
   ...: )
   ...: df = pd.DataFrame(data={'structure': [diamond]})
   ...:
   
In [2]: df
Out[2]:
                                           structure
0  [[2.554, 1.806, 4.423] C0+, [0.365, 0.258, 0.6...


In [3]: df.to_csv('test.csv')

In [4]: df2 = pd.DataFrame.from_csv('test.csv')


In [5]: df2
Out[5]:
                                           structure
0  Full Formula (C2)\nReduced Formula: C\nabc   :...
```

As `monty` is designed to work with json, it seems to make sense to use that.

PR #309, opened 2018-09-26
Add chunksize support to BaseFeaturizer

utf:
## Summary

`multiprocessing.Pool.map` supports the `chunksize` argument which determines how many entries are given to each processor simultaneously. For lightweight computational tasks, the overhead associated with passing data from `map` to the function being parallelised can dramatically increase the time taken for all tasks to be completed. By using the `chunksize` argument, the overhead associated with passing data to the tasks can be reduced. Note that there is only an advantage to using `chunksize` when the time take to pass the data from map to the function call is within several orders of magnitude to the time taken for the function call itself.

For the heavy featurizers such as `MinimumPackingEfficiency`, the chunksize should be set to 1 for maximum efficiency. However, for lightweight featurizers, such as some of the conversion featurizers, it is generally beneficial to use a `chunksize` greater than 1.

I've implemented support for `chunksize` in `BaseFeaturizer` and set a `chunksize` of 20 in the lightweight conversion featurizers. In my testing this was an optimal value and increased conversion times by 50-70%.

If anyone has a feeling for other featurizers that are very lightweight, please let me know and I can have a look at tuning chunksize for them.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -104,14 +104,16 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
     itself. For computationally expensive featurizers, the default
     chunksize of 1 will be the most efficient. However, for more lightweight
     featurizers, it is recommended that the implementor trial a range of
-    chunksize values to find the optimum.
+    chunksize values to find the optimum. As a general rule of thumb, if the
+    featurize function takes 0.1 seconds or less, a chunksize of around 30 will
+    perform best. For longer featurize times, a chunksize of 1 should be used.
 
     ## Documenting a BaseFeaturizer
 
     The class documentation for each featurizer must contain a description of
     the options and the features that will be computed. The options of the class
-     must all be defined in the `__init__` function of the class, and we
-     recommend documenting them using the
+    must all be defined in the `__init__` function of the class, and we
+    recommend documenting them using the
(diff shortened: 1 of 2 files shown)


WardLT:
Nice! Should we add a `set_chunk_size` function (kind of like `set_n_jobs`) to make it apparent that users can (and should!) set this parameter?

utf:
Good idea, I'll add that in. I think the ideal situation is that we will have sensible defaults for chunksize, which the user shouldn't have to worry about. However, at the end of the day, the ideal chunksize will depend somewhat on the dataset that you are featurizing, so it should be possible to change it.

ardunn:
We should also add to the BaseFeaturizer documentation that a sensible default should be chosen for the chunksize for any new featurizer added

computron:
Thanks!

Can you make the documentation a bit more specific for people implementing featurizers? i.e., people will likely know how long a featurization might take per entry but won't necessarily know how to translate this into a chunksize.

e.g., if it takes 1 second to do featurization per entry, what should I set the chunksize to be? How about if it takes 0.01 seconds? Some kind of mapping between featurization time and chunksize, even if very approximate, would help.

computron:
@utf FYI, just waiting for a little additional documentation before merging

Just some rough guideline is all I'm looking for. At what point should people start considering setting the chunksize to be >1?

Issue #307, opened 2018-09-26
Use seaborn model of dataset loading


utf:
Currently datasets are loaded by calling a specific function. E.g.:

```python
from matminer.datasets.dataframe_loader import load_elastic_tensor
df = load_elastic_tensor()
```

If we want to have multiple elastic datasets (e.g. one generated from Materials Project data and one from Citrine), this will quickly become messy. Instead, it would be nice to use the Seaborn model of loading datasets. E.g.:

```python
import seaborn as sns
iris = sns.load_dataset('iris')
```

In our case, we can use specific strings to identify datasets. E.g. `'elastic_tensor_Ward_2015'`, vs `'elastic_tensor_Jain_2018'`. We should make it clear what datasets are available both in the documentation and via the API (e.g. a `matminer.datasets.available_datasets` list).

Currently, the implementation in matminer performs dataset specific pre-processing to convert from a csv file to the Pandas dataframe. Instead, we should use the json serialization features to store the processed datasets, which can then be loaded without need for further adjustment. We should also use the compression features to minimize the dataset size.

It makes sense that this feature be implemented after #297. After chatting with @Doppe1g4nger, we think the easiest way to implement this would be a `matminer.datasets` module level dictionary, containing something like:

```python
datasets_data = {
	'elastic_tensor_Ward_2018': {
		'url': 'url_of_dataset',
		'citation': 'bibliography_data'
		'metadata': ['list of optional database columns']
	}
}
```

ardunn:
I agree with this! Converting between csv and inflated dataframe each time a dataset is loaded is not ideal.

computron:
Some things I like about CSV

- you can include a comment at the top to give file provenance (sometimes nice to have this in the file itself)
- people can load it easily into other software they have (Excel, various other statistics software, etc)

That said, neither of these are particularly strong reasons to stick with CSV. So I'd say go with whatever you guys decide

computron:
Also the string data set loader sounds good to me (better than individual functions for each data set)

Doppe1g4nger:
I think it would make sense to implement this in two stages. First with the generic load_dataset function using the current csv files, then later we can consider switching over to JSON. 

[...]

PR #312, opened 2018-10-04
Adding seaborn style data loaders

Doppe1g4nger:
## Summary

Added seaborn style dataset loading to matminer, as brought up in issue #307. Where the old style looked like this:
```python
df = load_flla()
```
The new style looks like
```
df = load_dataset('flla')
```
This adds significant flexibility for dataset management and ensures a consistent interface to get each dataset.

* Added seaborn style dataset loading and an available_datasets function
* Deprecated old load_x style functions
* Changed imported functions in module init to load_dataset and available_datasets

## TODO (if any)
Nothing needed for the update to work. In future releases I would like to add more functions to interface with the dataset metadata, remove the old load_x functions entirely once the examples have been updated, and move the dataset dictionary to something like a JSON file with some code to programmatically access and update it, as adding 20+ dataset metadata entries is unsustainable in the Python code directly.

### DIFF
--- a/matminer/datasets/__init__.py
+++ b/matminer/datasets/__init__.py
@@ -1,1 +1,1 @@
-from .dataframe_loader import RemoteFileMetadata, validate_dataset
+from .dataframe_loader import load_dataset, available_datasets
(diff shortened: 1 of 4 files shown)


ardunn:
Thanks daniel! Looks good (sans failing CircleCI)

Eventually we will replace these csvs with JSON, correct? Maybe we should include that in this PR as well

Also, re: deprecation, is there any advantage of warnings.warn over plain ol @deprecated decorator? AFAIK @deprecated is fairly standard practice

ardunn:
`from monty.dev import deprecated`

computron:
Thanks!

PR #316, opened 2018-10-09
Switch datasets over to JSON

Doppe1g4nger:
## Summary

Moved datasets over to JSON (See here: https://figshare.com/projects/Matminer_Example_Datasets/38906) and adjusted code to reflect that. This addresses issue #316 

* Moved datasets over to JSON
* Added "prep for figshare" script that will allow for easy conversion of future datasets
* Extensively refactored functions and tests


utf:
I think moving the script out of the package directory structure is probably best. Pymatgen has a `dev_scripts` folder in the root project directory: https://github.com/materialsproject/pymatgen/tree/master/dev_scripts

Maybe we can just copy that?

Issue #317, opened 2018-10-10
plot.py PlotlyFig.xy() throws an erroneous warning if the arguments are string identifiers


Doppe1g4nger:
I noticed when updating the example notebooks that the plotting class throws a length mismatch warning if you try to plot data from a dataframe with a tuple of column identifiers as such:

```python
pf = PlotlyFig(df, title='Elastic data', mode='offline', 
               x_scale='log', y_scale='log')
pf.xy(('poisson_ratio', 'elastic_anisotropy'), labels='formula')
```
The code erroneously checks the length of the titles against each other rather than the length of the data itself.

Currently I'm working around these in the examples by just passing the data itself rather than the keys, like so:
```python
pf.xy((df['poisson_ratio'], df['elastic_anisotropy']), labels='formula')
```

It should be a fairly easy fix but I'm accumulating side tasks very quickly and don't want this to become a time vampire so just raising the issue for now. If I start in on plotting fixes it'll probably turn into a whole refactoring project for me haha.

albalu:
thanks @Doppe1g4nger  for reporting this. I will work on fixing it

PR #318, opened 2018-10-10
fixed a bug in xy where the length of label was checked instead of data

albalu:
This addresses #317 . I kept the check and throw a warning explicitly as plotly would plot (w/o warning) even if x and y have different lengths.


ardunn:
@albalu [tip](https://help.github.com/articles/closing-issues-using-keywords/): you can type "Closes #317" to automatically close issues when the fixing PR is merged (works most of the time)

PR #319, opened 2018-10-10
Rewrite MultipleFeaturizer

utf:
## Summary

Currently `MultipleFeaturizer` suffers from a number of issues, including:

- Does not show progress bars for individual featurizers.
- Does not work with featurizers returning matrices, as detailed in #298.

I have reimplemented `MultipleFeaturizer` to address all of the above. The implementation now calls the `featurize_dataframe()` method for each featurizer in turn and updates the dataframe sequentially. This might be slightly less efficient than the previous implementation but the overhead associated with creating a new dataframe each time should be relatively minor.

In general I think the new simplicity and benefits warrant the minor performance hit.

I've also updated the tests to catch the issue raised in #298.

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -430,8 +430,8 @@ class MultipleFeaturizer(BaseFeaturizer):
         iterate_over_entries (bool): Whether to iterate over the entries or
             featurizers. Iterating over entries will enable increased caching
             but will only display a single progress bar for all featurizers.
-            If set to False, iteration will be performed over the featurizers,
-            resulting in reducing caching but individual progress bars for each
+            If set to False, iteration will be performed over featurizers,
+            resulting in reduced caching but individual progress bars for each
             featurizer.
     """
 
@@ -440,22 +440,18 @@ class MultipleFeaturizer(BaseFeaturizer):
         self.iterate_over_entries = iterate_over_entries
 
     def featurize(self, *x):
-        """See base class."""
         return [feature for f in self.featurizers
                 for feature in f.featurize(*x)]
 
     def feature_labels(self):
-        """See base class."""
         return sum([f.feature_labels() for f in self.featurizers], [])
 
     def fit(self, X, y=None, **fit_kwargs):
-        """See base class."""
         for f in self.featurizers:
             f.fit(X, y, **fit_kwargs)
         return self
 
     def featurize_many(self, entries, **kwargs):
-        """See base class."""
         if self.iterate_over_entries:
             return super(MultipleFeaturizer,
                          self).featurize_many(entries, **kwargs)
@@ -469,14 +465,13 @@ class MultipleFeaturizer(BaseFeaturizer):
             return [feature for f in self.featurizers
                     for feature in f.featurize_wrapper(x, **kwargs)]
         else:
-            return super(MultipleFeaturizer, self).featurize_many(x, **kwargs)
+            return super(MultipleFeaturizer, self).featurize_wrapper(
+                x, **kwargs)
 
     def citations(self):
-        """See base class."""
         return list(set(sum([f.citations() for f in self.featurizers], [])))
 
     def implementors(self):
-        """See base class."""
         return list(set(sum([f.implementors() for f in self.featurizers], [])))
 
     def _generate_column_labels(self, multiindex, return_errors):
@@ -484,7 +479,6 @@ class MultipleFeaturizer(BaseFeaturizer):
                           for f in self.featurizers])
 
     def set_n_jobs(self, n_jobs):
-        """See base class."""
         super(MultipleFeaturizer, self).set_n_jobs(n_jobs)
         for featurizer in self.featurizers:
             featurizer.set_n_jobs(n_jobs)
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -427,10 +427,17 @@ class MultipleFeaturizer(BaseFeaturizer):
 
     Args:
         featurizers (list of BaseFeaturizer): A list of featurizers to run.
+        iterate_over_entries (bool): Whether to iterate over the entries or
+            featurizers. Iterating over entries will enable increased caching
+            but will only display a single progress bar for all featurizers.
+            If set to False, iteration will be performed over the featurizers,
+            resulting in reducing caching but individual progress bars for each
+            featurizer.
     """
 
-    def __init__(self, featurizers):
+    def __init__(self, featurizers, iterate_over_entries=True):
         self.featurizers = featurizers
+        self.iterate_over_entries = iterate_over_entries
 
     def featurize(self, *x):
         """See base class."""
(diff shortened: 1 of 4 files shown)


utf:
Hmm, so there is a test for MultiFeaturizer which says:

> test whether MutliFeaturizer can handle featurizers that have both 1D vectors with singleton dimensions (e.g., shape==(4,1)) and those without (e.g., shape==(4,))

https://github.com/hackingmaterials/matminer/blob/27bc8afb54307b0f2490ace1aeb8978eb08f91e5/matminer/featurizers/tests/test_base.py#L66

Git blame says @montoyjh made this test. I'm wondering what this test is for, as normal featurizers don't have this requirement.

utf:
I also see my implementation breaks caching across multiple featurizers.

To clarify, I mean that if one featurizer uses `get_nearest_neighbours` and another featurizer also calls the same method, then the nearest neightbours are only calculated once. This is separate from caching within a single featurizer, e.g. when `get_nearest_neighbours` is called twice in the same featurizer.

Caching within a single featurizer is not affected.

There are two options here:
- Have individual progress bars for each featurizer, but no caching across multiple featurizers
- Have a single progress bar for the multi-featurizer as a whole and caching across multiple featurizers.

If people would prefer the second option, then I can try and refactor the current implementation to fix #298 and show the progress bar.

ardunn:
This is very similar to the way we originally had MultipleFeaturizer. I think @WardLT wanted the ability to cache nearest neighbor results though (e.g., #277), which gave rise to a few of these issues. Does this implementation still work with caching? The best solution would be one that fixes all these issues and still works with caching.


Perhaps what we can do is have two separate classes for this:

- MultiFeaturizer: Iterates over featurizers (breaks caching, works with pbars+matrices)
- CachedMultiFeaturizer: Iterates over materials (works with caching, breaks pbars+matrices)

Would be good to get @WardLT opinion...

Also can we refactor MultipleFeaturizer ---> MultiFeaturizer? Easier to remember

utf:
I've refactored my implementation slightly to override `featurize_many` instead. This still gives the benefit of showing individual progress bars. The two tests I've mentioned above will still fail however.

I've moved away from using `np.hstack` for stacking the features. This is the reason for the issue mentioned in #298, so if either way, we should be able to get the cached version of the featurizer to work with matrices quite easily.

I'm happy to implement two versions of the multi featurizers but I'll wait for @WardLT before I do anything else.

utf:
I added the ability to choose whether to iterate over entries or featurizers when initializing `MultipleFeaturizer`. Caching works as expected when iterating over entries. I also fixed #298 for both modes. This means you can choose whether want multiple progress bars or better caching.

I updated the tests so that they all are run for both modes (except the caching test).

One test is still failing. This is the test I mentioned above, where if a Featurizer returns its features as a nested list it is expected to be reduced to just a list. E.g.

```
[[2]] -> [2]
```

Again, I'm unsure why `MultipleFeaturizer` is expected to do this, but all other featurizers are not. It means you will get different behavior if running a featurizer in a `FeatureUnion` vs. `MultiFeaturizer`. To me this is quite confusing but maybe I'm missing something. Would be good to get input from @montoyjh or @WardLT.

WardLT:
Yea, the progress bars and caching MultiFeaturizers has been a long standing issue. I like having an `__init__` option that turns on/off caching rather than a separate class. I'm torn on whether to set `iterate_over_entires=True` as the default option, which has a big speed benefit when recreating a few methods from the literature. However, having a option that can yield errors as the default seems ill advised. That said, I'm not sold on the need to support both matrix and vector features at the same time (hard to use both from an ML model) and many progress bars is just noise to me (I only really care about when all will finish). 

A few other comments:

- Python automatically grabs the docstring from the base class, and the "see base class" docstrings interfere with that. I'd recommend removing them
- I think the test on handling arrays with singletons was originally mine. While we don't require that featurizers return either a 1D or 2D w/ singleton array, different featurizers do return different types, making them return the same type is tedious, and the error due to mixing them up is very confusing. So, I vote keeping that test as required.

Otherwise, this looks great! I like that you chose to alter `featurize_many`, added so many nice unit tests, and made sure to set `n_jobs` for all the featurizers.

Once you fix the docstrings and decide on whether you want caching as default, this looks good to me.

utf:
Ok, I don't have a personal preference for whether caching is on or off by default. Although, it shouldn't yield errors now that #298 is addressed, right? I'll defer to what others think here @ardunn @computron @albalu?

For your other points:
- I'll remove those comments.
- I'm not sure I understand the point about the singleton test. To make the behavior more consistent would it be better to put this somewhere in `BaseFeaturizer`?

WardLT:
Oh, maybe I misunderstood about #298 being fixed. Does this revised MultiFeaturizer still fail with matrix arrays and `iterate_over_entries=False`? We still have classes that return matrix features (e.g., `SiteStatsFeaturizer` under certain conditions), so there is still a need to support both.

The singleton test was put in place after we found that different featurizers returned "lists" and "nested lists" despite seemingly unimportant differences in the codes. I'm not sure which two featurizers were culprits, but it took me a bit to debug the problem in MultipleFeaturizer - if I recall correctly - and I added this test/features to avoid users from having to learn about this subtlety of how numpy does 1D vectors. MultipleFeaturizer is the only class that need support both "list"/"nested list" because no other featurizer attempts to concatenate features from several featurizers. 

Good question about putting some logic in BaseFeaturizer to deal with the "list" vs "nested list" features. I'm not sure where we would put it though. Enforcing that all `featurize` functions return "lists" is feasible. We could also stuff the squeezing logic into `featurize_wrapper`, which means `MultipleFeaturizer.featurize` would fail but `featurize_many` would not. Though, I like having the logic in `MultiFeaturizer.featurize` to avoid the work of policing that featurizers do not return "nested lists"

Does that help?

utf:
Multifeaturizer now works with matrix arrays and `iterate_over_entries=True` (and also `False`), so the difference between implementations is solely down to caching vs multiple progress bars.

Thanks for explaining the singleton test, the need for it makes sense. My main concern is that people using `FeatureUnion` (which is recommended in the matminer example notebooks) will get different behavior than if they used `MultipleFeaturizer`. But to save the time on debugging I will make sure the tests pass by using your `np.squeeze(np.array(x))` trick for now.

WardLT:
If matrices now work with `iterate_over_entries=True`, then I'm on board with keeping that as the default

utf:
Thinking about this a bit more, I'm not sure I agree with the singleton test. I think we should trust (and therefore enforce) featurizers to return the data in the correct format and not try and second guess their behavior.

You cannot have the singleton behavior you described AND compatibility with matrices for the following reason. Say you have a featurizer that returns one feature as a matrix, the output of `Featurizer.featurize` will be:

```python
[[[1.0, 1.0], [1.0, 1.0]]]
```

With the singleton behavior, this will get converted to:

```python
[[1.0, 1.0], [1.0, 1.0]]
```

Which is indistinguishable from a featurizer that returns two features. To me, this ambiguity and second guessing is confusing. If there are only two featurizers that cause this problem, maybe we should make sure they return their features in the correct format (perhaps by using squeeze), rather affecting the behavior of all other featurizers.

utf:
Ok, I've done a thorough testing of all featurizers implemented in matminer (structure, composition, dos and bs). Only the composition based `BandCenter` was causing an issue due to the fact that it was returning the value of the featurize rather than a list containing the feature value. This has been fixed.

My implementation of `MultipleFeaturizer` now works with all featurizers (iterating over entries and featurizers). I tested to see if any featurizers were returning singleton arrays of size `(4, 1)` etc. The only features that contained any any sort of multidimensional data were:

- coulomb matrix
- sine coulomb matrix
- orbital field matrix

But this is what we expected. 

I'm proposing that we remove the singleton test as it is no longer needed. In the future, featurizers should be responsible for returning their data in the right format.

If this gets the go-ahead, I'll make that change and all tests will pass.

WardLT:
Nice! Thanks for checking the featurizers. 

In that case, simplifying the code by removing the squeezing routines sounds like a nice plan. I'm not too concerned about policing featurizers.

PR #320, opened 2018-10-11
Deprecation and compatibility fixes

Doppe1g4nger:
## Summary

Fixed many warnings, deprecations, and hidden bugs in testing.

## TODO (if any)

There is an issue that still needs to be looked into with test_conversions.py for the featurizer tests. A warning is thrown whenever StrToComposition.featurize_dataframe is called with multiindex set to True. However this should be throwing an error instead of a warning. There is an active bug in conversions.py in the class that StrToComposition inherits from: ConversionFeaturizer. This featurizer is set to throw an error when multiindex is set to True, but there is a bug in the check where it looks for the keyword "multiiindex" instead which causes it to not fail. Instead it falls through to the BaseFeaturizer which throws a warning. I have no clue what the actual behavior of these featurizers is supposed to be when multiindex is set to True so I didn't fix it, but @utf should probably take a look.

### DIFF
--- a/matminer/data_retrieval/retrieve_Citrine.py
+++ b/matminer/data_retrieval/retrieve_Citrine.py
@@ -78,7 +78,7 @@ class CitrineDataRetrieval(BaseDataRetrieval):
             properties = list(set(properties))
         all_fields = []
         for prop_counter, requested_prop in enumerate(properties):
-            jsons = self.get_data(**criteria, prop=requested_prop)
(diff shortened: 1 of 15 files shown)


utf:
This looks great. Good spot with the multiindex misspelling. I actually fixed this in #314 its just not been merged yet. It's good that you didn't edit it as it should reduce the number of conflicts when merging.

Issue #324, opened 2018-10-29
Add dataset management guide and table of dataset descriptions to doc 


Doppe1g4nger:
First mentioned in #322, a guide to managing the datasets available in matminer should be added to the doc (already available as an md file in /dev_scripts/dataset_management/, just needs reformatted). Also an auto generated human readable list of datasets and information about them should be added to the doc in a similar manner to how the featurizers are listed. I'll get to work on this soon, but as a separate PR from the current work of adding datasets to the package.

Doppe1g4nger:
Closed by #328

PR #322, opened 2018-10-17
Further dataset management upgrades and addition of datasets

Doppe1g4nger:
## Summary

This PR further improves matminer's handling of datasets. Several getter functions were added for users to pull information about the various datasets available in matminer. 

The dataset metadata was moved to a dedicated JSON file for easier editing and access and a dedicated script was added to dev_scripts for interfacing with this file. Information can of course still be tweaked in the JSON file directly but using this script provides a standard interface to minimize mistakes during initial data entry. 

In addition several datasets were added from matbench. This is part of 16 to be added and serves as an example of how few changes need to be introduced to matminer to add a dataset, simply needing some small updates to the test and metadata files and a preprocessor added to the dataset prep script.

* Added castelli perovskites dataset to matminer
* Added boltztrap mp dataset to matminer
* Added phonon dielectric mp dataset to matminer
* Added getter functions for accessing dataset metadata
* Moved dataset metadata to JSON file (info is cached on first access to avoid heavy file io)
* Added script for editing the dataset metadata

Issue #279, opened 2018-08-03
add Jarvis data set to matminer sample data sets


computron:
Not sure if we should be moving larger sample data sets to another package, or coding a way to download the raw data file in JSON on-demand

https://github.com/usnistgov/jarvis/blob/master/jarvis/db/static/jarvis_ml-train.ipynb

ardunn:
@computron considering we have a majority of the JARVIS database in our Figshare datasets, as well as the MDF Forge JARVIS access, can we consider this issue closed?

computron:
yes please close

Issue #261, opened 2018-06-26
NIST JARVIS dataset as a matminer example data set


computron:
Recently Kamal Choudhary et al published some good ML results based on mining the JARVIS database:

https://arxiv.org/abs/1805.07325

It would be great if that data set can be available on matminer so others can try their ML algorithms. I asked Kamal about how to do this and he referred me to this code:

https://github.com/usnistgov/jarvis/blob/master/jarvis/db/static/explore_db.py

It seems like all the info needed to make the dataset available on matminer as an example is there. If the file is big, we may want to have a separate repo for large datasets so the matminer repo doesn't become bloated with datasets.

Separately, Kamal has discussed adding his descriptors to matminer. But for now, just having the data set would already be interesting.

WardLT:
We have indexed JARVIS in the Materials Data Facility, and I made an example for using our index. I ported that example to use matminer's data retrieval class. Should I add this to the example repository?
[jarvis-example.pdf](https://github.com/hackingmaterials/matminer/files/2138425/jarvis-example.pdf)

I still have a little bit of work to do to make it easier to access the crystal structures, which would allow for machine learning.

computron:
I do think you should add the example to matminer_examples

That said, it would be nice to have just a single function call to get both the crystal structures and the data for the referenced arxiv paper like the other methods without having to issue Forge queries, etc. So that the data is there like the other test datasets with a single Python function call -> pandas dataframe.

WardLT:
Sure, I'll go ahead and add it to matminer_examples. 

I certainly agree, issuing multiple queries to get the structures would be tedious and presents an unneeded learning curve. We're still figuring out best to approach that problem.

ardunn:
@computron considering we have a majority of the JARVIS database in our Figshare datasets, as well as the MDF Forge JARVIS access, can we consider this issue closed?

ardunn:
This issue is closed, JARVIS DFT3D and 2D and the exact training dataset Choudhary et al. used are already in matminer, see matminer dataset summary page

PR #328, opened 2018-11-05
Dataset management upgrades

Doppe1g4nger:
## Summary

This PR provides the improvements from #322 with fixes to the issues that caused it to be reverted. 

* Dataset metadata storage has been overhauled and a script has been created to provide an easier interface between the .json file containing metadata and users desiring to edit it
* Documentation has been added for adding new datasets to matminer and an extensive summary of each dataset is autogenerated. closing #324 
* 18 new datasets have been added to matminer in this PR, including Jarvis DFT, closing #279 #261 
* Added a set of getter functions for pulling data from the .json metadata file, added convenience functions for loading in datasets that may require postprocessing after loading.
* Added tests for all datasets, ones that take a long time are currently skipped to avoid excessively long CircleCI runs.

The dataset summary page looks like so:
image

## TODO (if any)

This PR should be ready to merge immediately. Work for near future PR's however include:

**EDIT: Documentation as detailed below has been added to this PR.**

~~Documentation:
As per issue #324 The guide for managing datasets should still be moved to the official doc pages and a table of available dataset info still needs generated.~~

Testing:
Current tests for datasets should be changed to run consistently, but not during every PR. The most obvious way of doing this I can think of is to set up a cronjob to run these tests and take them out of CircleCI, but it may be able to be done within CircleCI itself with the right configuration settings.

More Datasets:
@Qi-max has updated one of the datasets currently in matbench and a v2 version should be added to matminer, in addition I will soon add the superhard materials dataset from my work and the elastic tensor dataset should be updated to encompass the updated elastic tensor data in MP since 2015.

**EDIT: Unless overruled from on high, keeping the loaders as they are. ```load_dataset``` will be indicated as the main way of interfacing with datasets by keeping it the only function in the ```matminer.datasets``` namespace, convenience loader stay in ```matminer.datasets.convenience_loaders```**

~~Future of "Convenience Loaders" should be decided:
In the process of adding datasets I realized there are some dataset-specific processing routines that can't be done prior to serializing as JSON and storing as a dataframe. For example, choosing whether or not to drop nan values or what version of a dataset to load. To provide these options I wrote a set of "convenience loaders" that provide dataset specific options for loading data. However having these conflicts with the Python rule of thumb to only ever have one obvious way of doing things (e.g. using load_dataset("dataset_name") vs load_dataset_name() is non-obvious) and I don't know if I should keep them in matminer or not.~~

~~The pros I see are:~~
~~* Gives lots of convenient loading options directly in matminer~~

~~The cons:~~
~~* Adds confusion to proper API usage~~
~~* Goes against Pythonic design principles~~

~~I could resolve things by removing these loaders or making the load_dataset function private and make the convenience loaders the recommended way of loading datasets. Either way once this is decided some slight changes will need to be made to the docs on adding datasets procedure and the matminer examples repo will need a few slight updates.~~


ardunn:
@Doppe1g4nger I would vote against making load_dataset function private. 

Perhaps the dataset specific routines could be moved to the load_dataset function instead?

Doppe1g4nger:
@ardunn I'm against that as having those options in the load_dataset function seriously clutters the function and makes it a lot more confusing to use as there will be lots of dataset specific parameters. I actually wrote the loaders in the first place to avoid those problems haha.

If we want to keep load_dataset as the primary way of doing things in matminer I like the idea of moving the convenience loaders over to matbench. I wrote these functions mostly to keep things consistent with the loaders already in matbench anyway so we could take an approach of "if you want special parameters for modifying x dataset from matminer, write the loader in your own project instead of changing matminer itself".

Doppe1g4nger:
The four most recent commits add extensive documentation about the datasets to the project docs and provide a guide for adding new datasets. See the edited initial PR request for details and a screengrab of what the documentation looks like.

@computron during my update of the docs I noticed several leftover autogenerated docs for submodules that no longer exist such as "matminer.learner" as well as some docs that were being generated for tests because the test folders had __init__.py files in them. I took the liberty to remove these and fix the __init__ issue. I thought I should mention that to you in case these were left in on purpose. The relevant changes should all be in commits 6b07f48 and e4328bb above.

@ardunn once Anubhav gives the go ahead this should be ready to merge.

dev_scripts/dataset_management/modify_dataset_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+How would you like to interact with the dataset metadata?
+(1) See Current Datasets
+(2) Remove a Dataset
+(3) Add a Dataset
+(s) Save changes
+(q) Quit
+"""
+
+
+def generate_new_dataset(dataset):
```
WardLT: Could you document what this function does and its inputs?

matminer/datasets/tests/test_convenience_loaders.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_load_jarvis_dft_2d(self):
+        df = load_jarvis_dft_2d()
+        self.assertTrue(isinstance(df, pd.DataFrame))
+        self.assertEqual(len(df), 636)
+        df = load_jarvis_dft_2d(drop_nan_columns='epsilon_x opt')
+        self.assertEqual(len(df), 522)
+
+    # Skip for circleCI efficiency
+    @unittest.skip
```
WardLT: Could you use the "skipIf" function of unittests to have this one run when not on a CI tool. I think `https://discuss.circleci.com/t/skip-certain-tests-in-python-nosetests/14001` shows how to check if you are on a CI. Never tried this, so I'm not sure if my suggestions are possible.

matminer/datasets/tests/test_convenience_loaders.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertTrue(isinstance(df, pd.DataFrame))
+        self.assertEqual(len(df.columns), 17)
+
+        df = load_elastic_tensor(include_metadata=True)
+        self.assertTrue(isinstance(df, pd.DataFrame))
+        self.assertEqual(len(df.columns), 20)
+
+    # Skip for circleCI efficiency
+    @unittest.skip
+    def test_load_glass_ternary_landolt(self):
```
WardLT: See comment below about skipping on CI but having it run for everyone else.

matminer/datasets/dataset_retrieval.py
```diff
@@ -8,9 +8,10 @@
              "Anubhav Jain <ajain@lbl.gov>" \
              "Daniel Dopp <dbdopp@lbl.gov>"
 
+_dataset_dict = None
 
-def load_dataset(name, data_home=None, download_if_missing=True,
-                 include_metadata=False):
+
+def load_dataset(name, data_home=None, download_if_missing=True):
```
WardLT: @ardunn Is this this `load_dataset` method in question? It does seem like a really useful one to have public.

matminer/datasets/dataset_retrieval.py
```diff
@@ -90,26 +85,122 @@ def available_datasets(print_datasets=True, print_descriptions=True,
...
+            print(name, _dataset_dict[name]["description"], "", sep="\n")
     elif print_datasets:
         for name in dataset_names:
             print(name)
 
-    return list(dataset_names)
+    return dataset_names
+
+
+def get_dataset_attribute(dataset_name, attrib_key):
```
WardLT: Is there a utility function for just getting the whole dataset description?

matminer/datasets/tests/test_convenience_loaders.py
```diff
@@ -0,0 +1,181 @@
+import pandas as pd
+import unittest
+
+from matminer.datasets.tests.base import DataSetTest
+from matminer.datasets.convenience_loaders import load_glass_ternary_hipt, \
```
WardLT: In sklearn and keras, the loaders for the datasets are in the dataset module. Example, 

```python
from sklearn.datasets import load_iris
iris = load_iris
```

The pedantic side of me wants to suggest to overhaul our library to follow this convention. But, the only practical benefit of moving things from `convenience_loaders` into `datasets` are the names being quicker to type. So, I am just pointing this semantic difference out without any intention of making a stink about if you disagree

WardLT:
As you've seen, I left a few suggestions on the data loader code.

Overall, these loaders will be a really useful feature. Nice work on making the loaders and tracking down/organizing all of these datasets!

matminer/datasets/dataset_retrieval.py
ardunn: yes this is the one i'd like to keep public as well

matminer/datasets/dataset_retrieval.py
Doppe1g4nger: Yes, get_dataset_description(dataset_name) returns a human readable string description of the dataset.

WardLT: Oops, I must have overlooked it.

matminer/datasets/tests/test_convenience_loaders.py
Doppe1g4nger: It's a tough call, because the way seaborn does things is with the load_dataset(string_identifier) method like so.

```python
from seaborn import load_dataset
iris = load_dataset("iris")
```

Initially we thought this method would make for a better interface but then I realized it would be nice to have dataset specific loading options so I added the convenience loaders. But then you get this problem of which one do you consider the "right" way of getting datasets. I like keeping things where they are with the convenience loaders in their own namespace to show they aren't considered the "primary interface" but rather utilities for user convenience.

WardLT: Keeping them separate sounds like a good plan to me. Thanks for walking me through your thinking about this!

matminer/datasets/tests/test_convenience_loaders.py
Doppe1g4nger: I implemented this change, and it runs just fine locally, but it seems CircleCI is now failing to run the tests. I don't know if this is an issue caused by my update or something on the CIrcleCI side of things...

dev_scripts/dataset_management/modify_dataset_metadata.py
Doppe1g4nger: Done

matminer/datasets/tests/test_convenience_loaders.py
Doppe1g4nger: Update: It was an issue with how the tox environment was set up and the checking of environment variables. It's now been resolved

Doppe1g4nger:
As per @computron 's suggestion I added an environment variable check against a dedicated environment variable rather than the "CIRCLECI" environment variable. The tests default to the less comprehensive version of the dataset integrity check and users can enable the full tests by setting the MATMINER_DATASET_FULL_TEST environment variable to "True".

If we want to add a scheduled check in the future, we only need to set up a tox config that sets this variable, everything else should be able to be left the same.

PR #329, opened 2018-11-07
Larger default for chunksize

WardLT:
I think the current default value for `chunksize` used in multiprocessing is too small. 

When upgrading to the latest version of matminer, my performance dropped by ~100x for parallel featurization (not entirely sure because it too so long I killed the process). I traced it to the choice of `chunksize` default, and after switching to using a `chunksize` of `None` (which causes matminer to rely on the default for `multiprocessing`) I was back to having acceptable performance. 

The featurizerizers I was using are pretty representative of the others in matminer, so I suspect the default for `multiprocessing` is likely a good choice for the default for matminer.

## Summary

This commit changes the code to rely on the default for multiprocessing rather than 1 for the chunksize. 

## TODO (if any)

None

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -147,7 +147,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
 
     @property
     def chunksize(self):
-        return self._chunksize if hasattr(self, '_chunksize') else 1
+        return self._chunksize if hasattr(self, '_chunksize') else None
 
     def fit(self, X, y=None, **fit_kwargs):
         """Update the parameters of this featurizer based on available data
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -82,7 +82,7 @@ class BaseFeaturizer(BaseEstimator, TransformerMixin):
     that are set by fitting should be stored as class attributes that end with
     an underscore. (This follows the pattern used by ScikitLearn).
 
-    Another implementation to consider is whether it is worth making any utility
+    Another option to consider is whether it is worth making any utility
     operations for your featurizer. `featurize` must return a list of features,
     but this may not be the most natural representation for your features (e.g.,
     a `dict` could be better). Making a separate function for computing features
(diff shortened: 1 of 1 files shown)


utf:
Interesting, I was under the impression that the default chunksize for `multiprocessing.Pool.map` was 1. It turns out that the default chunksize for `multiprocessing.Pool.imap` is 1 but `map` is different. Looking at cython it is actually calculated as:

```
if chunksize is None:
    chunksize, extra = divmod(len(iterable), len(self._pool) * 4)
    if extra:
        chunksize += 1
if len(iterable) == 0:
    chunksize = 0
```

However, I'm still very surprised that it caused such a difference in featurization time. I did some testing and for the heavier featurizers I found that the difference between a chunksize of 1 and 100 was very little. I've also been using the new version fairly consistently and haven't noticed a significant increase in featurization time. I wonder if others have?

What featurizers in particular were slowing you down? And was this using `MultipleFeaturizer` as there could be some overhead there.

Anyway, I'm pro this change, but could you update the BaseFeaturizer docs to mention the new default chunksize of `None`.

WardLT:
Hello @utf ,

All good questions about what makes my case so pathological for chunksize=1. I'm using a prototype featurizer across 85k features, and I think the problem has to do with having some slow-to-serialize class attributes for the featurizer. I haven't open-sourced this project yet, but can show you the featurizer code if you want more details.

I suspect the issue is that these featurizers take awhile to serialize and send them to the worker threads, and this cost is becomes huge with small chunk sizes. I'm not sure if `multiprocessing` serializes and sends the function to be evaluated each time it sends a new chunk to a worker thread, but my performance numbers are acting that way.

I have two particular featurizers where I've observed the performance decrease. One has a minor amount of class attributes to be serialized, I a see a 8x slowdown between `chunksize=None` and `chunksize=1`. For a featurizer were I have a large numpy array as a class attribute, I see a huge slow down.

So, I probably have a pathological case for `chunksize=1`. But, it sounds like we're both in favor of using `multiprocessing`'s default. Do that logic and my changes to the chunksize guidance sound reasonable?

utf:
Hi @WardLT, thanks for going into so much detail about your pathological cases. It definitely sounds like both of those issues could cause a massive slow down. I'll have to bear this in mind in the future.

I think it's a good idea that we use the `multiprocessing` default (this was my intention originally, so I'm grateful that you've caught this). The updated docs also look good.

PR #334, opened 2018-11-15
Updating datasets and adding scripts for generating them

Doppe1g4nger:
This PR adds a new version of the glass_binary dataset @Qi-max has been working with to matminer. This version has duplicate entries merged/eliminated. The PR also adds a new script to generate the most recent elastic tensor dataset from MP and gives the option to compute oxidation state info when generating datasets.

I'm making this PR so @utf can run these scripts on Cori to generate the datasets much faster than I could on a local machine or trying to get access to NERSC myself.

EDIT: This PR also adds the brgoch dataset


Doppe1g4nger:
@computron This PR should now be ready to merge. It updates the glass_binary dataset, finalizes the brgoch dataset, and adds some dataset generation scripts. The tests are failing because of a failed figrecipes test, it passes on my local machine but fails on circleCI, perhaps @ardunn should take a look.

PR #339, opened 2018-12-06
Allow oxidation conversion featurizers to return original object

utf:
## Summary

Adds the ability for the `CompostionToOxidComposition` and `StructureToOxidStructure` conversion featurizers to return the original structure or composition object if adding oxidation states fails.


utf:
I've also made it so that the oxidation states are not recalculated if they already exist in the structure.

computron:
can you rename both parameters to:

return_original_on_error

?

utf:
Done

Issue #280, opened 2018-08-07
add Jarvis descriptors to matminer


computron:
See: 

https://github.com/usnistgov/jarvis/blob/master/jarvis/sklearn/get_desc.py

computron:
See also:

https://github.com/usnistgov/jarvis/blob/master/jarvis/db/static/jarvis_ml-train.ipynb

Kamal Choudhary has (friendly) dared people to try to beat the results...

ardunn:
this is done

PR #342, opened 2018-12-08
Add JARVIS Descriptors, + a few random things

ardunn:
## Summary

* Add the JARVIS CFID descriptors into a new Featurizer, `JarvisCFID`, which takes a structure and outputs anywhere from 4-1557 features (depending on init args). Implementation was based heavily on [the nist implementation](https://github.com/usnistgov/jarvis), with some major revisions to make the code clearer. Two reference json files are needed for `JarvisCFID` to work, and they are found in (edit) the same utils directory as the ElementProperty and MagPie data
* Add a simple test for this new featurizer
* Fix the doc string for a few featurizers which was causing the Table of Featurizers in the docs not to display the description
* an autoreformat of structure featurizers

## TODO (if any)
Docs should be regenerated either in this PR or soon after.


ardunn:
Closes #280

ardunn:
@WardLT you're right It's been moved :)

computron:
@ardunn FYI I am going to add Kamal as an implementor after pulling your request, as I believe a lot of the code was adapted from what he sent you regarding Jarvis

ardunn:
@computron kk sounds good

PR #350, opened 2018-12-20
CGCNN Updates and activating CGCNN tests

ardunn:
## Summary

Making CGCNNFeaturizer a bit easier to use.

* `matminer/utils/install_cgcnn`: A bash install script (now shown by `@requires` in CGCNNFeaturizer so people know where it's at) which makes it easier to install and use CGCNN. Without this, installing via anaconda is clunky because CGCNN doesn't have a setup.py and ins't pip installable either.
* Updated optional requirements to include those needed by CGCNNFeaturizer
* Updated circleci config to be able to run CGCNNFeaturizer tests
* A fix of `Variable` attrs in CGCNNFeaturizer to get the tests to pass


ardunn:
An example of the setup script for CGCNN:

screen shot 2018-12-20 at 12 56 41 pm

computron:
Are you sure we want to maintain an Install_cgcnn.sh script in matminer? I feel like that should go in cgcnn.

e.g., I don't want our group to be on the hook to fix if that script doesn't work.

ardunn:
@computron After all this trouble, no. I think I'll just make a PR to their repo

computron:
Ok another option is that you can create a gist in your personal github account for it. The gist can clearly say this is not maintained ...

but better if CGCNN folks maintain it

PR #351, opened 2018-12-21
Fix pip install of 0.4.7+

ardunn:
## Summary

Pip install from pypi appears to install ok, but in reality does not pass tests. This is because jarvis files were not included in manifest/setup (for bdist/sdist) when last version was released. This PR addresses this issue.

Also includes a fix for CGCNN test, so that when tests are run and user actually has CGCNN, it will work.


ardunn:
not sure why this is failing while https://circleci.com/gh/hackingmaterials/matminer/1515 is passing

computron:
Maybe it has to do with the convoluted series of commits leading the to the final changes? Let me try to merge and see what happens

ardunn:
@computron maybe has to do with changes on circleci's (or tox's) end? It looks like what is going on is something goes wrong with numpy compilation, which then prevents pymatgen's LinearAssignment from being linked/imported/generated?

PR #353, opened 2018-12-24
WIP: Fix for inaccurate progress bar

### LINKED ISSUE #336: Pbar not accurate
Pbar + multiprocessing current implementation is not accurate for the actual progress being made. When a process starts working on an entry, the pbar updates; a lot of the time, this means the bar will appear 100% finished but will actually still be featurizing some slow entries. 

The pbar should update after the entry is featurized. A possible implementation for a fix is manually updating a self._pbar attr in featurize_wrapper

KaranDSharma:
Fixes #336
Changed progress bar from using the .map function to use the .map_unordered function. Should fix the issue of the pbar appearing at 100% even though it is still handling some data. This way a long featurize call wont block the progress bar from updating.

(in base.py)
* added featurize_imap_wrapper function
* changed featurize_many function to implement map_unordered

## TODO
Verify the changes and make alterations as I have been experiencing clientside issues I am still in the process of fixing.
------------------------------------------
EDIT 2
• Resolved argument error in featurize_imap_wrapper helper function
• Minor indexing error fixes when appending to result in featurize_many
• Added check for self.chunksize=None as imap doesn't support None value chunksize
• Added detail to helper function docstring


setup.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+# Karan Test
```
utf: Remove this

matminer/featurizers/base.py
```diff
@@ -364,10 +364,28 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
...
+                    enumerate(entries))  # turns entries into [(0, entry_1), (1, entry_2), (2, entry_3), ... ]
+                results = []
+                for i, result in tqdm_func(
+                        p.imap_unordered(func, entries, chunksize=self.chunksize), total=len(entries)):
+                    results[i] = result
+                return results
+
+    def featurize_imap_wrapper(self, data):
+        """
+        A helper function to featurize the tuple form data
```
utf: Can you make this doc string explain what it is doing. E.g. it accepts the index of the entry and the entry and returns the index and the features.

matminer/featurizers/base.py
```diff
@@ -364,10 +364,28 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
                                            return_errors=return_errors,
                                            pbar=pbar)
             with Pool(self.n_jobs) as p:
-                func = partial(self.featurize_wrapper,
+                func = partial(self.featurize_imap_wrapper,
                                return_errors=return_errors,
                                ignore_errors=ignore_errors)
-                return p.map(func, entries, chunksize=self.chunksize)
+                entries = list(
```
utf: You can put the comment on the line above and put the rest of this on 1 line instead of 2.

matminer/featurizers/base.py
```diff
@@ -345,7 +345,7 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
             else:
                 tqdm_func = tqdm.tqdm
             # list() required, tqdm has issues with memory if generator given
-            entries = tqdm_func(list(entries), desc=self.__class__.__name__)
+            # (don't wrap entries in tqdm call) entries = tqdm_func(list(entries), desc=self.__class__.__name__)
```
utf: You will still need to wrap the entries in tqdm for the non-multiprocessing case. E.g. when n_jobs is 1.

WardLT:
Thanks for the contribution, Karan!

Beyond Alex's comments:

1. Any idea about the overhead of using this alternate iterator? I don't have any benchmarks for the parallelization of matminer, but you should run the notebooks in the examples to see if they run at similar speeds as before.
2. We'll want to be careful with the chunkszie because the default for imap (1) can lead to some very poor performance for some featurizers. I'd check if you can emulate the chunksize of map.
3. I think we should make the new wrapper function private, as it is likely only for internal use. Any thoughts?

utf:
The chunksize for map is calculated as:

```
if chunksize is None:
    chunksize, extra = divmod(len(iterable), len(self._pool) * 4)
    if extra:
        chunksize += 1
if len(iterable) == 0:
    chunksize = 0
```

Definitely agree about performance testing to check the overhead is minimal.

utf:
Also, it looks like imap doesn't support passing `chunksize=None`. So we will have to catch that (and calculate the chunksize using the formula above) before passing on `self.chunksize`.

matminer/featurizers/base.py
ardunn: Is there a reason we can't just replace featurize_wrapper with featurize_imap_wrapper? It seems like a bit much to have one featurize wrapper for non-tqdm featurization and a different featurize_wrapper for tqdm featurization...

ardunn:
btw @KaranDSharma our CI is failing rn due to bad pymatgen install. Those errors (frozenlib + ImportErrors + numpy compilation errors) are due to our CI config. However the other errors are from ur code and should be fixed before merging.

KaranDSharma:
Noted. Let me know if anything else needs revision from my second submission.

matminer/featurizers/base.py
```diff
@@ -364,10 +364,35 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
...
+                    if extra:
+                        chunksize += 1
+                else:
+                    chunksize = self.chunksize
+                for i, result in tqdm_func(
+                        p.imap_unordered(func, entries, chunksize=chunksize), total=len(entries)):
+                    results.append(result)
+                return results
+
+    def featurize_imap_wrapper(self, data, return_errors=False, ignore_errors=False):
```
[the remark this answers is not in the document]
utf: Hi Karan, please can you look into this?

setup.py
utf: Please fix this

matminer/featurizers/base.py
utf: I think a wrapper of this sort will be the cleanest way to achieve this. However, I think your lambda idea will tidy it up further.

matminer/featurizers/base.py
```diff
@@ -364,10 +364,35 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
                                            return_errors=return_errors,
                                            pbar=pbar)
             with Pool(self.n_jobs) as p:
-                func = partial(self.featurize_wrapper,
+                func = partial(self.featurize_imap_wrapper,
                                return_errors=return_errors,
                                ignore_errors=ignore_errors)
-                return p.map(func, entries, chunksize=self.chunksize)
+                entries = list(enumerate(entries))  # turns entries into [(0, entry_1), (1, entry_2), (2, entry_3), ... ]
```
utf: At this point entries is still wrapped by tqdm (see line 348). When you convert the tqdm function into a list it will produce a progress bar that goes from 0 to 100 % immediately. You then call tqdm again on line 379, thereby producing a secondary progress bar.

You need to rewrite the logic so that there is only one progress bar if p.imap_unordered is called. I.e. only call tqdm_func on entries in the `n_jobs == 1` or `sys.version_info[0] < 3` cases (provided `pbar == True`).

utf:
Hi @KaranDSharma, thanks again for your work on this.

Please can you fix the remaining issues highlighted in the code review. If you go to https://github.com/hackingmaterials/matminer/pull/353/files you can see the outstanding comments.

Additionally, you should respond to @WardLT comments. Specifically, you should run some benchmarks to test that your new implementation runs at a similar same speed to the old implementation. For this you should check out the `%timeit` magic keyword in iPython/jupyter notebooks.

Lastly, Logan's comment about making the wrapper function private should instead be fixed by replacing the imap wrapper function with a lambda. I.e. this should be as simple as replacing the `func = Partial(...` expression with:

```python
                func = lambda data: (data[0], self.featurize_wrapper(
                    data[1], return_errors=return_errors,
                    ignore_errors=ignore_errors))
```

utf:
Oh, and I guess it would be good to pull in the latest changes from the upstream master branch, as those should hopefully fix the failing tests.

E.g. from within the matminer repo folder run:
```
git pull upstream master
git push
```

matminer/featurizers/base.py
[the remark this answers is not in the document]
KaranDSharma: I tried a lambda and a local function and both tries failed because they couldn't be pickled.
lambda: AttributeError: Can't pickle local object 'BaseFeaturizer.featurize_many.<locals>.<lambda>'
local function: AttributeError: Can't pickle local object 'BaseFeaturizer.featurize_many.<locals>.func'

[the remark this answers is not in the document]
WardLT: I'm not sure if this can be done with a local function (lambda or otherwise). It might be possible to use some combination of `partial` and `partial_method` from [functools](https://docs.python.org/3/library/functools.html), but that sounds tricky. I'd be fine with leaving this as a class method, but think it should be a private function.

utf:
Hi Karan, is there any updates for the remaining comments/issues mentioned above?

KaranDSharma:
> Hi Karan, is there any updates for the remaining comments/issues mentioned above?

Sorry for the delay, I have been out of the state with limited wireless access. When comparing the original code to the new implementation the progress bar seems to load up to a full second faster. I believe all other comments have been addressed, though please let me know if I overlooked one!

matminer/featurizers/base.py
```diff
@@ -345,7 +345,10 @@ def featurize_many(self, entries, ignore_errors=False, return_errors=False,
             else:
                 tqdm_func = tqdm.tqdm
             # list() required, tqdm has issues with memory if generator given
-            entries = tqdm_func(list(entries), desc=self.__class__.__name__)
+            if sys.version_info[0] < 3 and n_jobs == 1 and pbar == True:
```
utf: This should be:

```python
if (n_jobs == 1 or sys.version_info[0] < 3) and pbar:
```

utf:
> Oh, and I guess it would be good to pull in the latest changes from the upstream master branch, as those should hopefully fix the failing tests.
> 
> E.g. from within the matminer repo folder run:
> 
> ```
> git pull upstream master
> git push
> ```

Thanks for the update. There are still several issues to address. See my comment quoted above for information on how to get the tests to pass. In addition, would you be able to give specific timing information as produced by `%timeit`. This will run the functions multiple times to give you more accurate statistics. I'd be interested to know which featurizers you tested and the timing information for each.

Just to summarise, before we can pull in the changes the following issues should be addressed:
- Make `featurize_imap_wrapper` private.
- Fix conditional statement to decide when to wrap entries in tqdm.
- Pull and push the latest changes from the upstream master branch to fix failing circleCI tests.
- A bit more information on the timings for the new vs old implementations.

Thanks!

matminer/featurizers/base.py
ardunn: We could also just remove this version check as pymatgen no longer officially supports Py2. We could move it to __init__ in root folder maybe so when stuff from matminer is imported and the version is <3, a warning is thrown saying the version is not supported?

computron:
@utf will you be updating this PR, or should I close it?

utf:
I will finish this off.

ardunn:
@utf :(

PR #358, opened 2019-01-12
added site-wise DOS featurizer

dyllamt:
Not sure if this featurizer should be included in dos.py or site.py. There could be advantages to both.

## Summary

Commit contents:
- helper function for retrieving s/p/d/f dos scores for a particular site
- SiteDOS featurizer, which returns s/p/d/f dos scores for the vbm and cbm for a particular site
- accompanying tests


matminer/featurizers/dos.py
```diff
@@ -6,6 +6,78 @@
...
+    """
+    report the fractional s/p/d/f dos for a particular site. a CompleteDos
+    object is required because knowledge of the structure is needed. this
+    featurizer will work for metals as well as semiconductors. if the dos is a
+    semiconductor, cbm and vbm will correspond to the two respective band
+    edges. if the dos is a metal, then cbm and vbm correspond to above and
+    below the fermi level, respectively.
+
+    Args:
+        decay_length (float in eV):
```
computron: why not use a Fermi function for decay?

matminer/featurizers/dos.py
```diff
@@ -6,6 +6,78 @@
...
+            cutoff at five times the decay length (1% sampling strength)
+        sampling_resolution (int):
+            number of points to sample dos
+        gaussian_smear (float in eV):
+            Gaussian smearing (sigma) around each sampled point in dos
+
+    Returns (list of floats):
+        cbm_score_i (float): fractional score for i in {s,p,d,f}
+        cbm_score_total (float): the total sum of all the {s,p,d,f} scores
+            this is usefull information when comparing the relative
```
computron: spelling: usefull --> useful

matminer/featurizers/dos.py
```diff
@@ -6,6 +6,78 @@
...
+            dos (pymatgen CompleteDos or their dict):
+                dos to featurize, must contain pdos and structure
+            idx (int): index of target site in structure.
+        """
+        if isinstance(dos, dict):
+            dos = CompleteDos.from_dict(dos)
+        if dos.structure is None:
+            raise ValueError('The input dos must contain the structure.')
+
+        orbscores = get_site_dos_scores(dos, self.decay_length,
```
computron: would be better if the order of parameters was:
- dos
- idx
- decay length
- sampling_resolution
- gaussian smear

It just makes more sense to me since the dos and idx are the most "important and required" parameters, the other ones are just adjustments that would sometimes even be optional params with defaults

matminer/featurizers/dos.py
dyllamt: Could be used, but after our discussion, I think the exponential decay is OK.

matminer/featurizers/dos.py
dyllamt: Fixed spellings

matminer/featurizers/dos.py
dyllamt: changed argument order

dyllamt:
The failing tests have to do with an import error for the materials data facility. That is the only failing test. Should I add `mdf_forge` to the list of mandatory required packages? It is currently listed as an optional package.


TEST ERROR:

`ModuleNotFoundError: No module named 'mdf_forge'`

matminer/featurizers/dos.py
computron: ok

computron:
Can you coordinate the test failing with @ardunn ? I thought he had the tests passing

dyllamt:
@ardunn we can add `mdf_forge` to "tests_require" in `setup.py`. This should fix the failing test, but not require the additional dependency in a basic installation. What do you think?

ardunn:
@dyllamt attempting to fix this in #361

ardunn:
@dyllamt I just verified that current master branch's tests are passing. This might be a problem with circleci. Maybe worth closing this PR and submitting a new one?

computron:
Looks like tests pass now, @dyllamt should I merge?

dyllamt:
@ardunn you can merge now. I'm working on the follow-up pull request now.

PR #362, opened 2019-01-29
AFLOW retriever and graceful handling of optional dependencies

dyllamt:
## Summary

* AFLOW retriever class
  * pymongo-like queries are converted into the AFLUX API syntax with `RetrievalQuery`. This class also handles collecting the API requests through HTTP.
  * `get_dataframe`can collect basic material properties and  pymatgen Structure objects from AFLOW
  * accompanying tests ensure that there is proper type casting of the results

* Dependency handling
  * dependencies are stored as .yaml files
  * there are "optional" and "required" dependencies
  * if a dependency is optional, then it is required for testing (this will ensure that circle ci tests pass without forcing users to install those requirements)

## TODO

* implement DOS and band structure retrievals for the AFLOW retriever


dyllamt:
need to rethink dependency strategy

PR #363, opened 2019-01-29
AFLOW Retriever

dyllamt:
## Summary

AFLOW uses the AFLUX API system. The (unoficial) *aflow* library helps in constructing and collecting HTTP requests within this system. This retriever can gather basic material quantities (band gap, composition, etc.) as well as the relaxed structures.

* the `get_dataframe` method returns results both basic material quantities as well as relaxed structures
* AFLUX queries are built by the `RetrievalQuery` class, which supports pymono-like
* accompanying tests ensure that there is proper type-casting of the results
* made aflow an optional feature of the matminer project (see extras requirements in setup.py)

## TODO (if any)

* implement retrieval of dos and bandstructure

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ jobs:
       - run:
           name: Set up env via pip
           command: |
+            export CI=circle
             python3 -m venv test_env
             . test_env/bin/activate
             pip install --upgrade pip
@@ -40,6 +41,7 @@ jobs:
       - run:
           name: Set up env via pip
           command: |
+            export CI=circle
             python3 -m venv test_env
             . test_env/bin/activate
             pip install --upgrade pip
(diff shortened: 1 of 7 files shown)


ardunn:
@dyllamt could we switch the test_suite to unittest instead? I've been meaning to do that for a while now. Nose is a deprecated package afaik. If we are going to use nose, we should use [Nose2](https://github.com/nose-devs/nose2), not nose. But still it might be better to just use the plain ol built-in unittests:

```python
test_suite="tests"
```

ardunn:
@dyllamt think I figured out the unittest fix for this.

1. Add an `__init__.py` to all test directories (should have already been done, but alas)
2. `test_suite="matminer"`

If this still doesn't work, then maybe it is worth looking into Nose2.

ardunn:
Also you'll need to make sure there are no rogue importables, e.g., httplib2 in mpds tests or base code.

computron:
Thanks!

PR #367, opened 2019-02-09
Flattener for *CoulombMatrix via Eigenvalues

ardunn:
## Summary

Takes eigenvalues of either CoulombMatrix or SineCoulombMatrix and pads them to equal length after being fit on data. 

Example:
```python
scme = CoulombMatrixEigenvalues(coulomb_matrix=SineCoulombMatrix())
df = pd.DataFrame({"s": [Al, Ni3Al]})
print(scme.fit_featurize_dataframe(df, "s"))
```

Out:
```
                                                   s  SineCoulombMatrix eig 0  SineCoulombMatrix eig 1  SineCoulombMatrix eig 2  SineCoulombMatrix eig 3
0                                    [[0. 0. 0.] Al]               235.740418                 0.000000                 0.000000                 0.000000
1  [[0. 0. 0.] Al, [1.76 1.76 0.  ] Ni, [1.76 0. ...               232.578562              1656.288171              1403.106576              1403.106576
```

@WardLT is this what you were referring to in #299 ?


WardLT:
Yea, this does cover generally what I had in mind in #299 . Have you considered whether to make the flattening an option of the *CoulombMatrix classes rather than a separate class? I like the idea of condensing the number of featurizers, but am not sure how much additional complexity this suggestion would add.

ardunn:
Yes, I did consider that. 

1. Mainly it would constitute adding a fit method to both *CM classes, coupled with weird logic that the object should only fit when user passes in `flatten=True` to `__init__` or something -- you don't need to fit if you want the matrix form. Objects that only need to be fit conditionally are confusing. 

Counterpoint to 1: It is not like the matrix form of these gets too much use, AFAIK

2.  There is already another method of flattening *CoulombMatrix, BagofBonds. Having an intrinsic flattening of CoulombMatrix would require changing that class (i.e., BagofBonds can't work with a CoulombMatrix flattened already) + there would be multiple ways to flatten a coulomb matrix (args to *CM init vs. passing in *CM object to BagofBonds)

Counterpoint to 2: I wouldn't be totally against changing or removing BoB.

3. Code duplication.

WardLT:
Thanks for walking me through your thinking! I'm persuaded. 

In that case, this PR looks good to me.

ardunn:
Actually, perhaps it would actually be better to just remove the ability to get matrix form, and also deprecate/remove BagofBonds? Then we can condense/remove two featurizers 

Reason being is that BoB is insanely slow and not very good (informative) of a featurizer, and I'm not sure it gives much more info than the eigenvalue methods.

computron:
Regarding removing BoB - in general I think it's good to keep a published featurizer around even if it's not very good. Once you remove it, it's hard to show people that it's not good since the code is no longer there. We just received a note on the help list for people that want to use this (albeit for molecules).

The points you made in response to Logan's idea were not so strong IMO. I actually think the inconvenience of having to run a second featurizer (and know that it exists) is worse than understanding the various flags within ColoumbMatrix, especially if you document those flags well.

For example you could consider:
- default is flatten=True and the fit() method is required. Honestly, most people using these featurizers want this, and an easy route to getting it would be helpful.
- if they have flatten=True but didn't fit, then calling featurize() throws an error, telling them to either fit() first or set flatten=False
- if someone uses flatten=False, then if someone does call fit(), the fit function just passes through. So there is no penalty to calling fit() here, it just doesn't do anything
- if someone wants to use a downstream thing that requires the full Coulomb matrix, then they need to know to use flatten=False. I see this as a more complex use case for which it's reasonable to expect the user to know more. Any I still see this as being less complicated than having a separate featurizer for flattening the matrices.

ardunn:
@computron 

After actually trying BoB(SCM) vs. EigenSCM yesterday (using same models), BoB actually does "better" on a few different problems (using technically the same SCM information), although it does generate far more features and takes a little bit longer (after taking into account #368). So on second thought I agree, definitely should keep BoB

Re conditional fitting: sure, will be nice to have one less featurizer

ardunn:
notable updates:
- BoB has been changed accordingly
- *CM default behavior is flatten, Eigen*CM removed
- OFM default behavior changed to flatten, just for consistency with these other matrix methods.

ardunn:
@computron merge if this is more along the lines of what you were thinking

PR #373, opened 2019-02-15
Update to include CITRINATION_API_KEY

matSciMalcolm:
## Summary

Most tutorials and documentation for Citrination ask users to store their API key under CITRINATION_API_KEY as opposed to CITRINE_KEY. Adding a conditional tree to accept either should serve more users.

* Added conditional logic to look in os.environ for CITRINATION_API_KEY as well as CITRINE_KEY


computron:
Thanks! But I think this will raise a KeyError if CITRINATION_API_KEY is not before you even get to the CITRINE_KEY part

You'll likely need to do ```if os.environ.get("CITRINATION_API_KEY")``` for example

computron:
I will merge anyway and make the fix, thanks!

PR #375, opened 2019-02-28
Prepend names of data sources to ElementProperty entries

ardunn:
## Summary

Some ElementProperty presets have overlapping feature names (e.g., deml preset and matminer preset on boiling_point) which will cause featurizing with ElementProperty.from_preset("deml") and ElementProperty.from_preset("matminer") in succession to fail. 

Also, featurizing with multiple ElementProperty presets will cause potential confusion even if there is not a name conflict. For example, if you featurize with the matminer data source and then the magpie data source, you'll have "minimum X" and "minimum Electronegativity". Which feature comes from which data source?

Yes, you could just do `featurizer.feature_labels()` to find out which data source it came from, but this is just an extra step. We can have it so any user will be able to tell from the feature names.

This PR prepends the class names of the `data_source` to all statistics computed with that ElementProperty object. Here are some examples of the changes:

Before:
```
'mean matscholar_el_147'     # Matscholar embeddings
'range Electronegativity'    # Magpie
'range X'                    # matminer
'range electronegativity'    # Deml
```

After:
```
'MatscholarElementData mean embedding 147'
'MagpieData range Electronegativity'
'PymatgenData range X'
'DemlData range electronegativity'
```

### Classes affected
ElementProperty
CationProperty

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -49,6 +49,20 @@ class ElementProperty(BaseFeaturizer):
 
     To initialize quickly, use the from_preset() method.
 
+    Features: Based on the statistics of the data_source chosen, computed
+    by element stoichiometry. The format generally is:
+
+    "{data source} {statistic} {property}"
+
+    For example:
+
+    "PymetgenData range X"  # Range of electronegativity from Pymatgen data
+
+    For a list of all statistics, see the PropertyStats documentation; for a
+    list of all attributes available for a given data_source, see the
+    documentation for the data sources (e.g., PymatgenData, MagpieData,
+    MatscholarElementData, etc.).
+
     Args:
         data_source (AbstractData or str): source from which to retrieve
             element property data (or use str for preset: "pymatgen",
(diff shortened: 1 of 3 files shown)


ardunn:
The reason this is a PR and not just pushed to master is bc it will make the matminer_examples notebooks out of date, as the features will not be the same as in the notebook (the numbers will be, but the labels will not).

ardunn:
if no one (e.g. @WardLT @computron) has any issues with this I'll just merge

WardLT:
This looks good to me, with only one requested change: Could you add a translation of what each feature name means to the docstring of the various classes? These classes all predate any effort to enforce standards among the featurizer documentation.

ardunn:
@WardLT given that there are a lot of properties (i.e. docstring would be huge), some of which are not given in the presets, I figured it is probably better to just show the general format in ElementProperty and CationProperty, and leave the specifics of each data source to the data_source classes. I've raised a corresponding issue #379

WardLT:
Having a general formula for the names (e.g., "(source name) (stat name)
(property)") in the docs sounds like a good plan to me. I think I did
something similar in the "local property difference" featurizer in
structure.py

PR #376, opened 2019-03-02
Fixed references

kjappelbaum:
## Summary

* Fixed references to Ward PRB 2017 paper in the featurizers. 

## TODO (if any)

On a separate topic I wondered if you considered trying to use numba's @jiit on any of your functions. Some of the featurizers might see a speedup. Since a lot of the featurizers use a pymatgen structure object, it might not be straightforward in all cases, but e.g. the `get_distributions()` function does a lot of looping with python types which could probably be optimized with numba. If you're interested, I could do some profiling and testing.

### DIFF
--- a/matminer/featurizers/structure.py
+++ b/matminer/featurizers/structure.py
@@ -1995,7 +1995,7 @@ class StructuralHeterogeneity(BaseFeaturizer):
             Divided by the mean Voronoi cell volume.
 
     References:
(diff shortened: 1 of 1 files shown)


ardunn:
Test was failing because of circleci outage, seems to be fixed now so I'll merge.

I don't think we have considered that! I will look into Numba for possibly speeding things up. I've raised an issue #377 so we don't forget about it.

If you have an idea for an implementation already though @kjappelbaum, please make a PR!

Issue #368, opened 2019-02-11
featurize_dataframe inplace is extremely slow for featurizers w/ lots (1k+) features


ardunn:
### problem
In featurize_dataframe, there is this section of code:

```python
        if inplace:
            # Update the existing dataframe
            for k in labels:
                df[k] = res[k]
            return df
        else:
            # Create new dataframe and ensure columns are ordered properly
            new = pd.concat([df, res], axis=1)
            return new[df.columns.tolist() + res.columns.tolist()]
```

When inplace is True, pandas goes about appending a new column for each feature. This is a problem for featurizers which return lots of features, because internally in pandas some lists are being appended to for each feature - appending to native Python lists is amortized O(1) (approximately O(n) non-amortized, but observed significantly worse), so these operations can be surprisingly slow (hours, or **much** longer than the featurization time!).

Changing the inplace code to `df[labels] = res[labels]` has the same problem.

## solution
The inplace=False option may also be amortized O(1), but it is significantly faster. For example, I featurized ~1.1k samples with 60k features:

```
inplace = True
elapsed time (s): 538.517

inplace = False
elapsed time (s): 3.16
```
Note that these times are *just for assigning the columns to the df, not featurization*. Featurization itself took less than 10s. 

The inplace = False time can be lowered further by not reordering the columns:
```python
        #else:
           # Create new dataframe and ensure columns are ordered properly
            #new = pd.concat([df, res], axis=1)
            #return new[df.columns.tolist() + res.columns.tolist()]
        else:
            return pd.concat([df, res], axis=1)
```

```
inplace = False
elapsed time (s): 0.58
```

For some featurizers, this inplace operation is prohibitively expensive, even for small datasets. In the extreme, you have something like BagofBonds which can generate 150k+ features on a 1.1k sample dataset in 30s, but will take _literally 12 hours_ to assign them to a dataframe inplace. Enabling inplace=False assigns to the df in 0.8s. 

## proposed change
For these speed reasons, I suggest we make the default behavior of `featurize_dataframe` inplace=False, cleaning up the code a bit in the process.

WardLT:
The proposed changes sound great! Do you already have a break with these changes? 

I'm totally fine with `inplace=False` as a default

ardunn:
yeah unfortunately it breaks a lot of stuff, see #371

ardunn:
closed via #383

PR #381, opened 2019-03-06
WIP: Inplace is now false by default

samysspace:


samysspace:
If we set the inplace default value to False by default, we run into issues because the way featurize_dataframe() works at the current moment is that it modifies the Pandas table being passed in. However, when inplace is False, we create a whole new DataFrame and return it. However, the tests expect you to have modified the passed in dataframe instead of returning a new one. Due to the way frames and pointers work in Python, it's impossible to change the original variable (the one that is passed into the function) from inside the function if we are creating a whole new object and pointing our parameter to it. I fixed this by changing the tests and setting the dataframe that we pass into the function equal to the result of featurize_dataframe(), e.g. data = multi_f.featurize_dataframe(data, 'x') instead of just multi_f.featurize_dataframe(data, 'x') as we had before. This is why we were getting a bunch of KeyErrors and mismatched length values.

If you want to default to inPlace = False, it must be made clear to the user that you need to **set the dataframe equal to the result of the featurize_dataframe() function**. This is simply optional for the inPlace = True case since we are modifying in place and therefore no pointers get changed. I don't know if this is mergeable since it's a pretty easy fix, but people might have become accustomed to not reassigning their dataframes.

I am still failing two tests but these are because of some Pandas Dataframes issues with the tolist() function but those should be easy to fix. All of the tests that previously had KeyErrors/mismatched column length issues now pass.

ardunn:
We could just trash what we have now and make the "inplace" option the same as how pandas handles it. If inplace, do it inplace and return None. If not inplace, return the updated df. 

So for inplace=False
```
df = some_df
df_featurized = some_featurizer.featurize_dataframe(df, "structure", inplace=False)
# df is still the original some_df and df_featurized is a featurized copy
```

And for inplace=True
```
df = some_df
df_featurized = some_featurizer.featurize_dataframe(df, "structure", inplace=True)
# df is the featurized dataframe, df_featurized is None
```

ardunn:
Also re: reassigning their dataframes

Idk about anyone else, but I always return the df, even when I had inplace=True
```
df = ft.featurize_dataframe(df, "comp")
```

Never just:
```
 ft.featurize_dataframe(df, "comp")
```

I think most of the examples in the notebook are similar to this. So I wouldn't worry too much about it

Just make sure nothing funky is going on with the index!!

computron:
To me it’s clearer if inplace = T does *not* return the data frame. I think this is what pandas does by default. But not a big deal either way IMO.

ardunn:
@computron I agree this should be the behavior (See first comment). Samy any update on finishing this PR? It appears only one test is failing right now so seems pretty close.

Also @samysspace if there is not already a test for inplace=True, could you add it?

samysspace:
Currently working on trying to get that last error fixed. I'll also add in a test for in place = True

ardunn:
Closes #368

matminer/featurizers/base.py
```diff
@@ -260,13 +260,13 @@ def featurize_dataframe(self, df, col_id, ignore_errors=False,
...
-            for k in labels:
-                df[k] = res[k]
+            df[labels] = res[labels]
             return df
         else:
             # Create new dataframe and ensure columns are ordered properly
             new = pd.concat([df, res], axis=1)
-            return new[df.columns.tolist() + res.columns.tolist()]
+            df = new[df.columns.tolist() + res.columns.tolist()]
+            return new
```
ardunn: seems like `df` should be returned, not `new`?

matminer/featurizers/tests/test_conversions.py
```diff
@@ -108,18 +108,27 @@ def test_structure_to_oxidstructure(self):
...
 
         # original is preserved
+        print(type(df["structure"].tolist()[0][0]))
         self.assertEqual(df["structure"].tolist()[0][0].specie, Element("Cl"))
 
         # test in-place
         sto = StructureToOxidStructure(target_col_id=None, overwrite_data=True)
         df = sto.featurize_dataframe(df, 'structure')
-        self.assertEqual(df["structure"].tolist()[0][0].specie.oxi_state, -1)
+        print(type(df["structure"]))
```
ardunn: in final pr please rm these print statements from the tests

ardunn:
@samysspace some things that might help here:

1. you can run the failing test locally in pycharm without having to run the entire test suite or use the CI. This is much faster and more efficient that pushing every commit to the CI to see if it passes, if the errors are not specific to the CI. --  To do this in Pycharm, go to the line for the test method (e.g., `def test_something(...)`), right click on it, and do "run unittest" - it should hopefully only run that method 

2. Not super important, but try to use somewhat informative commit msgs. This way if someone is trying to read through the commits you made, they can generally get a sense of what each commit is doing (even if it is very short), whereas "test" and "try" as commit messages are difficult to decode what ur doing at a glance

samysspace:
@ardunn Sorry about that, I couldn't figure out a way to run the unit tests locally so every time I made a small change I had to test after committing/pushing (which is why some of these commit messages are so short, usually because I'm just trying to test something that I will change back right after).

ardunn:
No problem, but can you run individual tests locally now? Your commits should be as minimal as possible (i.e., without changing something and changing it right back, otherwise it is hard to follow). 


BTW to run the whole test suite locally, do `python setup.py test` from within your environment in a terminal. For running individual tests I usually use Pycharm. For more info see the pycharm docs!

https://www.jetbrains.com/help/pycharm/performing-tests.html

samysspace:
Yep, sounds good (sorry for the extensive commit history).

PR #382, opened 2019-03-18
Add MEGNet 1Neuron embeddings

ardunn:
## Summary

MEGNet (https://github.com/materialsvirtuallab/megnet) recently added a set of element embeddings from training a 1-MEGNet-layer NN on the formation energy from 60k crystals on MP. They use these embeddings in their paper to help with transfer learning on smaller datasets.  I tried using these embeddings as simple composition features, turns out they perform similarly to the existing ElementProperty presets across a range of problems.

This PR adds these embeddings as an ElementProperty preset (in similar fashion to the MatScholar embeddings)

* Added MEGNet element embeddings (Z of 1-94, plus one dummy) from megnet repo json file
* The embeddings are handled by the `MEGNetElementData` and can be used as weighted features via `ElementProperty` preset `megnet_el`
* Added appropriate repo and arXiv citations, with TODOs for updating with MEGNet publication once it is published
* Added a test for MatScholarElementData which should have been in there but wasn't

Btw I have written permission from Chi/Shyue that it is ok to use these embeddings


computron:
Thanks!

PR #371, opened 2019-02-12
(WIP) change base default to inplace=False

ardunn:
This is a WIP that I'll just merge once tests pass

PR #383, opened 2019-03-20
Fix inplace

ardunn:
## Summary

* **Changes inplace behavior of featurize_dataframe to be like pandas: original obj in mem is updated and None is returned if True, else original obj remains the same and the updated obj is returned**
* **Changes default behavior of featurize_dataframe to inplace=False (pandas default)**
* Re-enabled a test that was previously commented out for unknown reasons (`test_multifeatures`) and added a case to test multiple featurizers with multiple arguments and multiple indices
* Added more tests for conversion featurizers
* Fixed a bug in BaseFeaturizer that caused the first letter of the first returned feature label to be the outer index name when using MultipleFeaturizer
* A few super small formatting fixes in `structure.py`
* Changes to several tests to accommodate this change

**THIS PR WILL BE MERGED WHEN TESTS PASS**


ardunn:
Closes #381 
Closes #368 
Closes #371

PR #384, opened 2019-03-25
New: prechecking

ardunn:
## Summary

Occasionally, it can be helpful to know if the entries in your dataset are within the scope of a featurizer you are using _before_ applying the featurizer. Say you have a relatively expensive featurizer (matminer contains many) and some dataset where 50% of the entries are within the scope of that featurizer; at best, 50% of the entries returned are nan if you did `ignore_errors=True`, and you spent a bunch of time applying an expensive featurizer to half of your dataset; at worst, that featurizer returns some numbers despite being essentially out of scope and you have no idea your features are bogus. 

This PR introduces a simple `precheck` method to `BaseFeaturizer` which can be optionally inherited by subclasses of `BaseFeaturizer`. You can pass in an entry (e.g., Composition) and `featurizer.precheck(compositions)` will return whether that entry passes the precheck or not. Also, there is a `precheck_dataframe` which works on dataframes (probably the form most people will be using tbh). If you want to add precheck functionality to a featurizer, add you need to do is override the precheck (i.e., pre-check a single entry) method of the inheriting class.

The idea here is that you can use `precheck_dataframe` generally to prescreen a dataframe before undergoing a lengthy (or error-prone) featurization operation. Things worth prescreening a list of entries might include whether entries
* will take forever
* will often have a featurizer return nan or throw errors 
* will be out of scope for a featurizer 

## Changes
* Added a `precheck` method to BaseFeaturizer, which by default returns True (meaning for any featurizer that doesn't override `precheck`, all entries will be appear valid). Also adds `precheck_dataframe`, which checks the entire dataframe and either returns the df or returns the fraction of True (passed `precheck` entries). 

Current featurizer implementations of `precheck` are:
* `Miedema` - only valid compositions are those having _all_ elemental empirical parameters defined (+ updated tests)
* `YangSolidSolution` - essentially the same as Miedema (+ updated tests)
* `DensityFeatures` - only ordered entries are valid (+ updated tests)

## Future work
Of course changing the above three featurizers above is incomplete application of `precheck`. Some other possible applications are:

* For featurizers relying on oxidation states (`CationProperty`, `OxidationStates`, `ElectronAffinity`, `ElectronegativityDiff`, `IonProperty`, etc.) all-metal entries' oxidation states are ill-defined, so they could be considered invalid
* For featurizers requiring nearest neighbors info, entries with huge numbers of sites could be considered invalid (e.g., 1000s of sites). Instead of marking as not passing precheck though, it could just be a good time to throw warnings.
* Probably many other possible uses in other featurizers


ardunn:
BTW shoutout to @Qi-max for originally implementing this in automatminer (albeit in much different fashion). 

Codacy errors are directly related to making sure BaseFeaturizer `valid_fraction` is readable while individual featurizer implementations are specific and minimal.

ardunn:
@computron updated to address ur concerns


Now with precheck (instead of `valid_fraction`) you can validate single samples, similar to BaseFeaturizer.featurize. With precheck_dataframe, you can validate an entire dataframe similar to BaseFeaturizer.featurize_dataframe.  You can return either 
* the fraction of passed prechecked entries (determined from boolean sum)
* or the dataframe with another column added (either inplace or returned as a copy, similar to featurize_dataframe)

So the behavior is now (example with Miedema):

```
>>>miedema.precheck(bad_composition)
False

>>>miedema.precheck(good_composition)
True

>>>miedema.precheck_dataframe(df, "composition", return_frac=False, inplace=False)
    composition  Miedema precheck pass
0      (Ti, Zr)                   True
1  (Mg, Cu, Ca)                   True
2       (Fe, O)                  False

>>>miedema.precheck_dataframe(df, "composition", return_frac=True)
0.66666
```


This removed most of the extra code from previous commits and made implementing for new featurizers much simpler and faster. Now, to add prechecking (including precheck_dataframe) to a featurizer, all you need to do is define the `precheck` method (i.e., determine whether a single entry is valid).

ardunn:
Considering I had already talked to @computron about it I may just merge but would be good to get opinions from at least one other before doing so

@utf @Qi-max any thoughts on current implementation?

matminer/featurizers/base.py
```diff
@@ -151,6 +155,84 @@ def set_chunksize(self, chunksize):
...
+                featurizer in on this dataframe.
+
+        By prechecking before featurizing, you can avoid applying featurizers
+        to data that will ultimately fail, return unreliable numbers, or
+        are out of scope. Prechecking is also a good time to throw/observe
+        warnings (such as long runtime warnings!).
+
+        Args:
+            df (pd.DataFrame): A dataframe
+            col (str): The column in df specifying the input to the featurizer.
```
WardLT: Can we allow this to be a list of strings, in the case of featurizers that take >1 argument as input?

matminer/featurizers/base.py
```diff
@@ -151,6 +155,84 @@ def set_chunksize(self, chunksize):
...
+        else:
+            precheck_col = "{} precheck pass".format(self.__class__.__name__)
+            if inplace:
+                df[precheck_col] = prechecks
+            else:
+                res = pd.DataFrame({precheck_col: prechecks})
+                df = pd.concat([df, res], axis=1)
+            return df
+
+    def precheck(self, x) -> bool:
```
WardLT: As above, should we change this to `*x` to allow for >1 inputs? 

PS. I like the use of typing! Yay for dropping 2.x support!

matminer/featurizers/composition.py
```diff
@@ -1391,6 +1409,35 @@ def __init__(self):
         # Load in a table of elemental properties
         self.elem_data = MagpieData()
 
+        element_list = [
```
WardLT: Would it make sense to move this list as a class attribute of MagpieData?

WardLT:
I think this is a great idea. 

I don't see `basic_composition_stats` or other other "prefeautization" utils. Are those TBD or am I missing something?

matminer/featurizers/base.py
ardunn: Sure will do

matminer/featurizers/base.py
ardunn: Sure

matminer/featurizers/composition.py
ardunn: Well this list is just a Yang-specific subset of elements in MagpieData. So it is not really specific to MagPie data, more specific to Yang. I mean we could add a .yang attr to MagpieData, but imo it is probably clearer being in the YangSS class itself.

WardLT: Sorry, should have read the code more. This seems like a better attribute to add to MixingEnthalpy? Maybe a "get_supported_elements" function? The MixingEnthalpy class is the one that seems to impose limits on which materials could be featurized.

ardunn:
> I think this is a great idea.
> 
> I don't see `basic_composition_stats` or other other "prefeautization" utils. Are those TBD or am I missing something?

Thanks! Oh, those were before I talked to Anubhav about it (they've since been remove, original PR comment is updated). They were no longer needed and this new code has way fewer SLOC

matminer/featurizers/composition.py
ardunn: Sure, added. Thanks @WardLT !

ardunn:
@WardLT ok, now should address all the review comments

Also I made it a point of **not** allowing the precheck to `ignore_errors` like `featurize_dataframe` does, since if your `precheck_dataframe` is failing, you probably will never want to ignore why (i.e., your compositions are strings, you're feeding a number into a structure featurizer, etc.).

WardLT:
Looks good to me!

Qi-max:
The implementation looks good to me! It is really better to implement prechecking like this in matminer than my previous version in automatminer. Thanks!

ardunn:
Ok then if @computron is cool with it we can merge

PR #391, opened 2019-04-24
Change class name to cls() reference

clegaspi:
Fixes issue with inheritance. Say one creates a class that inherits from `RetrievalQuery` called `MyQuery`. With the current implementation, `MyQuery.from_pymongo(...)` would return a `RetrievalQuery` object instead of a `MyQuery` object. With `cls`, it will return the type of the calling class.


ardunn:
@clegaspi looks good to me

PR #392, opened 2019-04-26
Fix LaTex error in Sphinx doc

kmu:
### Fixed LaTex errors

- Escape sequences including `\b`, `\f`, and `\n` caused errors in LaTeX.
- This PR replaces them with double-backslash.
- Another option is to put `r` before comments like this: ```r"""  :math:`\frac{A}{B}` """```.

### Update resulting documentations


computron:
Thanks!

computron:
(note that in the future, you could just update the .rst files and we can recompile the docs on our end)

kmu:
Noted. Thanks.

PR #393, opened 2019-04-29
Add global instability index featurizer

mtdg-wagner:
## Summary

Adding a structure featurizer that computes the global instability index using bond valence methods. The International Union of Crystallography bond valence parameter tables are included in this PR along with some tests and the modified structure.py.

* GlobalInstabilityIndex structure featurizer
* IUCr 2016 bond valence parameter table

## TODO

For disordered structures, the code currently falls back on pymatgen's bond valence sum calculator. However, the adjustable length scale parameter can lead to unrealistic sums unless tuned carefully. It might be better to use an optimizer to pick this value such that the GII is minimized.


matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+    can be used instead. 
+    
+    Note that pymatgen's bond valence sum method is prone to error unless 
+    the correct scale factor is supplied. A scale factor based on testing 
+    with perovskites is used here.
+    TODO: Use scipy to optimize scale factor for minimizing GII
+    
+    Based on the following publication:
+    
+    "Structural characterization of R2BaCuO5 (R = Y, Lu, Yb, Tm, Er, Ho,
```
WardLT: Could you add this citation information to the `citations` function as well?

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
     def implementors(self):
         return ["Alex Dunn", "Lauri Himanen"]
 
+    
+class GlobalInstabilityIndex(BaseFeaturizer):
+    '''
```
WardLT: Could you change these to double rather than single quotes? (Python style docs recommend double)

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+                                               an_val=neighbor_val)
+                        bvs += self.compute_bv(params, dist)
+                    elif site_el == elements[-1] and neighbor_el != elements[-1]:
+                        params = self.get_bv_params(cation=neighbor_el,
+                                               anion=site_el,
+                                               cat_val=neighbor_val, 
+                                               an_val=site_val)
+                        bvs -= self.compute_bv(params, dist)
+                except:
+                    raise ValueError('BV parameters for {} with valence {} and {}{} not found in table'.format(
```
WardLT: Nice error messages!

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+        elements = [str(i) for i in struct.composition.element_composition.elements]
+        bond_valence_sums = []
+        cutoff = self.r_cut
+        
+        # for loop to calculate the BV sum on each site
+        for site in struct:
+            site_val = site.species.elements[0].oxi_state
+            site_el = str(site.species.element_composition.elements[0])
+            bvs = 0
+            neighbors = struct.get_neighbors(site, r=cutoff)
```
WardLT: There is a `get_all_neighbors` function for structure that could be a little faster. That said, I'm not sure how much faster and whether performance is currently a concern for your featurizer. So, regard this comment as optional.

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+        r_cut: Float, how far to search for neighbors when computing bond valences
+        disordered_pymatgen: Boolean, whether to fall back on pymatgen's bond 
+            valence sum method for disordered structures
+    '''
+    
+    def __init__(self, r_cut=4.0, disordered_pymatgen=False):
+        
+        basedir = os.path.dirname(os.path.realpath(__file__))
+        bv_path = os.path.join(basedir, "../utils/data_files/Bond_valences2016.csv")
+        self.bv_values = pd.read_csv(bv_path)
```
WardLT: These values seem like they'd be useful in other featurizers. Could you break them off into their own class? See [MixingEnthalpy](https://github.com/hackingmaterials/matminer/blob/4fc634830c032f12e477f4a9abe51a3411f4e99c/matminer/utils/data.py#L264) for an example.

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+    Args:
+        r_cut: Float, how far to search for neighbors when computing bond valences
+        disordered_pymatgen: Boolean, whether to fall back on pymatgen's bond 
+            valence sum method for disordered structures
+    '''
+    
+    def __init__(self, r_cut=4.0, disordered_pymatgen=False):
+        
+        basedir = os.path.dirname(os.path.realpath(__file__))
+        bv_path = os.path.join(basedir, "../utils/data_files/Bond_valences2016.csv")
```
WardLT: I'm concerned that using both a path with `/` and `os.path.join` will cause problems for Windows users. Could you break the path up into individual folders/paths (i.e., `(basedir, '..', 'utils', ...)`)?

matminer/featurizers/structure.py
```diff
@@ -3612,3 +3613,189 @@ def citations(self):
...
+    Based on the following publication:
+    
+    "Structural characterization of R2BaCuO5 (R = Y, Lu, Yb, Tm, Er, Ho, 
+        Dy, Gd, Eu and Sm) oxides by X-ray and neutron diffraction", 
+        A.Salinas-Sanchez, J.L.Garcia-Muñoz, J.Rodriguez-Carvajal,
+        R.Saez-Puche, and J.L.Martinez, Journal of Solid State Chemistry,
+        100, 201-211 (1992),
+        https://doi.org/10.1016/0022-4596(92)90094-C
+        
+    Args:
```
WardLT: Could you add a "Features:" block to the documentation that describes the feature being generated. See [PRDF](https://github.com/hackingmaterials/matminer/blob/4fc634830c032f12e477f4a9abe51a3411f4e99c/matminer/featurizers/structure.py#L281) as an example.

matminer/featurizers/structure.py
mtdg-wagner: Thanks!

matminer/featurizers/structure.py
mtdg-wagner: Thanks for letting me know. I'm going to leave as is for now to avoid refactoring the whole thing, but if performance is noticeably bad, I'll switch.

mtdg-wagner:
Do the commits from today address your comments @WardLT?

WardLT:
I think it's ready to merge. Anyone else want to look it over?

ardunn:
Looks good by me.

ardunn:
Thanks for contributing @mtdg-wagner !

Issue #394, opened 2019-05-04
GlobalInstabilityIndex precheck is highly inaccurate (too optimistic)


ardunn:
Example code:

```python
df = load_dataset("castelli_perovskites")

gii = GlobalInstabilityIndex()
sto = StructureToOxidStructure(target_col_id="structure", overwrite_data=True)

df = df[["structure", "e_form"]]

# Assign oxidation states
df = sto.featurize_dataframe(df, "structure", ignore_errors=True)

# The expected (prechecked) non-nan fraction
print("Expected non-nan fraction:", gii.precheck_dataframe(df, "structure"))

# The actual fraction of non-nan entries
df = gii.featurize_dataframe(df, "structure", ignore_errors=True)
print("Actual non-nan fraction:", 1 - df["global instability index"].isna().sum()/df.shape[0])
```

Out:
```
Expected non-nan fraction: 0.8998309382924767
Actual non-nan fraction: 0.022400676246830065
```

The precheck says the expected non-nan fraction will be nearly 90%; in reality, 98% of samples fail. For other datasets , the difference is less egregious but still present:

```
# For dielectric_constant
Expected non-nan fraction: 0.9024621212121212
Actual non-nan fraction: 0.5909090909090908
```

Prechecking of course will not be perfect but should be within ~10% or so generally of the actual failing percentage. This should be addressed in this featurizer's precheck somehow.

@mtdg-wagner any ideas on how we could address this?

mtdg-wagner:
The precheck is only screening for the presence of non-zero oxidation states in the structure. It appears the majority of featurization failures are due to species pairs not being found in the bond valence parameters table. There are also some structures with non-integer valences, which are not tabulated at all and some cases were the output GII is higher than 0.6 (indicating the parameters where found but might not be a good use for the particular case).

The prevalence is:

- 15548 missing parameters
- 1866 non-integer valences
- 1090 unusually large GII

I'm thinking I could change the precheck to looking up if the parameters exist for one pair of sites as a tighter criterion,

PR #395, opened 2019-05-05
Make precheck of GII more in line with featurization errors

mtdg-wagner:
Now checks each cation-anion pair in formula for existing bond valence parameters in table.
Should address discrepancy noticed in issue #394


ardunn:
@mtdg-wagner thanks for updating! I think there are some conflicting files but once those are fixed I will merge. What does prechecking the perovskites dataset give with this new prechecking routine?

Closes #394

mtdg-wagner:
The perovskites dataset now gives: 
```
Expected non-nan fraction: 0.23039940828402367
Actual non-nan fraction: 0.14264581572273882
```

ardunn:
Ok great thank you! Will close once GH says conflicts are resolved.

PR #396, opened 2019-05-07
Resolve bugs in IUCrBondValenceData

kmu:
## Summary

* `IUCrBondValenceData` is very useful but missed a line for execution. A test is added to ensure that the patch resolves this problem.
* `setup.py` is modified to include bvparam2016.cif in the package.
* URL for PRB 2017 is fixed.


ardunn:
Looks good to me. Thank you for contributing and correcting these errors @kmu! Much appreciated.

PR #398, opened 2019-05-10
polish miedema docs and typo fix

Qi-max:
## Summary

* polish Miedema docs: add the location of data file, add units of returned formation enthalpies
* set the default structure types of Miedema to "all" (including "inter", "amor" and "ss" instead of only "inter")
* fix a typo in cgcnn SGD Optimizer


ardunn:
@Qi-max if you could include the precheck change I will merge!

Also there does not seem to be any CGCNN update?

Qi-max:
@ardunn sorry I just find the comment here. I have submitted a separate PR for precheck. The CGCNN update is in the third commit (replace Adam with SGD).

ardunn:
Great thanks!

PR #399, opened 2019-05-14
fix precheck for composition objects with oxidation states

Qi-max:
## Summary

The current `precheck` for Miedema and YangSolidSolution Featurizers will naturally fail for composition objects with oxidation states which causes a problem in automatminer featurization. This PR fixes this issue. 

Tests are also added to make sure `precheck` + `featurization` work equally for compositions with or without oxidation states.


ardunn:
Thanks

PR #402, opened 2019-06-07
Some optimizations for global instability index featurizer

mtdg-wagner:
## Summary

Some changes to global instability index featurizer to speed it up

* Caches bond valence differences for Wyckoff sites
* Caches bond valence parameters after Pandas query
* Pre-computes entire neighbor list rather than for each site in iteration


WardLT:
Thanks, Nick! Minus the codacy issue (whitespace that should be deleted), this looks OK to merge to me. 

I do have a few concerns about the caching of bv parameters:

- Is caching in a dictionary actually faster than looking up from the bond valence table? 
- If caching in a dictionary is indeed faster, have you considered using `lru_cache` from functools to permit caching between different invocations of `featurize`?

mtdg-wagner:
Yes, it appears caching the bond valence values improves the speed. I ran a sample of 85 titanates where I compared the original code against a version with the get_all_neighbors method and site symmetry caching and a version with the bond valence parameter caching.
- Original: 66.66 seconds
- Symmetry and neighbor caching, no bond valence caching: 23.17 seconds
- All caching: 14.84 seconds

I hadn't heard of the lru_cache decorator before. I would think it would be best applied to the get_bv_params function, but would that require a recursive function call?

WardLT:
Wow, didn't expect that large of a benefit from dataframe querying. Good catch!

The `lru_cache` decorator is pretty similar to the dictionary-based approach you use for caching already (i.e., checking a lookup-table for the result, returning value if found, ...). The difference is that it drops entries from the cache if the cache gets too large. It doesn't require any recursive function calls, and only decorating the `get_bv_params` function with a single line of code ([example](https://docs.python.org/3/library/functools.html#functools.lru_cache), in case you haven't seen it)

If you have time to explore the change, I'd recommend testing out the use of `lru_cache` over your currently-used custom version of it.

WardLT:
Looks good to me. Are you testing whether using lru_cache gives you the same performance benefits?

mtdg-wagner:
Yeah, I'm running a test now. Should have results in a sec

mtdg-wagner:
375 titanate test set.
Without lru_cache: 131.65
With lru_cache: 73.17
Probably because my code from two commits ago didn't reuse the dictionary across calls to calc_gii_iucr

mtdg-wagner:
Thanks for the pointer about lru_cache. Great functionality!

WardLT:
No problem! Are you ready for this to be merged?

mtdg-wagner:
Yep, good to me

WardLT:
Thanks!

PR #404, opened 2019-08-01
Fixes so matminer works with updated PMG

ardunn:
## Summary

1. Update requirements to use new pymatgen
2. Update DopingFermi to play nicely with new pymatgen

If tests pass i will just merge

### DIFF
--- a/matminer/featurizers/tests/test_structure.py
+++ b/matminer/featurizers/tests/test_structure.py
@@ -863,6 +863,8 @@ class StructureFeaturesTest(PymatgenTest):
         self.assertTupleEqual(df.shape, (3, 451))
         self.assertAlmostEqual(df["SOAP_449"].iloc[1], 3.029167413711548, places=5)
 
+    # todo: remove skip once https://github.com/materialsproject/pymatgen/pull/1548 is versioned into PMG
+    @unittest.skip()
     def test_GlobalInstabilityIndex(self):
         # Test diamond and ni3al fail precheck
         gii = GlobalInstabilityIndex(r_cut=4.0, disordered_pymatgen=False)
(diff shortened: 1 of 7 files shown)


ardunn:
Well this is not good, updating to new pymatgen fails a bunch of tests...

computron:
Do you know what the underlying issue is with the PMG update?

ardunn:
@computron Seems to be with `Structure` `get_neighbors` which is commonly used among our structure featurizers. For example:

```python
sites, dists = zip(*struct.get_neighbors(my_site, self.cutoff))
ValueError: too many values to unpack (expected 2)
```

Will fix this week

ardunn:
@computron Fixed

The test for global instability index is skipped at present until new pymatgen version is released including the changes in [this PR](https://github.com/materialsproject/pymatgen/pull/1548).

matminer/featurizers/tests/test_structure.py
```diff
@@ -864,7 +864,8 @@ def test_SOAP(self):
         self.assertAlmostEqual(df["SOAP_449"].iloc[1], 3.029167413711548, places=5)
 
     # todo: remove skip once https://github.com/materialsproject/pymatgen/pull/1548 is versioned into PMG
-    @unittest.skip()
+    @unittest.skip(reason="Pymatgen bv_sum interally broken, will be fixed "
+                          "in next PMG release")
```
computron: I would change this in two ways:

1. Write the specific PMG release in which this is expected to be fixed
2. Instead of skipping, uses a @unittest.expectedFailure decorator. Thus, when the problem is fixed, we will be forced to remove the decorator in order to get the tests to pass (otherwise, we might forget to remove the skip test).

computron:
I just put a comment on one of your commits, see above

matminer/featurizers/tests/test_structure.py
ardunn: Perfect, thanks. Addressed in 32f6224ff8bf438cb9af20b1c0d53fb32163ebb7

PR #405, opened 2019-08-01
Added Meredig composition featurizer, using features defined in 10.11…

AmalieT:
## Summary
Added "Meredig" as a new composition featurizer, using features as defined in 10.1103/PhysRevB.89.094104


ardunn:
Hey @AmalieT thanks for the contribution!

I have a few questions/requests before merging:

1. Is there a reason why we could not put Meredig featurizer as a preset of the ElementProperty class? It seems (as least just from skimming the code) it could fit into that paradigm.

2. If it really does warrant its own featurizer, the `Meredig` class needs more documentation. Please see the "Documenting a BaseFeaturizer" section of the [BaseFeaturizer documentation](https://github.com/hackingmaterials/matminer/blob/master/matminer/featurizers/base.py).

AmalieT:
Hey Alex,

In my view Meredig features didn't quite fit naturally into the ElementProperty class for a couple of reasons;

1. ElementProperty is set up to take a list of properties, and a list of statistics (e.g. mean, range etc.), then calculate every statistic on every property. As it was implemented in the original paper, the Meredig features calculate different statistics on each property. My view was that adding the capability to ElementProperty to calculate statistics only on certain properties would add too much complexity to what is currently a very compact and readable class.

2. Certain features in the Meredig featurizer depend on multiple properties (e.g. element fractions, fraction of valence electrons in s orbitals), which is again, currently unsupported by ElementProperty, and would require adding a reasonable amount of complexity.

Happy to integrate it in if you think differently though!

WardLT:
Hello @AmalieT , 

Thanks for the contribution!

I'll second Alex's comment about needing to add the documentation to this class. 

I am OK with having Meredig featurizers as a separate class, but there are a few places we can eliminate. 

I'll do a quick code review and mark these places that need changes.

matminer/featurizers/composition.py
```diff
@@ -254,6 +255,83 @@ def citations(self):
     def implementors(self):
         return ["Jiming Chen", "Logan Ward", "Anubhav Jain", "Alex Dunn"]
 
+class Meredig(BaseFeaturizer):
+    """
+    Class to calculate features as defined in Meredig et. al.
```
WardLT: We should expand on the documentation here, following the style described in the link Alex pointed out. E.g., what are the names of the features generated by this class.

matminer/featurizers/composition.py
```diff
@@ -254,6 +255,83 @@ def citations(self):
...
+
+        Args:
+            comp: Pymatgen composition object
+
+        Returns:
+            all_attributes: Specified property statistics of features
+        """
+
+        #First 103 features are element fractions
+        n_elements = 103
```
WardLT: You should make a call out to the ElementFraction featurizer rather than implementing your own version here.

matminer/featurizers/composition.py
```diff
@@ -254,6 +255,83 @@ def citations(self):
...
 
+class Meredig(BaseFeaturizer):
+    """
+    Class to calculate features as defined in Meredig et. al.
+    """
+
+    def __init__(self):
+        self.data_source = MeredigData()
+
+        self.features = [e.strip()+" fraction" for e in self.data_source.element_names]
```
WardLT: As "self.features" is only used internally by the featurizer, you should make it a protected variable. I.e., change the name from `self.features` to `self._features`.

matminer/featurizers/composition.py
```diff
@@ -254,6 +255,83 @@ def citations(self):
...
+            atomic_number_i = obj[0].number - 1
+            element_fractions[atomic_number_i] = obj[1]
+
+        #Calculate the remaining features
+        # Get the element names and fractions
+        elements, fractions = zip(*comp.element_composition.items())
+        attributes = [0] * (len(self.features) - n_elements)
+
+        #Total valence electrons, stored so we can calculate orbital fractions
+        valence_sum = 0
```
WardLT: These features are the same as those computed by the [`ValenceOrbital`](https://hackingmaterials.lbl.gov/matminer/matminer.featurizers.html#matminer.featurizers.composition.ValenceOrbital) featurizer. Please make a call out to the existing featurizer rather than making duplicate code.

matminer/featurizers/composition.py
```diff
@@ -254,6 +255,83 @@ def citations(self):
...
+        for i,feat in enumerate(self.features[n_elements:]):
```
WardLT: Suggestion, not required change: The code here is a little more complex than necessary. Rather than having a single loop that has different behavior on different iterations, consider splitting this into 2 or more loops. Further, determining which attribute to compute based on the name is also rather complicated. 

Because the order of the attributes does not matter for this featurizer, a simpler/faster algorithm could be:

1. Loop over properties that are averaged, compute averages and add to list
2. Loop over "range" properties, compute/add to list
3. Call out to `ValenceOrbital` for the remaining properties

This would also allow you to skip splitting strings and the `calc_stat` operation.

matminer/utils/data.py
```diff
@@ -224,6 +224,45 @@ def get_elemental_property(self, elem, property_name):
     def get_oxidation_states(self, elem):
         return self.all_elemental_props["OxidationStates"][elem.symbol]
 
+class MeredigData(AbstractData):
```
WardLT: This module seems redundant with the MagpieData and PymatgenData. Unless I am mistaken, could you eliminate this class and use one of the existing databases instead?

AmalieT: My reasoning for adding a new module was that there was already significant overlap between MagpieData, PymatgenData, and DemlData, and so I had assumed that one featurizer -> one data retrieval class was a style choice. I can change to use one of the existing classes if this isn't the case?

WardLT: Ah, that makes sense. We do not have a requirement of one data class per featurizer, but do have many almost-exactly-similar classes to err on the side exactly reproducing the featurizers from those papers. Since the Meredig features use the same data as the Magpie ones, you could be get away with using MagpieData for your featurizers.

For full context, I would be happiest if we could merge all of the data classes into one but have not had the time to figure out how to do that effectively.

matminer/featurizers/composition.py
AmalieT: I expanded the documentation in the latest commit, let me know if it looks good

matminer/featurizers/composition.py
AmalieT: Changed

matminer/featurizers/composition.py
AmalieT: Changed

matminer/featurizers/composition.py
AmalieT: Changed

matminer/featurizers/composition.py
AmalieT: I split out the valence orbital calculations, and the loop is now only 4 lines.

For the moment it's still calculating statistics/attributes based on string splitting - do you think in the new form that's still too complex?

matminer/utils/data.py
AmalieT: Changed to use Magpie data source

AmalieT:
Awesome, thanks @WardLT ! :)

PR #409, opened 2019-08-19
Add bond valence parameters for soft anions, split off bond valence sum function

mtdg-wagner:
## Summary

Following a suggestion in the original data source of the bond valence parameters, I added a default option to use the values specified for 9+ oxidation states in the original table for all other oxidation states of softer anions such as S, I, Br, N, Se, etc. This has the effect of expanding the pool of parameterized cation-anion pairs.

I also broke out the bond valence sum calculation into a separate function for those interested in checking differences between formal and calculated valence states rather than just the global instability index.

* More bond valence parameters
* Break out bond valence sum calculation into own function.


WardLT:
@ardunn , should I squash this PR into a single commit since we have a few small commits? What are your thoughts on squashing?

ardunn:
@WardLT I don't really think it matters for something small like this (but I'm far from a git expert). Up to you.

To squash or not to squash, that is the question

WardLT:
Personally, I like to use lots of small commits. Merge as is!

PR #412, opened 2019-09-10
Updating pymatgen requirement to see if any tests break

ardunn:
* Updated PMG and monty to newest versions
* Regenerated util files broken by newest monty
* Removed expectedfailure from GlobalInstabilityIndex (GIIX) test
* Fixed GIIX test broken for other reasons (positional arg error)
* Make GIIX precheck more robust 

@mtdg-wagner it would be great if you could give the GIIX and GIIX test a quick once over to double check everything is working as intended, as the GIIX test has been broken for so long


computron:
@ardunn feel free to merge yourself when ready

PR #416, opened 2019-10-08
Update Google Groups references to Discourse 

utf:
The Matminer support forum is moving from Google Groups to Discourse (https://hackingmaterials.discourse.group/c/matminer). I've updated all references to Google Groups to point to the new forum.

Once merged, the docs should be regenerated to reflect the updates.


ardunn:
@computron just merging so we can release a new matminer version (important for new automatminer updates)

PR #418, opened 2019-10-11
update sklearn requirement

ardunn:
will merge if passes tests

fixes some downstream Automatminer issues with cloudpickle which are fixed in newer releases of sklearn


ardunn:
appears to be circleci-specific dependency problem

Issue #420, opened 2019-10-14
SOAP implementation details need to be checked


ardunn:
Related to https://github.com/SINGROUP/dscribe/issues/29

@lauri-codes

The class can be found in `matminer.featurizers.structure` as `SOAP`.

lauri-codes:
I'm now doing some minor changes to the Featurizer and I will open a pull request soon.

A couple of questions first:

1. What kind of SOAP default values do you want? Do you prefer speed or accuracy? The current ones (lmax=2, nmax=4, rcut=3.0, sigma=1.0) are fast but not so accurate in reproducing the environment.
2. The Featurizer currently only provides the average SOAP output (averaged over all sites in the structure). The full SOAP output that contains a vector for each atom would be more general, but I don't know how this fits the matminer-thinking. Do you only support a single vector per structure, or is it possible to output a 2D-array for each structure? Another possibility would be to use e.g. the [REMatch kernel](https://singroup.github.io/dscribe/tutorials/kernels.html#rematch-kernel) that still provides a single vector/structure.

computron:
Hi Lauri,

Thanks for getting involved! Some responses to your questions

1. The general policy we have followed so far is:
- The constructor doesn't specify defaults for the arguments when there are multiple "good" answers (i.e., we are not 80-90% certain that the user would want to use a single default).
- Instead, we add a method to the featurizer called "from_preset()", which takes in a String argument. Thus, your from_preset() might have possible strings called "fast" or "accurate" that initialize the featurizer with the user-desired behavior
- If you search for "from_preset" in the codebase, you should see multiple places where this was done.

2. The featurize() method is intended to return only a single array (not a 2D array). There are a few different ways to address your concerns:
- The featurizers.site package contains featurizers that are intended to return an array per site (i.e., per atom), not an array per structure. Thus, anything where you are featurizing sites could go in that package.
- The featurizers.structure package should return an array per structure. You could also create an additional featurizer here that does the individual site descriptors and transforms to a 1D array using whatever algorithm you prefer. Note that if there are multiple algorithms to choose from, this can be a parameter in the constructor and might also be a candidate for "from_preset()" .

[...]

PR #428, opened 2019-10-15
Moved the SOAP featurizer from matminer.featurizers.structure to matm…

lauri-codes:
…iner.featurizers.site. Also improved the interface and added proper labels.

## Summary

* Moved the SOAP featurizer from matminer.featurizers.structure to matminer.featurizers.site. This makes sense since SOAP is inherently a local descriptor.
* If a user wants to describe a full structure, they should use e.g. SiteStatsFingerprint or roll their own custom kernel for comparing structures.
* Uses the latest DScribe interface 0.2.9.
* Support for almost all DScribe parameters (expect "sparse", since sparse matrices not supported).
* Updated SOAP regression test. The test is however quite primitive...

## TODO (if any)

* I'm not too familiar with how the matminer interface works, and what kind of tests should be performed. I only checked that the constructor, fit(), feature_labels(), and featurize() worked with very simple input. Could someone check that this implementation works as it should and that the code style is OK?

### DIFF
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -2288,10 +2288,10 @@ class SOAP(BaseFeaturizer):
         self.length = self.soap.get_number_of_features()
         return self
 
-    def featurize(self, s):
+    def featurize(self, struct, idx):
         self._check_fitted()
-        s_ase = self.adaptor.get_atoms(s)
-        return self.soap.create(s_ase).tolist()[0]
+        s_ase = self.adaptor.get_atoms(struct)
+        return self.soap.create(s_ase, positions=[idx]).tolist()[0]
 
     def feature_labels(self):
         self._check_fitted()
--- a/matminer/featurizers/site.py
+++ b/matminer/featurizers/site.py
@@ -2,6 +2,9 @@ from __future__ import division
 
 import copy
 from functools import lru_cache
+from monty.dev import requires
+from pymatgen.io.ase import AseAtomsAdaptor
+from sklearn.exceptions import NotFittedError
 
 from matminer.featurizers.utils.grdf import Gaussian, Histogram
 from matminer.utils.caching import get_nearest_neighbors
@@ -49,6 +52,13 @@ from pymatgen.analysis.chemenv.coordination_environments.chemenv_strategies \
 
 from matminer.featurizers.utils.stats import PropertyStats
 from sklearn.utils.validation import check_is_fitted
+
+# SOAPFeaturizer
+try:
+    import dscribe
+    from dscribe.descriptors import SOAP as SOAP_dscribe
+except ImportError:
+    dscribe, SOAP_dscribe = None, None
 
 cn_motif_op_params = {}
 with open(os.path.join(os.path.dirname(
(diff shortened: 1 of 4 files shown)


ardunn:
Hey @lauri-codes thanks for the PR! 

The current matminer implementation looks good to me. I have two higher level questions though:

1. What do you think is the best way of making the SOAP site featurizer a structure featurizer? We currently have a SiteStatsFeaturizer which takes statistics over the features created for each site (e.g., mean, so an example feature might be `"mean Z=11,Z'=14,l=3,n=1,n'=2"` averaged over all the sites in the structure. Do you see this as being a useful way to convert the site features into structure features?

2. We want to make sure the default implementation will give decent results for the intended purposes (e.g., predicting formation energies). Given the defaults for the featurizer are not specified, I could imagine it is hard for someone who is not familiar with SOAP to pick reasonable values for their problem. Do you have any recommended defaults or perhaps a heuristic for picking good values (apart from going through the heavy math on the original papers)?

Thanks!

ardunn:
closes #420

lauri-codes:
Hi @ardunn,

1. I cannot really say what the best way would be, as it depends on the ML method one is using. If you want to stay agnostic with respect to the ML method, averaging done with SiteStatsFeaturizer is perfectly reasonable. It is also what we did when using SOAP to predict formation energies.

2. The values that we used for the DScribe paper for formation energy prediction are available in the supplementary info. If you want, I can dig them up for you. I would hesitate to use them as default values, but providing them under the `from_preset`-function sounds good.

ardunn:
@utf what would you think about the following:

1. merging this PR
2. adding a from_preset SOAP option to SiteStatsFeaturizer
3. Adding a SOAP from_preset option from the default values @lauri-codes referenced

?
I can help do these things 

@lauri-codes if you have good presets for those values, that would be awesome! We could even have multiple from_preset methods based on use case. If you have 2-3 common ones that would be perfect.

utf:
@ardunn This plan sounds sensible to me.

ardunn:
Great! @lauri-codes if you can commit the parameters as `from_preset` methods here that would be ideal. Then I will go about merging this PR and adding a structural featurizer.

ardunn:
Hey @lauri-codes any update on the from_preset implementation? I'd be glad to add it myself but not sure which parameters would be best.

lauri-codes:
Sorry for being inactive on this. I can't right now add the functionality myself, but here are the presets we used in https://doi.org/10.1016/j.cpc.2019.106949):

Preset 1: Formation energy prediction for inorganic crystal structures:
- `r_cut`: 6 (Angstrom)
- `sigma`: 0.4
- `rbf`: "gto"
- `n_max`: 8
- `l_max`: 8
- `crossover`: True
- `periodic`: True

Presets 2-6: Charge prediction for small organic molecules (r_cut and sigma are different for each species): `rbf`: "gto", `n_max`: 8, `l_max`: 8, `crossover`: True, `periodic`: False
- C:
    - `r_cut`: 4 (Angstrom)
    - `sigma`: 0.1
- H:
    - `r_cut`: 8 (Angstrom)
    - `sigma`: 0.05
- O:
    - `r_cut`: 4 (Angstrom)
    - `sigma`: 0.05
- N:
    - `r_cut`: 4 (Angstrom)
    - `sigma`: 0.05
- F:
    - `r_cut`: 6 (Angstrom)
    - `sigma`: 0.1

ardunn:
Great, thanks @lauri-codes ! just saw your update, and very much appreciate your work on this. I will integrate these now.

PR #441, opened 2020-01-27
Update dimensionality

### LINKED ISSUE #440: matminers.featurizers.structure.Dimensionality uses a deprecated method
@utf Can you update with the latest and greatest?

utf:
## Summary

Update dimensionality featurizer to use the dimensionality finding algorithm detailed in https://journals.aps.org/prmaterials/abstract/10.1103/PhysRevMaterials.3.034003 and as implemented in pymatgen.

This algorithm requires a `StructureGraph` as input which can be generated using one the `NearNeighbor` classes implemented in the local_env module of Pymatgen.

The dimensionality featurizer defaults to using CrystalNN, however, the user can specify a different NN class if desired.

Fixes #440


utf:
The tests were failing due to issues unrelated to this PR.

I had a go fixing them and also updated the pymatgen requirement to the latest version.

PR #444, opened 2020-03-03
Updating notebook tqdm method to satisfy Warning Gods

brandonkrull:
## Summary

I've been using the `BaseFeaturizer` and appreciate its wide utility, really appreciate the work on this front.

This PR updates the notebook invocation of tqdm, removing the deprecated version in favor of the change expected in the next major release version of tqdm, so I don't have to keep seeing the deprecation warning whenever I use the `BaseFeaturizer` (or its subclasses).

* `tqdm.tqdm_notebook` -> `tqdm.notebook.tqdm`


WardLT:
Thanks for this PR and sorry for the delay in replying to you

Have you considered elimininating the `in_notebook` logic in matminer and using `from tqdm.auto import tqdm` instead?

ardunn:
I have no problem with either implementation. @brandonkrull if you want to use `auto` instead of `is_notebook` just go ahead and update this thread and we will merge it. Thanks!

brandonkrull:
thanks for the feedback!

`tqdm.auto` is definitely the way to go, so i updated that. 

i also looked around to see if/where `is_notebook()` is called and it was only used to determine which method of `tqdm` to use, so i nuked it. the method does appear in some of the docs, but presumably the docs are autogenerated? good to go other than that!

utf:
Thanks @brandonkrull. This looks great.

PR #452, opened 2020-05-01
Fix Magpie data

utf:
## Summary

This fixes Magpie data when using recent versions of pymatgen.

The issue seems to be that additional elements were added to the pymatgen periodic table for which magpie data hasn't been generated.

These new elements are now just ignored.


utf:
I seemed to have opened up a can of worms by updating the pymatgen version. I had to fix a bunch of new failing tests to do with the Structure `get_neighbors` method. Also, pymatgen requires an updated version of plotly, so I had to update the figrecipes module to change support from plotly v3 to v4.5.

ardunn:
@utf If you want, I am ok deleting figrecipes. I have been advocating for this for a long time. I don't think it is particularly useful or maintainable. Perhaps @WardLT or @computron have different thoughts on this...

utf:
Ok, this should be ready to merge. I think by updating all the project requirements to the latest versions this should fix: https://discuss.matsci.org/t/conda-install-broken/3427/10

Once this is merged can you push a new release to github/pypi.

utf:
I don't have an opinion on figrecipes. Maybe it could moved to a separate repository (`matminer-plot` or something)?

But as all tests are passing I think this PR can be merged.

ardunn:
Ok I think we need @computron to push a PyPi release but yes that sounds good.

If you're using conda, can you confirm this version fixes the conda install problems?

---

> I don't have an opinion on figrecipes. Maybe it could be moved to a separate repository (matminer-plot or something)?

My vote would be no, but I'm interested in hearing what other people have to say about thjis. Given that it doesn't have anything to do with this PR though, I'll stop discussing it here

Issue #378, opened 2019-03-02
External Libraries and Matminer


WardLT:
Issue #256 has caused me to continue thinking about our approach with building featurizers from external libraries. Do you think we should start breaking off featurizers that reference external libraries into their own submodules? 

So far, I think the answer is "no." Knowing the capabilities of the modules requires familiarity with the external libraries, and that outweighs the benefits of better recognizing external codes (our docs and citations methods handle this well) and any benefits from organizing (I'm starting to worry about the length of structure.py is interfering with maintainability). I also like that we use extra requirements for external libraries as it keeps matminer lighter weight, and brings further emphasis about the other codes (e.g., so people know about AMP and will judge if that code fits their needs better than matminer). 

I presume somebody else on this project has been mulling this over. Are external libraries worth discussing further at the moment?

ardunn:
I've been thinking about this too.

Probably the easiest thing is to do what you proposed: just add a external.py module and put everything that requires external featurization libraries to be in there.

Alternatively, maybe a long-term solution would be to have each featurizer type be its own folder/subpackage, rather than submodule. For example:
```
featurizers
  structure
    __init.py (would import all classes from modules so syntax of notebooks etc. would remain same)
    symmetry.py (GlobalSymmetryFeatures, etc.)
    distribution.py (RadialDistributionFunction, etc.)
    bonding.py (SiteStatsFingerprint, BondFractions)
    matrix.py (CM, SCM, OFM)
    external.py (CGCNN, SOAP, etc.)
    misc.py (EwaldEnergy, etc.)
    conversions (StructureToIStructure, StructuretoComposition, etc.)
```

We could do the same thing for composition features. Wouldn't be much point in doing it for BS or DOS since there are only like 2-5 featurizers for each one in there.

computron:
I like the idea of splitting up the featurizers as per @ardunn long-term solution

Shall we go for it?

Best,
Anubhav

utf:
I like the idea of splitting the featurizers up too. I'd even go further and have separate files for each external code (e.g. `cgcnn.py` etc). As otherwise this is liable to get just as messy if more external codes are added.

WardLT:
I agree with @utf about making a separate file for each external code

computron:
Think we are all in agreement then!

ardunn:
[...]

Issue #445, opened 2020-03-19
Add models


ardunn:
In addition to featurizers and datasets, it might be useful if this code could access models in the same way datasets are downloaded.

This would extend the usage of matminer as a general purpose "toolbox" for materials science. It would also make matminer a curated resource which would hold many valuable models otherwise spread out across random GH repos and figshare accounts.

For example:
```
from matminer.models import retrieve_model

m = retrieve_model("automatminer_KVRH_2020")
```

Or:
```
m = retrieve_model("CGCNN_2018")
```

Then:
```
my_predictions = m.predict(target="K_VRH", inputs=["composition", "structure"])
```

Or similar. Automatminer already has an infrastructure for `predict` syntax which can work with arbitrary ML pipelines (CGCNN, automatminer, regular single-model matminer, etc.).

Just teesting the idea and seeing what others think

CompRhys:
https://github.com/gomes-lab/H-CLMP is the link for the CLMP model

ardunn:
Thanks @CompRhys, I think RooSt will be a great addition to matminer. 


The way i am considering implementing this `models` module is something like this, for each external package we want to use:

```
- models/
    - roost/
        - tests/
        - main.py
        - requirements.txt
  ```


Each model has a directory.
* Main.py holds classes which implement simple methods for predicting/getting embeddings for every single model. 
    - something like `class RoostModel(BaseModel)`, which would implement methods for getting predictions, embeddings, and a list of the primitives the class was trained on
    - Multiple fitted models can be inside one class (e.g., `RoostModel(pretrained="e_form")`, or `RoostModel(pretrained="k_vrh")`, etc.)
    - the class would also have a method for fetching weights from matminer figshare and putting them into the model, using the pre-existing functions in the matminer.datasets module
* Requirements.txt 
    - Holds +git@commit or pip requirements needed to run the model
 * Tests must be included

CompRhys:
I've been thinking about this a little bit more and I think the most effective way to quickly add structure-based graphnets to matminer will be to wrap https://github.com/Open-Catalyst-Project/ocp/tree/master/ocpmodels/models - this way matminer would have access to more models for free as the OCP benchmarking suite is expanded over time. I would say the only issue you might have with this is that it neglects MegNET but you could have a separate wrapper or their package or perhaps the MegNET authors might want to add their model to the OCP benchmark. 

[...]

PR #470, opened 2020-05-11
Add featurizer based on Roost

CompRhys:
## Summary

A WIP PR that aims to add a featurizer based on the Roost (Representation Learning from Stoichiometry) model. (https://arxiv.org/abs/1910.00617)

## TODO 

- [x] Determine which parts of the Roost model to expose as features. 
- [x] Write a wrapper to Roost with `fit()` and `featurize()` functions.
- [ ] Add functionality to use pre-trained models
- [ ] Add functionality to fine-tune from pre-trained models
- [x] Test for regression
- [x] Test for classification
- [ ] Test for multiprocessing
- [ ] Document everything properly (make use of google docsting standard)

if needed:

- [ ] Test for pre-trained models
- [ ] Test for fine-tuning


ardunn:
Hey @CompRhys! First off, very excited for this contribution. 


### What features to expose:
I don't see any reason why we couldn't specify in the __init__ which features to expose. For example:


```
class RoostFeaturizer(BaseFeaturizer):
    def __init__(self, representation=True, prediction=False):
    """
    Args:
        representation (bool): Include the intermediate representation of the network as features.
        prediction (bool): Include the prediction of the network as a feature.
    """
    ....
```

This way you could choose one or the other, or both. I could imagine scenarios where having one or the other (or both) would be useful, it's probably best to give people the option.


### Fit and featurize methods

To me, these seem pretty straightforward. `fit` retrains the NN on a new dataset, and `featurize` returns features (whether hidden rep or prediction) on a stoichiometry. 

What would you think about also including a "`from_preset`" method (as is present in many other matminer featurizers)? The point of this could be to load a pre-trained NN such that you don't have to train it yourself. For example, usage like:

```
rf = RoostFeaturizer.from_preset("e_form")

mgbr3 = Composition("MgBr3")
prediction = rf.featurize(mgbr3)
```

The point being that no fitting is necessary if user is using a preset.

CompRhys:
### Features
I think that having access to pre-trained files is definitely helpful, I have some recent tests that show that starting from a pre-trained network helps with the optimization even if the pre-training task is quite different to the target task - this use would fall under the `warm_restarts` concept you have for cgcnn. A slight issue is that the model files are relatively large ~28Mb and so probably not a good idea to store them as part of any library - do you have a convenient place for such files in the matminer codebase. If these are to be saved then I agree it also makes sense to also expose them as pre-trained featurisers.   

### Tensorboard and Checkpoints
Currently, I have the model automatically write `tensorboard` summaries during training. Is this done anywhere else in `matminer`. Whilst I can reasonably dictate a sensible directory structure in my codebase I am not sure of the etiquette here. Similarly with the checkpoint files.

### Nice to haves
It might be also worth adding a `PyTorch` data set class that works nicely with the composition object in `matminer`. Currently, we read data from a .csv import a `pandas` dataframe and then the 'graphs' are cached, memory permitting, after they are called by the dataloader. We might be able to change it so that the data set class uses the `matminer` data frame. 

### Dependencies
On the front of installing Roost I recently updated the readme and it now provides much better descriptions about how to install the correct dependencies - `pytorch scatter` is the challenge here because you need to install the correct version based on your CUDA version. I am completely happy to make it pip installable but I am not sure if there is a good way to handle this CUDA issue and do that. 

The other issue is that while I am pretty happy with the state of the functional parts of the code it's still very much research code where I regularly make refactors in terms of the structure that break things - attempting to settle on a good structure will be part of this work. For instance, I have re-implementation `cgcnn` I have been playing around with that doesn't use zero padding have been trying to work out how to refactor such that I can share as much of the non-model code as possible..

WardLT:
Would hosting the models on Figshare be worthwhile? That is what's being done with our datasets, and solves the problem of storing large datasets needed by only a subset of matminer users.

Regarding dependencies, is PyTorch scatter available through conda? Perhaps one thing we could do is introduce an environment specification for installing the dependencies needed for Roost, since installation with Pip would be difficult. I have had good success with using conda environment YML files with other packages involving neural network libraries.

CompRhys:
Figshare sounds good if you have experience of using it, thinking long term I think that it's best to manage things like that on the side of `matminer` as I imagine it has a much longer roadmap than the remainder of my PhD!

Dependencies wise the author of `pytorch-geometric` of which `pytorch-scatter` is a modular library doesn't have conda releases for any of the parts, it's limited to pip3. There is always the option to default to the cpu-versions and then offering a warning or there may be something clever that can be done to extract the correct cuda version and download the correct library but I have very little experience of distributing software.

I've been working on some quite major refactors to the `roost` code that should make it easier to get it to work with `matminer` but I am currently stuck of a few issues. I'll have a deeper look at how to go about integrating it in once (if) I get around them.

CompRhys:
@ardunn Currently I use "robust losses" to get an estimate of the aleatoric uncertainty, as a result the model currently has two outputs one for the `expectation` and another for the `log_std`. Is this something that `automatminer` can be flexible with or do I need to work out how to abstract how this is set up such that "normal losses" can be used? The difficultly in finding a neat solution for this is the same reason as to why I haven't implemented classification out of the box yet (another potential point to raise).

ardunn:
@CompRhys at the moment, I'm still trying to figure out how Roost will figure into Automatminer.


### Losses/Featurization
Right now, I think the best thing is if the training and Roost Prediction is entirely internal to the featurizer. I.e., Automatminer won't _need_ access to the Roost losses for the learning stage at least... but it would be fine if those were also included in the features. In essence, all Roost needs to do is implement `.featurize` and Automatminer will take care of the rest.

### Model storage
Regarding storing the model and retrieving it, this might necessitate changing the matminer structure somewhat to accommodate (as we did when adding datasets) - see the issue I opened for this #445 

I will work on this and update the thread when it's ready (might not have time to work on it for another week or so). In the meantime though, don't worry about this, we'll figure it out in due time

### Roost refactors and installations
@CompRhys I have some experience setting up pip installations, I might be able to help with the Roost refactors for ease of pip installations. Feel free to tag me in that codebase as well!

Personally, I strongly favor a traditional pip-only (i.e., venv friendly) installation rather than a conda env install but if @WardLT has good experience with .yml files and you could get it working, that could be the way to go. The goal I think (from a matminer perspective) that roost could be listed in the `requirements-optional.txt` like `roost==1.0.0` and the installation "just works", but if that can't be done I'm open to other options

CompRhys:
## Update

So I spent most of the week refactoring `roost` to add in various features (the big change being bringing classification up to the same level at regression) and have added a `setup.py` to make it into a simple library that can be installed once downloaded from github.

As I view it in order to get it to work with `matminer` now it should be relatively easy as all that needs to be done is to replicate the functionality of the training sections of `examples/train.py` into some `matminer/featurizers/utils/roost.py` script most of which will be copy and paste. However, I won't be able to get round to this for a while as I have other projects that need some love having been neglected but I would be looking at trying to do it in the next few weeks.

@ardunn if you have some time to look at whether the library setup avoids the troubles you encountered before that would be quite helpful.

ardunn:
Hey Rhys,

I've managed to get the old version installed (not sure about the newest version, I'll report back!!) using _only_ pip without _too_ much trouble. 

Regarding the implementation here, it might be better if **all** the roost training parameters and code was located _within_ the Roost Featurizer, i.e., so calling RoostFeaturizer programmatically wouldn't require an external call to a command-line script in utils. This also makes debugging and maintenance easier as there is only one class or section of code to debug - RoostFeaturizer.

Regarding the timeline, no worries! I am also moving slowly on Automatminer at the moment as I've also had many projects. Feel free to ping me on this thread or over email when you start working on it again :)

CompRhys:
@ardunn I've made a first attempt that works if you have time to take a look.

In order to get it to play nicely I have removed some of the functionality I have in my main codebase (mainly to do with uncertainty estimation and resuming/fine tuning models). I have set it up so that it takes the composition object from str2composition as input and I take the embedding from the `MatscholarElementData` class that is already in `matminer` (this prevents the use of custom embedding but I feel that this is fine for the `matminer` version?)

I still need to add tests (and pass the code checks) but would appreciate some guidance on what sort of tests would be needed.

Functionality to perform "warm_starts" isn't currently implemented but can be added quite easily.

Here is the snippet I have been using for testing:
```
import pandas as pd
from matminer.featurizers.base import MultipleFeaturizer
from matminer.featurizers import composition as cf
from matminer.featurizers.conversions import StrToComposition

df = pd.read_csv("somedataset.csv")[:100]

# Convert the string for matminer to use
df = StrToComposition(target_col_id="composition_obj").featurize_dataframe(
    df, "composition")

roost = cf.RoostFeaturizer(task="regression", epochs=10)

roost.fit(df["composition_obj"][:80].values, df["target"][:80].values)

df_pred = roost.featurize_dataframe(df[80:], col_id="composition_obj")

print(df_pred)
```

CompRhys:
Just as an addition I also think that breaking down the composition and structure files into directories as described in #378 would be good. I feel having to split the code between `compositions.py` and `utils/roost.py` probably amplifies the technical debt of the Roost Featuriser

matminer/featurizers/composition.py
```diff
@@ -2004,3 +2025,211 @@ def get_ideal_radius_ratio(self, n_neighbors):
...
+
+        train_set = CompositionData(X, targets=y, task=self.task, emb=self.elem_emb)
+
+        train_generator = DataLoader(train_set, **self.data_params)
+
+        # if val_set is not None:
+        #     data_params.update({"batch_size": 16 * data_params["batch_size"]})
+        #     val_generator = DataLoader(self.val_set, **data_params)
+        # else:
+        #     val_generator = None
```
CompRhys, line 2155: Not sure what your workflow for handling validation sets is so left this out

matminer/featurizers/composition.py
```diff
@@ -2004,3 +2025,211 @@ def get_ideal_radius_ratio(self, n_neighbors):
...
+
+        test_set = CompositionData(comp, task=self.task)
+
+        test_generator = DataLoader(test_set, **self.data_params)
+        
+        # NOTE featurisation is faster on the cpu unless data is batched
+        # currently featurise_many doesn't use batches and instead applies
+        # the featuriser to each row in turn.
+        self.model.device = "cpu"
+        self.model.to("cpu")
```
CompRhys, line 2209: it's about 5x faster on CPU for me if we are running individual datapoints through the network

matminer/featurizers/composition.py
```diff
@@ -2004,3 +2025,211 @@ def get_ideal_radius_ratio(self, n_neighbors):
...
+        seed=42,
+        epochs=100,
+        log=True,
+        optim="AdamW",
+        learning_rate=3e-4,
+        momentum=0.9,
+        weight_decay=1e-6,
+        batch_size=128,
+        workers=0, # load data in main process -- allows caching
+        device=torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu"),
```
CompRhys, line 2052: This is one of the reasons we fail the tests. I am not sure how to work around this without #378

matminer/featurizers/utils/roost.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        self.comp = comp
+
+        if targets is not None:
+            self.targets = targets
+        else:
+            self.targets = np.full_like(comp, np.nan).ravel()
+
+        if emb == "matscholar":
+            self.elem_features = MatscholarElementData()
```
CompRhys, line 36: I can add the other feature embeddings I have as similar data classes if interested. https://github.com/CompRhys/element-embeddings

matminer/featurizers/utils/roost.py
```diff
@@ -0,0 +1,308 @@
...
+            raise NotImplementedError(
+                "Currently only the matscholar embedding is implemented"
+            )
+
+        self.task = task
+        if self.task == "regression":
+            self.n_targets = 1
+        elif self.task == "classification":
+            if targets is not None:
+                self.n_targets = np.max(self.targets) + 1
```
CompRhys, line 47: pytorch losses for classification take label encoded classes. I am not sure what the standard approach for `matminer` is, if it's one-hot then I need to add something to convert between the two

matminer/featurizers/utils/roost.py
```diff
@@ -0,0 +1,308 @@
...
+):
+
+    if fine_tune is not None:
+        print(f"Use material_nn and output_nn from '{fine_tune}' as a starting point")
+        checkpoint = torch.load(fine_tune, map_location=device)
+        model = Roost(**checkpoint["model_params"], device=device,)
+        model.load_state_dict(checkpoint["state_dict"])
+
+        if model.model_params["robust"] == True:
+            raise NotImplementedError("Robust losses not supported within Matminer")
```
CompRhys, line 239: left in to add `warm_restarts` although not yet configured. Will wait until a decision is made on where to store the default models.

matminer/utils/data.py
```diff
@@ -348,6 +348,9 @@ def __init__(self):
     def get_elemental_property(self, elem, property_name):
         return self.all_element_data[str(elem)][property_name]
 
+    def get_elemental_embedding(self, elem):
+        return list(self.all_element_data[str(elem)].values())
+
```
CompRhys, line 353: added this for ease as other solutions using the existing functions were messy inside the featuriser

ardunn:
Thanks for the update! I will look at it soon when I have a chance. 

However, splitting up the structure/composition classes into subpackages (as per #378) may introduce some problems to this PR unrelated to Roost which we'll have to fix later. I expect (1) example notebooks will be broken in a myriad of ways and (2) Automatic-documentation generation will likely be broken too. There are also downstream effects for Automatminer which will break things. 

From here, I think we have two options:

(1) Recommended: Revert the changes fixing #378 and just stick Roost at the end of structure.py (we can move things around later/in a separate PR). The objective should be to have a working Roost implementation with robust tests; exactly where the code resides at this point is not a big deal.

Or...

(2) Try to fix and update this PR to pass tests and also fix #378. I think this will require a bit more work from both of us. This will also probably necessitate a major release of matminer since so much of the code will be moved around.

CompRhys:
@ardnn all the tests in the full CI suite work locally. I am hopeful that it also won’t break any examples/code as a consequence of putting everything in `__init__.py`. I have no idea about documentation generation, I guess this might break it.

If #378 is the longer term intention for the software it seemed better to split out composition.py now as it meant I could keep all the roost code together. Hopefully making it easier to follow and less work maintain vs say the cgcnn implementation which has all those wrappers. This way I am only responsible for one file (+ tests) rather than multiple files. If it does break things I guess it’s not that bad to do it the other way but I would prefer splitting out composition.py if viable from your side.

Splitting structure.py was pretty uncessary but i did it because I had managed to link this PR to #378. If I was going to change things I thought it would be best to be consistent and change both rather than only changing composition.

ardunn:
Oh gotcha. In that case, would it be possible or desired to put all the external featurizers in `external.py` modules? e.g.,

```
from matminer.featurizers.composition.external import Roost
```

Just to avoid tons of individual modules (e.g., a separate module for cgcnn, etc.)

CompRhys:
I tried to group things based on concepts where possible and then gave the named/external things their own files. I guess the only risk of grouping is that eventually, you may end up with a 4000 line `external.py` (like `structure.py` now) as most new additions would go into `external.py`? 

Another issue with things that are truly external (need optional dependencies) is they may be easier to maintain if separated i.e. `roost` and another package might both need `torch` but the other package wouldn't need other dependencies of `roost` which might make complicate the `try: except:` constructions that are currently used to avoid errors with `dscribe` and `cgcnn` when the packages aren't present.

I am happy to do what you would prefer I guess the more important thing is have I broken the automatic documentation.

The remaining todo list would then be:
1. Unit tests 
1. Pretrained models/warm-restarts

ardunn:
That sounds like a good plan @CompRhys . I have fixed the CI issue for cohesive energy (I think the discrepancy may come from different pymatgen versions). Note I'm still in the process of reviewing this PR as is

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+        self.task = task
+        self.epochs = epochs
+
+        self.fitted = False
+
+        # NOTE multiprocessing doesn't play nicely with pytorch, apparently you
+        # can set the multiprocessing start method to spawn but it's safer just
+        # to only use one job?
+        # set_start_method('spawn', force=True)
+        self.set_n_jobs(1)
```
ardunn, line 151: Since most people (and automatminer) will `set_n_jobs` **after instantiation**, this won't function as intended - it will just be superceded. We should keep this in the `__init__`, but should also override the `BaseFeaturizer.set_n_jobs` method; e.g., 

```
class RoostFeaturizer(BaseFeaturizer):
    def __init__(...)
        ...
        self.set_n_jobs(1)

    def set_n_jobs(self, n_jobs):
        if n_jobs != 1:
            warnings.warn("Multiprocessing n_jobs invalid for pytorch. Please use <some option in __init__> to set the number of jobs. Setting n_jobs to 1...")
        self._n_jobs = 1
```

There should also be a test which directly tests this. I can see this causing a lot of problems in the future if someone tries to use n_jobs and it doesn't work as intended. It's one of those very sneaky things which can turn into a troubleshooting nightmare

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+
```
ardunn, line 152: Are there any compositions which this featurizer won't work for, and which can be checked easily?

E.g., if there is a composition with 15 elements, will roost still work?

If so, RoostFeaturizer needs a `precheck` method. See the `BaseFeaturizer.precheck` method for more details (reproduced here):

```
"""
        Precheck (provide an estimate of whether a featurizer will work or not)
        for a single entry (e.g., a single composition). If the entry fails the
        precheck, it will most likely fail featurization; if it passes, it is
        likely (but not guaranteed) to featurize correctly.

        Prechecks should be:
            * accurate (but can be good estimates rather than ground truth)
            * fast to evaluate
            * unlikely to be obsolete via changes in the featurizer in the near
                future

        This method should be overridden by any featurizer requiring its
        use, as by default all entries will pass prechecking. Also, precheck
        is a good opportunity to throw warnings about long runtimes (e.g., doing
        nearest neighbors computations on a structure with many thousand sites).

        See the documentation for precheck_dataframe for more information.

        Args:
            *x (Composition, Structure, etc.): Input to-be-featurized. Can be
                a single input or multiple inputs.

        Returns:
            (bool): True, if passes the precheck. False, if fails.

```

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+        train_set = CompositionData(X, targets=y, task=self.task, emb=self.elem_emb)
+
+        # NOTE this is needed as we declare the model upfront without knowledge of the
+        # dataset.
+        if train_set.n_targets != self.model_params["n_targets"]:
+            raise ValueError("Number of targets for dataset does not match model")
+
+        train_generator = DataLoader(train_set, **self.data_params)
+
+        # if val_set is not None:
```
ardunn, line 164: We don't have any guidelines for validation sets. Do as you wish!

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+        #         )
+        #         if self.model.task == "regression":
+        #             val_score = v_metrics["MAE"]
+        #             print(f"Validation Baseline: MAE {val_score:.3f}\n")
+        #         elif self.model.task == "classification":
+        #             val_score = v_metrics["Acc"]
+        #             print(f"Validation Baseline: Acc {val_score:.3f}\n")
+        #         self.model.best_val_score = val_score
+
+        if checkpoint:
```
ardunn, line 193: Should `checkpoint` and `log` be paths? - up to you, but just a thought. That way people can specify where to keep their models and logs. The default can remain what it is right now.

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+                # move tensors to device
+                input_ = (tensor.to(self.model.device) for tensor in input_)
+                # compute intermediate output
+                output = self.model.material_nn(*input_).numpy().ravel()
+                # output = self.model.material_nn(*input_).data.cpu().numpy().ravel()
+
+        return output
+
+    def feature_labels(self):
+        return [f"Roost_feature_{x}" for x in range(self.elem_fea_len)]
```
ardunn, line 260: Does this featurizer return a learned embedding of the material, or a prediction on the property on interest?

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+        ]
+
+
+class CompositionData(Dataset):
+    """
+    The CompositionData dataset is a wrapper for a dataset data points are
+    automatically constructed from composition strings.
+    """
+
+    def __init__(self, comp, targets=None, task="regression", emb="matscholar"):
```
ardunn, line 285: Need docstrings for all of these methods, but putting all the Roost utils stuff in one file is a good call.

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+            cry_id,
+        )
+
+
+def collate_batch(dataset_list):
+    """
+    Collate a list of data and return a batch for predicting crystal
+    properties.
+
+    Parameters
```
ardunn, line 387: Docstring should be Google style

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+            torch.cat(batch_nbr_fea_idx, dim=0),
+            torch.cat(crystal_atom_idx),
+        ),
+        torch.stack(batch_target, dim=0),
+        batch_comp,
+        batch_cry_ids,
+    )
+
+
+def init_roost(
```
ardunn, line 467: Docstrings needed

matminer/featurizers/composition/yang.py
```diff
@@ -0,0 +1,154 @@
+from __future__ import division
```
ardunn, line 1: Move YSS to `thermo` or similar module, which is currently called `cohesive`

matminer/featurizers/composition_deprecated.py
```diff
@@ -1,3 +1,5 @@
+## Deprecated
```
ardunn, line 1: I'd say just remove this file, ti's confusing to have a specific composition deprecated file outside of the composition subpackage

ardunn: Unless this is being used in some way I'm not familiar with, like what is the purpose of this file now?

matminer/featurizers/structure/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+)
+from matminer.featurizers.structure.cgcnn import (
+    CGCNNFeaturizer
+)
+
+__authors__ = ("Anubhav Jain <ajain@lbl.gov>, "
+                "Saurabh Bajaj <sbajaj@lbl.gov>, "
+                "Nils E.R. Zimmerman <nils.e.r.zimmermann@gmail.com>, "
+                "Alex Dunn <ardunn@lbl.gov>, "
+                "Qi Wang <wqthu11@gmail.com>")
```
ardunn, line 49: Overall this is a pretty good organization of the featurizers, but we need to make an `external` subpackage like I mentioned for the composition featurizers.

Might also rename the current jarvis module to "composite" or something as the Jarvis featurizer is essentially a composite of several featurization techniques. I could see future featurizers being put into that module also.

matminer/featurizers/structure_depricated.py
```diff
@@ -1,3 +1,5 @@
+# Deprecated
```
ardunn, line 1: Similar to composition, we should just delete this?

ardunn:
@CompRhys we should also separate the site.py module as you've done for composition and structure, since it is quite long and requires some external libraries. 

Also, since I am working on #488, I imagine there will be some pretty brutal merge conflicts in the future for this PR. I can help fix them when that happens though

matminer/featurizers/composition_deprecated.py
CompRhys: I was using it to check that everything was still working in the same way. I will remove it once the PR is ready but I found it helpful to keep it for reference purposes in the meantime

CompRhys:
> Keep in mind I'm being extra nitpicky (lol) 

No problem it's helpful, most of my stuff is still very much scripting based and without any formal CS experience comments are the best way to learn about better software design

> If it is ok with you, I may just locally integrate your non-Roost changes + my non-Roost review comments into that PR. Then, all we'd need to do - instead of trying to fix heinous merge commits - is open a new PR adding the single `roost.py` file and updates to `requirements-optional.txt`.

That sounds more sensible than conflating the changes. I'll do another quick refactor then to deal with the organisation stuff you flagged and split up site.py and push that. Then I'll wait till you give me the heads up to close this PR and open another one with just the single "roost.py" file and its tests.

matminer/featurizers/structure/__init__.py
CompRhys: sure!

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
+import functools
+import os
+import numpy as np
+from matminer.featurizers.base import BaseFeaturizer
+from matminer.utils.data import MatscholarElementData
+
+# NOTE comprhys: is this really the only thing we can do here?
+# won't think not throw an error for someone without Roost installed
+# trying to use the RoostFeaturizer?
+try:
```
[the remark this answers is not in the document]
CompRhys: It was still failing one of the tests without it but the error with a link is a good idea

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+    Roost = object
+    default_device = None
+
+
+class RoostFeaturizer(BaseFeaturizer):
+    """
+    Representation learning from stiochiometry
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
CompRhys: Generally, even for the 5k data set I used for developing I saw that the 3-3-256-256 setup for the graph was still locally pretty good and that making the output nn larger boosted performance. I settled on the defaults because anything much larger comes with diminishing returns.

matminer/featurizers/composition/roost.py
```diff
@@ -0,0 +1,569 @@
...
+            )
+
+        if n_targets is None:
+            if task == "regression":
+                n_targets = 1  # hard coded for single target regression
+            elif task == "classification":
+                n_targets = 2  # hard coded for binary clf
+
+        if elem_emb == "matscholar":
+            elem_emb_len = 200  # hard coded for matscholar
```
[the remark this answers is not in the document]
CompRhys: So there are two things `elem_fea_len` and `elem_emb_len`, `elem_fea_len` is a hyperparameter for which 64 was the default I settled on and  `elem_emb_len` is the dimension of the initial representations used for the elements. The `matscholar` vectors are 200 long so I just hard coded this as I wanted to make use of the data store already in `matminer`. It would also be possible to add a `MatscholarElementData.len` property or something and set it using that. 

For big datasets the choice of initial embedding doesn't matter but for small datasets I saw `matscholar` typically performed slightly better than other things which is why I would be happy just to fix the choice of `matscholar` in the `matminer` stuff. If having used it here and it worked for them and they want to dig deeper then they can play around with the full code where they would have more freedom to set it up the model or add in tweaks etc.

matminer/featurizers/composition/roost.py
CompRhys: It cannot deal with pure systems and it can't deal with compositions that include elements not present in the initial embedding dictionary.

It can scale up for something with 15 elements but the requirements do grow as O(n^2) within each material.

matminer/featurizers/composition/roost.py
CompRhys: I will remove this then as partitioning training data would hurt the model performance and in my testing I haven't yet encountered examples where early stopping would be significant. As above I think if someone wants to do an extensive search over hyperparameters I think it would be easier outside of `matminer`.

matminer/featurizers/composition/roost.py
CompRhys: I think the default should be not to log or write files to peoples systems here unless they actively set the option but I think providing paths is a sensible change

matminer/featurizers/composition/roost.py
CompRhys: So I broke the model into two modules, the descriptor module which is the core of the method and then an output nn. There's nothing special about the output nn so I thought it would be more in keeping with `matminer` to return the features from the descriptor module.

For transparency from some recent chemoinformatics work I have reason to believe that trained `roost` features + random forests should perform very well on small datasets in particular. So I was making choices based on how to make this use case as easy as possible when setting up the `RoostFeaturizer`

ardunn:
@CompRhys that's a great plan. Don't worry too much about the non-Roost related review comments I left, as I can change all of those once I start integrating this with #488

ardunn:
@CompRhys related to the RoostFeaturizer imports, I think a workable example to follow is how SOAPFeaturizer is implemented currently in matminer.featurizers.site:

```python
from monty.dev import requires

# SOAPFeaturizer
try:
    import dscribe
    from dscribe.descriptors import SOAP as SOAP_dscribe
except ImportError:
    dscribe, SOAP_dscribe = None, None

class SOAP(BaseFeaturizer):
    @requires(dscribe, "SOAPFeaturizer requires DScribe. Install from github.com/SINGROUP/dscribe")
    def __init__(...)
...

```

So 

1. Conditional import, as you have currently done
2. Use `monty.dev`'s `@requires` decorator, and include a link to docs/installation/etc

PR #484, opened 2020-06-03
Allow ElementProperty.featurize use pmg structure

afonari:
I want to compute a list of features from different classes (ElementProperty, OxidationStates, DensityFeatures, etc). It seems easier if all of them take pmg structure as an argument in the featurize function. This case is to allow this for ElementProperty.

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -187,6 +187,7 @@ class ElementProperty(BaseFeaturizer):
 
         all_attributes = []
         
+        # Allow pmg structure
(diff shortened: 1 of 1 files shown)


computron:
Hi,

Thank you for the contribution, but we built the ``StructureComposition`` featurizer for this instance. It allows once to use *any* composition featurizer (not just ElementProperty) by providing a structure as an input. We prefer this method since it works for all composition featurizers and doesn't involve adding extra code or input handling to each composition featurizer.

PR #488, opened 2020-06-09
Featurizer tuneup (WIP)

### LINKED ISSUE #142: move has_oxidation_states function to pymatgen?
See note from @WardLT 

matminer/featurizers/composition.py:29

### LINKED ISSUE #241: Add the Many-Body Tensor Representation
Matthias Rupp will be open sourcing source code the [Many-Body Tensor Representation](https://arxiv.org/pdf/1704.06439.pdf). We should build a wrapper to implement the BaseFeaturizer API, either for his package or ours.

@computron - Mathias mentions you have already done this, right?

### LINKED ISSUE #253: A structure descriptor that is the number of symmetry operations in the spacegroup
This could be added to the existing GlobalSymmetryFeaturizer

Henrique Miranda suggested that instead of just numbering the spacegroups from 1-235, to have a number representing the number of operations in the spacegroup. More operations means more symmetry.

It's a good idea.

Need to see how to do it in pymatgen. Need to make sure that something like a 4-fold axis is represented 3 times (1/4, 2/4, and 3/4 rotations - with 4/4 just being existing identity element)

### LINKED ISSUE #257: issue in GDRF?
See:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/matminer/WoSShQmpgKA/3MgdVWL3AgAJ

### LINKED ISSUE #262: it's not clear ElectronicRadialDistributionFunction needs oxidation states due to uninformative error
it seems like the following line is not enough?

# Add oxidation states.
struct = ValenceIonicRadiusEvaluator(struct).structure

I get AttributeError due to the line "this_charge = float(site.specie.oxi_state)"; it was not obvious that I can fix this by calling structure_to_oxidstructure on the structure column of my input data first.

I could fix this by adding warning to docs that oxidation states are required and raise an error when they are not but I wasn't sure what is struct = ValenceIonicRadiusEvaluator(struct).structure doing then? @nisse3000 , can you look for a fix for this? My suggestion is just to raise exception if oxidation states are not there (similar to how it's handled in EwaldEnergy).

### LINKED ISSUE #300: Provide Flat Outputs for All Structure Featurizers
Most machine learning libraries work best with 1D, "flat" features. Several of our current structure featurizers return 2D matrices or dictionaries rather than a 1D vector. We should modify these classes to provide the ability to output 1D features, if desired (see PR #299 as an example).

The relevant classes are:

- [x] CoulombMatrix
- [x] SineCoulombMatrix
- [ ] MinimumRelativeDifferences 
- [ ] RadialDistributionFunction 
- [ ] ElectronicRadialDistributionFunction 

For the two CoulombMatrix classes, we should flatten the outputs by computing eigenvalues and then padding the eigenvalues with zeros to make all outputs the same length. This is a standard practice when using CMs, where the length of vectors is chosen to be the maximum number of atoms in the unit cell for the training set. We should probably add a fitting routine to determine the maximum atom count.

### LINKED ISSUE #347: Unexpected behavior of CrystalNNFingerprint
The output of CrystalNNFingerprint does not match up with the output of the CrystalNN class.

For example the following code snippet:
```python
import numpy as np
from matminer.featurizers.site import CrystalNNFingerprint
from pymatgen.ext.matproj import MPRester
from pymatgen.analysis.local_env import CrystalNN

s = MPRester().get_structure_by_material_id("mp-707852")
s.add_oxidation_state_by_guess()
bs = CrystalNN().get_bonded_structure(s)

site_of_interest = 51

cnnf = CrystalNNFingerprint.from_preset("ops")
features = cnnf.featurize(s, 51)

sort_indices = np.argsort(features)
top_id = sort_indices[-1]
second_top_id = sort_indices[-3]

print("coordination of site {}: {}".format(
    site_of_interest, bs.get_coordination_of_site(site_of_interest)))
print("top feature: {} with OP: {}".format(
    cnnf.feature_labels()[top_id], features[top_id]))
print("second top feature: {} with OP: {}".format(
    cnnf.feature_labels()[second_top_id], features[second_top_id]))
```

Produces the output:
```
coordination of site 51: 2
top feature: wt CN_1 with OP: 0.6286992249320102
second top feature: wt CN_3 with OP: 0.3295052806667878
```

The coordination of site 51 obtained by the CrystalNN class does not match up with the coordination indicated by CrystalNNFingerprint. Even the second highest order parameter is for the wrong coordination number.

### LINKED ISSUE #380: Many featurizers are missing "Feature" blocks in docstrings as specified in BaseFeaturizer
BaseFeaturizer indicates each featurizer doc should have a "Features" block. However many featurizers do not have this! Probably would be good to go through the featurizers and add this, and standardize the docs in the featurizers generally, too.

### LINKED ISSUE #385: Featurizers missing prechecking
While #384 introduced 3 prechecks for featurizers, there are many more featurizers that could benefit from a precheck. This issue can be used for tracking these featurizers 

- [ ] `CationProperty`, `OxidationStates`, `ElectronAffinity`, `ElectronegativityDiff`, `IonProperty`: all-metal entries' oxidation states are ill-defined, so they could be considered invalid
- [ ] NN-based featurizers (e.g., `SiteStatsFingerprint`): Throw warnings (not neccessarily fail precheck) when large structures are prechecked, cuz they're gonna take forever
- [ ] Some others? Please add more featurizers to this issue list if you think of them

### LINKED ISSUE #420: SOAP implementation details need to be checked

### LINKED ISSUE #479: Add presets to soap featurizer and include structure featurizer
based on discussion in #428

### LINKED ISSUE #520: Add delta G descriptor
https://www.nature.com/articles/s41467-018-06682-4

https://github.com/CJBartel/predict-gibbs-energies/blob/master/PredictG.py

ardunn:
#### Making minor adjustments to existing featurizers:

- [X] update RDF to have flat output (fixes #300)
- [X] ~~update PRDF to have flat output (fixes #300)~~ (seems to have already been done)
- [X] update ERDF to have flat output (fixes #300)
- [X] add prechecking to (P/E/_)RDF
- [X] throw warning/clear error when ERDF tries to featurize without oxi states (fixes #262)
- [X] update MinimumRelativeDistances to have flat output (fixes #300)
- [X] ~~fix GRDF binning problem (fixes #257)~~ already fixed in #260 
- [ ] append number of symm ops to GlobalSymmetryFeaturizer (fixes #253)
- [X] update SOAP with sensible presets and a structure featurizer (or SiteStats preset) (fix #420, related to #428, fixes #479)
- [ ] add prechecking for featurizers missing it (fixes #385)
- [ ] fix CrystalNNFingerprint mismatch with CrystalNN (fixes #347)
- [ ] fix docstring issue with featurizers (fixes #380)

#### Other
- [ ] add MBTR thru dscribe (fixes #241)
- [ ] add dG descriptor (fixes #520)
- [ ] change `hasattr` to `isinstance` for composition `has_oxidation_states` and close #142 as there seems to be no plan to move `has_oxidation_states` into pymatgen


ardunn:
# WIP comments for provenance/discussion on major changes

### RDF/ReDF:
Both of these featurizers were relatively easily converted to "flat-style" structure featurizers; this is mostly because the distance bins for each are fixed for each structure _regardless_ of the structure or the number of sites. Instead of having separate "distances" and "distribution" bins, the distances are now shown in the feature labels and also stored internally as a user-facing attribute. The distribution in each bin are the features themselves. These featurizers still require no fitting to use yet provide flat output.

The one major difference is removing the auto-cutoff determination in ReDF. Previously this was done on an individual basis for each structure based on the max diagonal measurement of the unit cell. However, the difference in cutoff would require fitting to featurize multiple samples, in which case all but the largest-diagonal unit cell would have at least one invalid distance bin. Keeping this auto-cutoff would essentially require a non-flat output.

ardunn:
superseded by #634

Issue #494, opened 2020-06-30
cohesive_energy unit tests fail on CI platform but not locally


CompRhys:
No changes made to `cohesive_energy` and `cohesive_energy_MP` but tests for it fail on CI despite not failing on my local machine when I run `source dev_scripts/run_circleci_tests.sh`. From `Dependabot` it appears as though it didn't fail up until 22 days ago

Example of failure:
```
======================================================================
FAIL: test_cohesive_energy (matminer.featurizers.tests.test_composition.CompositionFeaturesTest)
----------------------------------------------------------------------
 
Traceback (most recent call last):
  File "/home/circleci/matminer/matminer/featurizers/tests/test_composition.py", line 210, in test_cohesive_energy
    self.assertAlmostEqual(df_cohesive_energy["cohesive energy"][0], 5.15768, 2)
AssertionError: 5.1793583427500005 != 5.15768 within 2 places
 
======================================================================
FAIL: test_cohesive_energy_mp (matminer.featurizers.tests.test_composition.CompositionFeaturesTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/matminer/matminer/featurizers/tests/test_composition.py", line 217, in test_cohesive_energy_mp
    self.assertAlmostEqual(df_cohesive_energy["cohesive energy (MP)"][0], 5.945, 2)
AssertionError: 5.978027364000001 != 5.945 within 2 places
----------------------------------------------------------------------
```

ardunn:
Hey Rhys, thanks for bringing this to our attention. I'll fix this. Don't worry if this is the only test is failing for your PR's CI!

PR #497, opened 2020-07-08
Fix hashes for expt_* matbench datasets, accounting for 1128G typo

ardunn:
Also, correct cohesive energy test


ardunn:
fixes #494

Issue #503, opened 2020-07-21
Type stubs


janosh:
What's your take on shipping [type stubs](https://mypy.readthedocs.io/en/stable/stubs.html)? This would help modern code analysis tools such as [Pylance](https://devblogs.microsoft.com/python/announcing-pylance-fast-feature-rich-language-support-for-python-in-visual-studio-code/) do a better job and as a result make it easier for end users to work with `matminer`.

Specifically, it would prevent issues such as [this one](https://github.com/microsoft/pylance-release/issues/149).

Irrespective of type stubs, it's probably a good idea to address the problem @erictraut described in [pylance-release#149](https://github.com/microsoft/pylance-release/issues/149#issuecomment-661913283)

> the `ConversionFeaturizer.featurize_dataframe` method is calling a method by the same name in its super class, `BaseFeaturizer`. That method, in turn, is calling `_generate_column_labels`, which is calling method `feature_labels`. This method is implemented with a `raise NotImplementedError()` statement, which will throw an exception. It appears that the `BaseFeaturizer` class is intended to be an abstract base class, and `feature_labels` is intended to be an abstract method that is overridden by subclasses. However, `BaseFeaturizer` is not marked as an ABC, nor is `feature_labels` decorated with `@abstractmethod`.

Happy to submit a PR for this if maintainers are busy.

ardunn:
Hey @janosh great idea!

BaseFeaturizer should indeed be ABC. Not sure why it never was. 

I'm also a fan of inline type-hinting (much of which matminer is missing, just because no one has gone through and done it). I'm not fully convinced about stub files though, from a maintainability perspective... mainly mismatching outputs between featurizers in .py files and their stubs in .pyi. 

And yes, PRs welcome!! Please link to  issue if you open it.

WardLT:
Yea, using ABC would be excellent! I second @ardunn and would be happy to help review a PR!

janosh:
Thanks for your offers to review.

@ardunn To be honest, I also prefer inline type hints.

computron:
I see this is closed already, but just wanted to mention that:

- yes, the abc decorators should have already been there, thanks for catching this and offering to add them!
- as discussed by others, I prefer inline versus type stub files, as the former is easier to keep synced and maintain. We've dropped Py2 support since the underlying libraries of matminer (e.g., pymatgen) no longer support Py2 anyway.

janosh:
Oops, closing must have happened by mistake.

PR #504, opened 2020-07-22
make BaseFeaturizer an ABC with abstractmethods

janosh:
Addresses the minor point in #503.

Question: Shouldn't `BaseFeaturizer.fit` be an `abstractmethod` and raise `NotImplementedError`? Right now it's

```py
    def fit(self, X, y=None, **fit_kwargs):
        """Update the parameters of this featurizer based on available data

        Args:
            X - [list of tuples], training data
        Returns:
            self
            """
        return self
```

### DIFF
--- a/matminer/featurizers/base.py
+++ b/matminer/featurizers/base.py
@@ -3,19 +3,20 @@ from __future__ import division, unicode_literals
 import sys
 import traceback
 import warnings
-from multiprocessing import Pool, cpu_count
+from abc import ABC, abstractmethod
 from functools import partial
+from multiprocessing import Pool, cpu_count
 
 import numpy as np
 import pandas as pd
-from six import string_types, reraise
-from sklearn.base import TransformerMixin, BaseEstimator, is_classifier
+from six import reraise, string_types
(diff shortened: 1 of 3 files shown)


janosh:
@ardunn @computron @WardLT Just noticed that the tests aren't passing because a bunch of derived classes don't implement some methods I decorated with `@abstractmethod`.

Specifically, `MultiArgs2`, `MultiTypeFeaturizer`, `DOSFeaturizer`, `DosAsymmetry`, `Hybridization` and `SiteDOS` don't implement `citations()` and/or `implementors()`. Should I remove `@abstractmethod` from those methods or do you want to implement them?

computron:
Hi Janosh,

Thanks for asking about this - could you do the following?
- Just have these return an empty list for now (i.e., implement to return an empty list)
- In the pull request, put a note that you did this. Then I can remember to go in and add implementors and citations for some of these, where the original authors did not add them in.

computron:
Actually nevermind the second point, I'll just add an issue about that now

janosh:
@janosh Done. Hope I didn't miss any.

computron:
Thanks!

PR #547, opened 2020-11-11
Changed the equation of band center estimation

utopianf:
## Summary

* The χ values in the paper [Butler, M A and Ginley, D S, 1978] are the arithmetic mean of the atomic electron affinity and
the first ionization energy , but not electronegativity (pymatgen.Element.X).
* Fixed the equation. 
* Assuming the band gap of TiO2 to be 3.2 eV, we can obtain the electron affinity of TiO2 to be about 4.2 eV, which is near the value in the paper.


ardunn:
Thanks for the contribution!

PR #576, opened 2021-02-05
A set of features based on 'Machine learning assisted design of high entropy alloys with desired property' by Wen et al.

RossKunz:
Addition to Composition.py by adding specific alloy features developed by Wen et al.

Self Contained Features:
('Weight Fraction')
('Atomic Fraction')
('Yang delta')
('Yang omega')
('APE mean')
('Radii local mismatch')
('Radii gamma')
('Configuration entropy')
('Atomic weight mean')
('Total weight')
('Lambda entropy')
('Electronegtivity delta')
('Electronegativity local mismatch')
('VEC mean')
('Mixing enthalpy')
('Mean cohesive energy')
('Interant electrons')
('Interant s electrons')
('Interant p electrons')
('Interant d electrons')
('Interant f electrons')
('Shear modulus mean')
('Shear modulus delta')
('Shear modulus local mismatch')
('Shear modulus strength model')

### DIFF
--- a/matminer/featurizers/composition.py
+++ b/matminer/featurizers/composition.py
@@ -2036,7 +2036,6 @@ class WenAlloys(BaseFeaturizer):
         self.data_source_magpie = MagpieData().all_elemental_props
         self.data_source_cohesive_energy = CohesiveEnergyData()
         self.data_source_enthalpy = MixingEnthalpy()
-        self.structure = 'fcc'
         self.element_feature_labels = []
         self.element_feature_labels.append('Weight Fraction')
         self.element_feature_labels.append('Atomic Fraction')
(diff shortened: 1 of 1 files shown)


ardunn:
Hey @RossKunz thanks for the contribution! This is great, sorry for the delay in response. 

There's just a few things we need before merging it in:

# tests
The class needs some dedicated tests; see the other composition featurizers tests for some examples of how to do this! It doesn't have to be totally comprehensive, but should at least test the fundamental parts of the class.

### docstring
Could you change the top line of the featurizer's docstring to `Calculate features for alloy properties.`? The copyright can be kept, just maybe move it down a line. This is just needed for our auto-generated documentation.

### prechecking
It looks like this class only applies to alloys; would you mind adding a`precheck` method to help check compositions to make sure they are valid for featurization? All that `precheck` should do is take in one composition, and return a `bool` whether the composition is valid for this featurizer or not. See the documentation for `BaseFeaturizer` (search "precheck") or see `MiedemaFeaturizer` for an example.

ardunn (changes requested):

ardunn:
merged for now, I will add the above requirements. Again @RossKunz thanks for the contribution!

PR #603, opened 2021-03-25
Bump pymatgen and fix tests

utf:
## Summary

Update to the latest pymatgen version and fix the failing tests.


utf:
The latest pymatgen is python 3.7+ only, so I also updated the CircleCI testing environments to be 3.7 and 3.8 (rather than 3.6 and 3.7).

utf:
@ardunn does this look ok to you?

ardunn:
looks good to me!

PR #621, opened 2021-05-20
(1.) Fix links to example ipynbs from *master* to *main* branch in Nbviewer, (2.) Fix MPDS retrieval

blokhin:
The links to the example notebooks are now defunct, so fixing them everywhere.


ardunn:
Thanks for the contribution @blokhin !

PR #641, opened 2021-06-10
Bugfix for installing package data

ncfrey:
## Summary

I was getting a file not found error when using the `SineCoulombMatrix` featurizer because [cn_target_motif_op.yaml](https://github.com/hackingmaterials/matminer/blob/main/matminer/featurizers/site/cn_target_motif_op.yaml) wasn't being included when I pip installed matminer. It looks like `MANIFEST.in` controls this, so I added a line to `setup.py` to make sure package data is included. Now it seems to be working.

* Fix pip install package data

I'm using python 3.7.10, pymatgen v2022.0.8, and the latest version of matminer.


ardunn:
Thanks @ncfrey !

PR #644, opened 2021-06-14
Fix package data

utf:
## Summary

New package layout broke installing matminer using pip without an editable installation. I've corrected `package_data` in setup.py which hopefully should fix this issue (although it is hard to test without releasing a new pypi version).

I guess this is also related to #641


ardunn:
yep lol I was just about to commit this exact change, thanks

PR #759, opened 2022-01-06
Flake8+isort

janosh:
Manual fixes for issues reported by `flake8 matminer/**/*.py` and automatic import sorting from `isort matminer/**/*.py `.

Worth considering adding these as `pre-commit` hooks for future code.


janosh:
CI error 

```py
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

appears unrelated as it also occurs on `main`:

https://github.com/hackingmaterials/matminer/actions/runs/1664563599

Not sure what caused it...

ardunn:
@janosh I believe the insane errors described above have been fixed. matminer only supports py38+ now as newest pymatgen has that same requirement. Worth editing this PR or should I close?

janosh:
Thanks for fixing deps! 👍

PR #765, opened 2022-01-13
add vicker's hardness dataset

ardunn:
@sgbaird could you answer the following for me?

1. Is it ok to upload a .json.gz version of this dataset to our figshare so there is a persistent resource for download?
2. Would it be possible to include any or all of the following as additional columns:

- Original source reference (doi and/or bibtex preferred)
- Spacegroup number or HM symbol
- Any other pertinent sample descriptors that might have been recorded in earlier versions of the dataset

Also go ahead and look over the metadata addition and make sure the name/description are correct and complete (and feel free to make any needed revisions).

Current data looks like this:
```
                composition  hardness  load
0     Ag0.05Gd0.048Pd0.902      1.810  0.49
1      Ag0.05Y0.048Pd0.902      1.640  0.49
2       Ag0.25Pb0.5Sb0.25Te     0.578  2.94
3       Al1.5Si1.5N2.5O1.5     15.030  0.98
4                 Al1.67B22    23.800  2.00
                     ...       ...   ...
1057                  ZrV2      3.600  0.98
1058                   ZrW2    11.100  0.98
1059                   OsB2    34.800  0.25
1060                   OsB2    27.000  0.49
1061                   OsB2    17.800  1.96
```

Once we figure out the above questions I'll upload the final version to figshare and the tests should pass.


sgbaird:
@ardunn I appreciate you taking the initiative on this! I'm actually not affiliated with the group and [had a similar question about the LICENSE](https://github.com/ziyan1996/VickersHardnessPrediction/issues/1).

@ziyan1996 @BrgochGroup do you mind taking a look at these questions? As some background, I [suggested](https://github.com/materialsproject/matbench/issues/2#issuecomment-1011803930) the dataset in https://github.com/ziyan1996/VickersHardnessPrediction as an addition to the 13 materials benchmarking tasks in the next version of [matbench](https://matbench.materialsproject.org/). I think this would both be a useful dataset for people to use for benchmarking their materials informatics model as well as give additional limelight for the large data mining effort.

ardunn:
@sgbaird it might be worth for us to email them, perhaps they are not on github as actively as we are

sgbaird:
@ziyan1996 mentioned being in the middle of job interviews, so it might take some time. Something else worth mentioning is [the hardness dataset](https://mpds.io/#inquiry/props=vickers%20hardness%20number) available through MPDS, but unfortunately, it doesn't seem to have applied load information (https://github.com/mpds-io/mpds-api/issues/38). See also [matsci post](https://matsci.org/t/do-you-know-of-curated-experimental-vickers-hardness-datasets-with-composition-temperature-and-load/40488).

EDIT: based on some initial tests, without knowing the applied load, it doesn't seem like the dataset has very much predictive potential (even when predicting the mean hardness for a given composition).

ardunn:
Ok, that is fine since I'm sure they are busy. Let's maybe wait a month or two and re-evaluate then

PR #786, opened 2022-03-04
Vicker's hardness dataset

ardunn:
See #765, commits got messed up so had to make a new pr


ardunn:
For reference, we are still waiting on @ziyan1996 to make a call on whether their License will allow this or not

ardunn:
Closes #795

PR #880, opened 2022-10-27
suggestion to switch to `pip-tools`

tschaume:
The explicit dependency of `matminer` on `jsonschema>=4.5.1` causes a conflict in the MP website dependencies due to MPContribs needing `jsonschema<4.0`:

```
There are incompatible versions in the resolved dependencies:
  jsonschema<4.0 (from mpcontribs-client==4.5.0->mp-web (pyproject.toml))
  jsonschema>=4.5.1 (from matminer==0.7.8->robocrys==0.2.7->crystal-toolkit[server]==2022.10.27->mp-web (pyproject.toml))
```

Does `matminer` need `jsonschema>4.0` explicitly?

We could avoid other similar/future conflicts caused by reading pinned requirements files into `setup.py` by switching to [pip-tools](https://github.com/jazzband/pip-tools). This PR moves dependencies to the `setup.py` and uses `pip-compile` to create the requirements files needed for the github actions:

```bash
pip-compile # creates requirements.txt
pip-compile --all-extras -o requirements-extras.txt
```


ardunn:
So I don't think it needs jsonschema pinned at that version specifically, the only thing it needs it for is retrieving MPDS data

But I am open to using pip-tool instead, I like the idea of generating the requirements.txt. But does dependabot mess with it? @tschaume

tschaume:
@ardunn ok, good to know. I moved `jsonschema` to the `mpds` extra.

Good question about dependabot. It's been bothering me how noisy it is, too. Using `pip-tools`, I'm considering switching to [this approach](https://www.oddbird.net/2022/06/01/dependabot-single-pull-request/) for my repos.

tschaume:
@ardunn I added the `upgrade-dependencies` action as outlined [here](https://www.oddbird.net/2022/06/01/dependabot-single-pull-request/). A personal access token named `PAT` will have to be added to the repo for the action to be able to create the PR containing all updated dependencies.

PR #897, opened 2023-03-25
Optimize `BandCenter` and `IntersticeDistribution` featurizers

ml-evs:
This PR simply initialises various data sources at the class or instance level, rather than in the featurizer loop, which was causing the data sources to be loaded from disk for every featurization leading to serious performance issues (as one might expect).

`BandCenter`, for example, is now 5 orders of magnitude faster on my machine.

I couldn't spot any other featurizers that have this issue though my search wasn't exhaustive.

### DIFF
--- a/matminer/featurizers/composition/element.py
+++ b/matminer/featurizers/composition/element.py
@@ -207,6 +207,9 @@ class BandCenter(BaseFeaturizer):
         - Band center
     """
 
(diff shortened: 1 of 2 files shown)


ml-evs:
Only tests are failing as I don't have access to the MP API key from my fork, will merge this.

Issue #909, opened 2023-06-27
WenAlloy wrong valence electron counts


ardunn:
https://matsci.org/t/a-probable-mistake-of-wenalloy-featurizer/48445

PR #917, opened 2023-12-27
Correct valence electron numbers

DanielYang59:
## Summary

Corrected a few valence electron numbers as pointed out in #909.


ml-evs (approved):
Thanks @DanielYang59!

DanielYang59:
@ml-evs Thanks for reviewing!

Issue #920, opened 2024-02-06
Simple composition-based featurization fails due to an upgrade in pymatgen


JaGeo:
```Python
from matminer.datasets import load_dataset
from matminer.featurizers.composition import ElementProperty
from matminer.featurizers.conversions import StrToComposition

bandgaps = load_dataset('matbench_expt_gap')
bandgaps = bandgaps[bandgaps['gap expt'] > 0].reset_index(drop=True)
bandgaps.head()

str_to_comp = StrToComposition(target_col_id='composition_pmg')
bandgaps_comp = str_to_comp.featurize_dataframe(bandgaps, col_id='composition')

featurizer = ElementProperty.from_preset('magpie')
bandgaps_data_featurized = featurizer.featurize_dataframe(bandgaps_comp, col_id='composition_pmg', ignore_errors=True,
                                                          return_errors=True)
y = bandgaps_data_featurized['gap expt']
bandgaps_data_featurized.head()

```

fails with:

```
Traceback (most recent call last):
  File "/hpc-user/jgeorge/PycharmProjects/test_new/use_matminer.py", line 2, in <module>
    from matminer.featurizers.composition import ElementProperty
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/featurizers/composition/__init__.py", line 1, in <module>
    from matminer.featurizers.composition.alloy import Miedema, WenAlloys, YangSolidSolution
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/featurizers/composition/alloy.py", line 16, in <module>
    from matminer.featurizers.composition.packing import AtomicPackingEfficiency
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/featurizers/composition/packing.py", line 13, in <module>
    from matminer.featurizers.composition.element import ElementFraction
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/featurizers/composition/element.py", line 202, in <module>
    class BandCenter(BaseFeaturizer):
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/featurizers/composition/element.py", line 210, in BandCenter
    magpie_data = MagpieData()
  File "/home/jgeorge/miniconda3/envs/test_new/lib/python3.10/site-packages/matminer/utils/data.py", line 217, in __init__
[...]

PR #921, opened 2024-02-06
Fix maximum element failure

JaGeo:
Should fix https://github.com/hackingmaterials/matminer/issues/920 by introducing a maximum element number. Og is probably enough.

### DIFF
--- a/matminer/utils/data.py
+++ b/matminer/utils/data.py
@@ -11,7 +11,7 @@ from glob import glob
 
 import numpy as np
 import pandas as pd
-from pymatgen.core.periodic_table import Element, _pt_data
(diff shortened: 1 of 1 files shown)


JaGeo:
@ml-evs Any ideas where the failures could come from? To dig into this, I would need to fix the installation locally first as well ...

ml-evs:
Hi @JaGeo, the changes in this PR look fine to me, the failing tests are simply from the lack of pandas 2.0 support (and the rest are passing fine for me locally). If you did want to verify this yourself, you can install the pinned deps in the various requirements files in this repo rather than using the loose pyproject versions.

ml-evs (approved):
Thanks @JaGeo!

PR #926, opened 2024-03-21
Extend deml_elementdata.py

zzccchen:
## Extend the heat_cap item

Extend the heat_cap item in the properties dictionary based on https://www.engineersedge.com/heat_transfer/heat_capacity_of_the_elements_13260.htm , especially the rare earth element data

* Fixed an exception where NaN appears when calculating deml heat_cap when rare earth elements such as Ce appear in the composition.

## TODO

* Data are still lacking for some actinides


ml-evs:
Thanks for this @zzccchen! 

>     * Fixed an exception where NaN appears when calculating deml heat_cap when rare earth elements such as Ce appear in the composition.

Would you be able to add a test that triggers this case? I'm hoping to do a release very soon and it would be good to get this PR in

zzccchen:
Hi @ml-evs 
I added a test case with the composition Mg99.21Zn0.18Mn0.02Ce0.48Si0.01La0.1. 
Nonetheless, I found that an exception for the WenAlloy case was triggered during the automated test, which puzzled me because the deml_elementdata.py I modified was not called in the WenAlloy method.

zzccchen:
@ml-evs Thank you very much for your reply and support! I may not have enough time and energy to troubleshoot and fix it in depth. Still, I'm glad to see changes in #928 making testing more stable.

ml-evs:
> @ml-evs Thank you very much for your reply and support! I may not have enough time and energy to troubleshoot and fix it in depth. Still, I'm glad to see changes in #928 making testing more stable.

No problem; we can always make another release soon afterwards now that these other issues are sorted. Just ping me when you want another review.

PR #927, opened 2024-03-21
Update deml_elementdata.py

zzccchen:
## Update the electric_pol item

Update the **electric_pol** item in the properties dictionary based on the latest literature [1], especially the rare earth element data.

[1] Schwerdtfeger, Peter, and Jeffrey K. Nagle. "2018 Table of static dipole polarizabilities of the neutral elements in the periodic table." Molecular Physics 117.9-12 (2019): 1200-1225. https://doi.org/10.1080/00268976.2018.1535143

* Fixed an exception where NaN appears when calculating deml **electric_pol** when rare earth elements such as Ce appear in the composition.

## TODO

* Past calculation results may not be reproducible on deployed devices.


ml-evs:
Has the source data changed in this PR? I see that many existing values also update (e.g., Li goes from 24.33 to 24.31896322). It might be good to add this as a new dataset and keep the old one for posterity too.

zzccchen:
I agree with your idea. It may be more appropriate to add a new feature to the features of the "deml" preset, such as "electric_pol_2019".

ml-evs:
> I agree with your idea. It may be more appropriate to add a new feature to the features of the "deml" preset, such as "electric_pol_2019".

That would be great if you're willing to do that!

PR #937, opened 2024-04-22
Added missing `impute_nan` for `WenAlloys` and removed unnecessary warning

gbrunin:
## Summary

This fixes 2 small things: 
- A warning about NaN imputation was written in case ElementProperty was initialiazed with impute_nan=False, even though the data source could be itself initialized with impute_nan=True. The warning is removed, since the imputation is only happening at the level of the data source.
- I missed an imputation in WenAlloys featurizer.

I'm make some tests now to decide whether the default of impute_nan should be switched to True.


ml-evs (approved):
:+1:

PR #940, opened 2024-07-29
Disable release creation workflow on forks

gbrunin:
I kept having this issue with our fork:
https://github.com/Matgenix/matminer/actions/runs/10141810059/job/28039988779

which I fixed by setting GITHUB_TOKEN and PAT explicitly. I'm not super familiar with the test.yml file so maybe there's a better fix (or maybe it should not be fixed), let me know.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,7 +92,8 @@ jobs:
       - test
     runs-on: ubuntu-latest
     env:
-      GITHUB_TOKEN: ${{ secrets.PAT }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PAT: ${{ secrets.PAT }}
     steps:
       - uses: rymndhng/release-on-push-action@v0.25.0
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,9 +91,9 @@ jobs:
     needs:
       - test
     runs-on: ubuntu-latest
+    if: github.repository == 'hackingmaterials/matminer'
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


gbrunin:
Not sure why the 2 tests on datasets are failing with python 3.9

.github/workflows/test.yml
```diff
@@ -92,7 +92,8 @@ jobs:
       - test
     runs-on: ubuntu-latest
```
ml-evs, line 93: I think it's probably easier to just disable this on forks with something like (might need to move it into the steps)

```suggestion
    runs-on: ubuntu-latest
    if: github.repository == 'hackingmaterials/matminer'
```

This part of the release doesn't seem to be used much; in the past I've just manually made the release and its all seemed to work.

.github/workflows/test.yml
```diff
@@ -91,9 +91,9 @@ jobs:
     needs:
       - test
     runs-on: ubuntu-latest
+    if: github.repository == 'hackingmaterials/matminer'
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      PAT: ${{ secrets.PAT }}
+      GITHUB_TOKEN: ${{ secrets.PAT }}
```
ml-evs, line 96:
```suggestion
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

ml-evs (commented):
Not sure why the test pypi build is failing, or why it is even triggering...

.github/workflows/test.yml
ml-evs: Whoops, misread the original

ml-evs (approved):
I think the other failures are unrelated, so happy to merge this now if it fixes things at your end @gbrunin.

PR #941, opened 2024-07-30
Disable test PyPI build

ml-evs:
The credentials and mechanism used for the test PyPI workflow are now defunct. I'd rather remove it than take the maintenance burden. The package build test is more robust.


ml-evs:
Hmm, might have just been an intermittent failure. Closing for now.

PR #942, opened 2024-08-21
Bug fixes in pseudo inverse computations

gbrunin:
## Summary

There were two small bugs in the pseudo inverse computations. 
1) in utils/get_elem_in_data, if e.g. "I" was not present in the dataset but "Ir" was, "I" was mistakenly found. 
2) the case of hydrogen was not fully handled, as there is also deuterium and tritium in `pymatgen.core.Element`. Hydrogen's pseudo inverse was a third of what if should have been (sum of all isotope values). This affects the average slightly, so a few values in the tests had to be adapted. 

Additionally, I also had issues with pre-commit due to types-all, which has been deprecated, and I fixed it by listing all needed types. 

Tests are passing in my branch with python 3.9 and 3.10, not with 3.11 but the tests that are failing are not related to these changes, so I don't know.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,10 @@ repos:
     hooks:
       - id: mypy
         additional_dependencies:
-            - types-paramiko
+          - types-setuptools
+          - types-paramiko
+          - types-requests
+          - types-ujson
 
 #  - repo: local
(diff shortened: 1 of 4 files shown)


ml-evs (approved):
:+1:

PR #943, opened 2024-08-22
spglib 2.5 compatibility

esoteric-ephemera:
Simple changes to accommodate updates to spglib's `get_symmetry_dataset` method in 2.5.0, which requires dot-access rather than dict-key access. Also pins the minimum version of spglib to account for this change.


janosh (changes requested):
`spglib` is a regular, not a build time dep. add it here

https://github.com/hackingmaterials/matminer/blob/8e99cf6c531588b79abfbcbf734fac49ce3914a7/setup.py#L65

esoteric-ephemera:
@janosh my bad, fixed!

janosh (approved):
thanks a lot @esoteric-ephemera! 👍

janosh:
@esoteric-ephemera unfortunately it's installing 2.4 in CI, not 2.5 because of

https://github.com/hackingmaterials/matminer/blob/8e99cf6c531588b79abfbcbf734fac49ce3914a7/requirements/ubuntu-latest_py3.10.txt#L293

need to update (or delete) those files as well

janosh:
some tests are genuinely failing, apparently due to a discontinued/moved superconductivity2018 website. don't look that important, i think we can `pytest.mark.skip(reason="...")` these for now

```py
_______________ MatminerDatasetsTest.test_superconductivity2018 ________________

self = <matminer.datasets.tests.test_datasets.MatminerDatasetsTest testMethod=test_superconductivity2018>

    def test_superconductivity2018(self):
        object_headers = ["composition"]
        numeric_headers = ["Tc"]
    
>       self.universal_dataset_check("superconductivity2018", object_headers, numeric_headers)

matminer/datasets/tests/test_datasets.py:643: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
matminer/datasets/tests/test_datasets.py:[79](https://github.com/hackingmaterials/matminer/actions/runs/10530585976/job/29182138461?pr=943#step:8:80): in universal_dataset_check
    self.assertTrue(download_page.ok)
E   AssertionError: False is not true
```

esoteric-ephemera:
Really weird - the tests all pass on my side, and for 3.10 (also manually checked the download links). Added the option to repeat trying the connection a few times in the test to see if any pass

janosh:
in case you want to to exclude `.DS_store` to your global gitignore, something like this should do it (didn't test)

```
git config --global core.excludesfile ~/.gitignore_global && echo .DS_Store >> ~/.gitignore_global
```

matminer/featurizers/structure/bonding.py
```diff
@@ -805,8 +805,8 @@ def get_equiv_sites(self, s, site):
         sga = SpacegroupAnalyzer(s, symprec=0.01)
         sg = sga.get_space_group_operations
         sym_data = sga.get_symmetry_dataset()
-        equiv_atoms = sym_data["equivalent_atoms"]
-        wyckoffs = sym_data["wyckoffs"]
+        equiv_atoms = sym_data.equivalent_atoms
+        wyckoffs = sym_data.wyckoffs
```
ml-evs, line 809: Do we need to break old spglib compatibility here? The deps of matminer are really not kept super up to date anymore, so I'm worried that adding new lower pins for stuff like this will cause breakages down the line. How about something like:

```suggestion
        equiv_atoms = getattr(sym_data, "equivalent_atoms", sym_data["equivalent_atoms"])
        wyckoffs = getattr(sym_data, "wyckoffs", sym_data["wyckoffs"])
```

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
      # pin NumPy version used in the build
      "numpy>=1.20.1",
```
ml-evs, line 4:
```suggestion
     "numpy >= 1.20.1, <2",
```

Just because I noticed it here... we definitely aren't testing numpy 2 yet

matminer/featurizers/structure/bonding.py
esoteric-ephemera: Good with me, reverted the lower pin on spglib and implemented your suggestion!

ml-evs: Thanks! Let's see if the test pass with the current pin (2.4), and if they do, we can bump the pinned version to 2.5.

esoteric-ephemera:
Tests are passing for with the lower version pin of spglib, but not sure why auto-gen-release can't find a github token

ml-evs:
> Tests are passing for with the lower version pin of spglib, but not sure why auto-gen-release can't find a github token

There are many things I don't understand with the CI set up here, don't worry about it... otherwise, sounds great, if you re-commit the spglib 2.5 pin (and now leave it off of setup.py) then I'll merge!

ml-evs:
> > Tests are passing for with the lower version pin of spglib, but not sure why auto-gen-release can't find a github token
> 
> There are many things I don't understand with the CI set up here, don't worry about it... otherwise, sounds great, if you re-commit the spglib 2.5 pin (and now leave it off of setup.py) then I'll merge!

Huh, actually looks like my latest auto-update PR also picked up the new spglib, without hitting this breakage... Perhaps we could just add a little test here too?

esoteric-ephemera:
> Huh, actually looks like my latest auto-update PR also picked up the new spglib, without hitting this breakage... Perhaps we could just add a little test here too?

Trying to see if setting the workflow permissions explicitly resolves this - don't think this is spglib related

esoteric-ephemera:
It seems pretty consistent that the `superconductivity2018` and `tholander_nitrides` datasets fail the URL status check - not sure why that is, but is it OK to skip these for now?

ml-evs:
Sure, they seem unrelated, let's merge this for now

ml-evs (approved):
Thanks @esoteric-ephemera!

PR #949, opened 2024-10-06
Support NumPy 2

DanielYang59:
### Summary

- Support NumPy 2
- [x] Test both [NumPy 1](https://github.com/hackingmaterials/matminer/pull/949/checks?check_run_id=31133704592) and [NumPy 2](https://github.com/hackingmaterials/matminer/pull/949/checks?check_run_id=31133500186)
- `pytest` workflow only show top 20 duration instead of all, otherwise the log page is too long to navigate through

### Future PR
- Completely migrate to `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-     # pin NumPy version used in the build
-     "numpy>=1.20.1, <2",
+     "numpy>=2.0.0",
      "setuptools>=43.0.0",
 ]
 build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,5 @@
 [build-system]
 requires = [
-     "numpy>=2.0.0",
      "setuptools>=43.0.0",
 ]
 build-backend = "setuptools.build_meta"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,7 +73,7 @@ jobs:
 
     - name: Run tests
       run: |
-        python${{ matrix.python-version }} -m pytest --cov=matminer matminer --durations=0 --timeout=360
+        python${{ matrix.python-version }} -m pytest --cov=matminer matminer --durations=20 --timeout=360
 
     - name: Build package
       if: matrix.python-version == 3.9
--- a/matminer/__init__.py
+++ b/matminer/__init__.py
@@ -1,9 +1,9 @@
 """data mining materials properties"""
 
-from pkg_resources import DistributionNotFound, get_distribution
+from importlib.metadata import PackageNotFoundError, version
 
 try:
-    __version__ = get_distribution("matminer").version
-except DistributionNotFound:  # pragma: no cover
+    __version__ = version("matminer")
+except PackageNotFoundError:  # pragma: no cover
     # package is not installed
     pass
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ if __name__ == "__main__":
         include_package_data=True,
         zip_safe=False,
         install_requires=[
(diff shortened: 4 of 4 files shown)


DanielYang59:
@ml-evs It seems unit test are passing for both numpy 1 and numpy 2, is there any reason to pin `numpy < 2` that I missed?

The test failure in `auto-gen-release` should be [restricted access to repo secrets from a fork](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions?tool=webui):
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

```
Type:     clojure.lang.ExceptionInfo
Message:  Expected environment variable to be set: GITHUB_TOKEN
Data:     {:env/name "GITHUB_TOKEN"}
Location: /var/src/release-on-push-action/src/release_on_push_action/core.clj:13:7
```

pyproject.toml
```diff
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-     # pin NumPy version used in the build
-     "numpy>=1.20.1, <2",
+     "numpy>=2.0.0",
```
ml-evs, line 3:
```suggestion
     "numpy>=1.23",
```

Is there a reason to be stricter here? I'm not actually sure anything in matminer needs to be built with numpy, but imagine this is legacy and its effects are untested

ml-evs (commented):
> @ml-evs It seems unit test are passing for both numpy 1 and numpy 2, is there any reason to pin numpy < 2 that I missed?

Not that I know of, though can imagine that test coverage isn't good enough to say for sure. I'm happy to merge this and see if anything breaks downstream

pyproject.toml
DanielYang59: > Is there a reason to be stricter here?

Yes we have a build with numpy 2 as ["NumPy’s ABI is forward but not backward compatible"](https://numpy.org/devdocs/dev/depending_on_numpy.html#build-time-dependency):
> When you build wheels for your package using a NumPy 1.xx version at build time, those will not work with NumPy 2.0.
> When you build wheels for your package using a NumPy 2.x version at build time, those will work with NumPy 1.xx.

DanielYang59:
> Not that I know of, though can imagine that test coverage isn't good enough to say for sure.

Thanks for your reply, in `pymatgen` we have seen a lot of numpy2 compatible issues surrounding cython code for Windows as the default int type changed for Windows in Numpy 2, but it seem there's no cython code for `matminer`?

pyproject.toml
DanielYang59: > I'm not actually sure anything in matminer needs to be built with numpy, but imagine this is legacy and its effects are untested

I would have a look later. I'm not sure familiar with the code base, but it seems it's a pure-python package? i.e. no c extersion or cython module and so on?

ml-evs (approved):
> Thanks for your reply, in `pymatgen` we have seen a lot of numpy2 compatible issues surrounding cython code for Windows as the default int type changed for Windows in Numpy 2, but it seem there's no cython code for `matminer`?

Not that I'm aware of, so let's just remove it (as you have done).

Thanks for this!

DanielYang59:
no problem, thanks for reviewing. I might push another PR to completely migrate `setup.py` to `pyproject.toml`, is there any reason not to migrate at this moment?

ml-evs:
> I might push another PR to completely migrate `setup.py` to `pyproject.toml`, is there any reason not to migrate at this moment?

Only lack of time/interest -- I don't mind maintaining pyproject.toml over setup.py long term but I don't think we have enough people to support many other changes

DanielYang59:
Cool, I will give a hand on this one

PR #950, opened 2024-10-07
Migrate setup.py to pyproject.toml

DanielYang59:
## Summary

- Migrate setup.py to pyproject.toml


pyproject.toml
```diff
@@ -1,9 +1,77 @@
...
+requires-python = ">=3.9"
+dynamic = ["version"]
+classifiers = [
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Development Status :: 4 - Beta",
+  "Intended Audience :: Science/Research",
+  "Intended Audience :: System Administrators",
```
DanielYang59, line 23: This looks a bit off, why is `matminer` intended for `system admins`?

JaGeo: Just change if you think this doesn't fit.

DanielYang59: Thanks for the comment, I would remove it for now unless it's intended for other purposes @ml-evs

ml-evs: No idea, I'm just a caretaker

ml-evs: Don't think trove classifiers are that load-bearing though

pyproject.toml
```diff
@@ -1,9 +1,76 @@
...
+    "coverage",
+    "coveralls",
+    "flake8",
+    "black",
+    "pylint",
+    "sphinx"
+]
+tests = ["ujson", "jmespath", "httplib2", "ase", "jsonschema",
+  "dscribe~=2.1", "mdf-forge", "aflow", "citrination-client"
+]
```
ml-evs, line 63: Might be cleaner to do

```suggestion
tests = ["matminer[mpds,describe,mdfforge,aflow,citrine,dev]"]
```

DanielYang59: Beautiful! I wasn't aware of this even exists! This is what I was looking for

DanielYang59:
No problem, thanks for reviewing and the new trick :)

